PR #4, opened 2020-12-08
Make amount and logger optional

limemloh:
This PR introduces a `payable` and a `enable_logger` attribute for both the `#[init]` and `#[receive]` proc macro.

A payable init/receive function is allowing the amount to *not* be zero.

### Without having `payable` and `enable_logger`

- The init function takes just the init context as argument (also the state if `low_level` is set).<br>
- Receive functions take the receive context and the current state.

```rust
#[init(contract = "two-step-transfer", parameter = "InitParams")]
fn contract_init(ctx: &impl HasInitContext<()>) -> Result<State, InitError> {
```

Code for ensuring the amount is zero is generated.

Notice: the amount and the logger are not arguments anymore in this case.

### Adding `payable`

- init/receive functions take an extra argument, which is the `amount: Amount` and it is positioned right after the context.
- The code ensuring the amount is zero is *not* generated in this case.

```rust
#[init(contract = "two-step-transfer", parameter = "InitParams", payable)]
fn contract_init(ctx: &impl HasInitContext<()>, amount: Amount) -> Result<State, InitError> {
```

### Adding `enable_logger`

- init/receive functions take an extra argument, which is the `logger: impl HasLogger` and it is positioned to the right of the amount if `payable` is also set otherwise to the right of the context.

```rust
#[init(contract = "two-step-transfer", parameter = "InitParams", payable, enable_logger)]
fn contract_init(ctx: &impl HasInitContext<()>, amount: Amount, logger: &impl HasLogger) -> Result<State, InitError> {
```

### DIFF
--- a/concordium-std-derive/src/lib.rs
+++ b/concordium-std-derive/src/lib.rs
@@ -30,12 +30,9 @@ fn get_attribute_value<'a, I: IntoIterator<Item = &'a Meta>>(
     })
 }
 
-// Return whether the low-level item is present.
-fn get_low_level<'a, I: IntoIterator<Item = &'a Meta>>(iter: I) -> bool {
-    iter.into_iter().any(|attr| match attr {
-        Meta::Path(path) => path.is_ident("low_level"),
-        _ => false,
-    })
+// Return whether a attribute item is present.
+fn contains_attribute<'a, I: IntoIterator<Item = &'a Meta>>(iter: I, name: &str) -> bool {
+    iter.into_iter().any(|attr| attr.path().is_ident(name))
 }
 
 /// Derive the appropriate export for an annotated init function.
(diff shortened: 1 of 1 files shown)


concordium-std-derive/src/lib.rs
```diff
@@ -81,14 +103,14 @@ pub fn init(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+                #setup_fn_optional_args
                 let ctx = ExternContext::<InitContextExtern>::open(());
-                let mut logger = Logger::init();
-                match #fn_name(&ctx, amount, &mut logger) {
+                match #fn_name(&ctx, #(#fn_optional_args),*) {
                     Ok(state) => {
                         let mut state_bytes = ContractState::open(());
                         if state.serial(&mut state_bytes).is_err() {
-                            trap() // Could not initialize contract.
+                            concordium_std::trap() // Could not initialize contract.
```
abizjak, line 113: I think either we don't import `trap`, or we use it here without qualifying.

concordium-std-derive/src/lib.rs
```diff
@@ -60,18 +57,43 @@ pub fn init(attr: TokenStream, item: TokenStream) -> TokenStream {
 
     let ast: syn::ItemFn = syn::parse(item).expect("Init can only be applied to functions.");
 
+    let mut setup_fn_optional_args = proc_macro2::TokenStream::new();
```
abizjak, line 60: Could you insert a paragraph above describing how this new section achieves what it does.

Concretely, that `setup_fn_optional_args` sets up the datastructures, and `fn_optional_args` are the arguments that will be added to the function call.

Moreover we should document what the relationship is between them.

concordium-std-derive/src/lib.rs
```diff
@@ -81,14 +103,14 @@ pub fn init(attr: TokenStream, item: TokenStream) -> TokenStream {
         quote! {
             #[export_name = #wasm_export_fn_name]
             pub extern "C" fn #rust_export_fn_name(amount: Amount) -> i32 {
-                use concordium_std::{Logger, trap};
+                use concordium_std::{trap, ExternContext, InitContextExtern, ContractState};
+                #setup_fn_optional_args
                 let ctx = ExternContext::<InitContextExtern>::open(());
-                let mut logger = Logger::init();
-                match #fn_name(&ctx, amount, &mut logger) {
+                match #fn_name(&ctx, #(#fn_optional_args),*) {
```
abizjak, line 109: To be helpful, we should check that the number of arguments the macro is applied to matches the number we will provide, and fail if this is not the case.

The error message could include the exact arguments we expect, which we do know at  this point.

concordium-std-derive/src/lib.rs
```diff
@@ -135,36 +157,61 @@ pub fn receive(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+        );
+    };
+
+    if contains_attribute(&attrs, "enable_logger") {
+        let logger_ident = format_ident!("logger");
+        setup_fn_optional_args.extend(
+            quote!(let mut #logger_ident = concordium_std::Logger::init();)
+        );
+        fn_optional_args.push(quote!(&mut #logger_ident));
+    }
```
abizjak, line 183: Isn't this very much the same as above for init. In which case, I think we should unify them in a function that returns two token streams.

abizjak (changes requested):

abizjak (approved):

PR #5, opened 2020-12-09
Fix the fail macro.

abizjak:
As it was before the macro generated the code with a `#[cfg(feature = ...)]
directive, which is both unhygienic, as well as buggy.


limemloh (approved):
Hard to find much about `core_intrinsics`, but looks good to me otherwise

PR #6, opened 2020-12-10
Improve error reporting when procedural macros are misused.

abizjak:
An example, forgotten amount in a function argument
```console
error: Incorrect number of function arguments, the expected arguments are (ctx: &impl HasReceiveContext, amount: Amount, state: &mut MyState) 
   --> src/lib.rs:140:5
    |
140 | /     _ctx: &impl HasReceiveContext,
141 | |     _state: &mut State,
    | |_______________________^

error: aborting due to previous error
```

This could be further refined to localize the error more precisely, but that can come in the future.

### DIFF
--- a/concordium-std-derive/Cargo.toml
+++ b/concordium-std-derive/Cargo.toml
@@ -11,7 +11,7 @@ readme = "../README.md"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [dependencies]
-syn = { version = "1.0", features = [ "full", "extra-traits" ] }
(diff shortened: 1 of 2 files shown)


limemloh (dismissed):
LGTM

We should consider refactoring most of `init` and `receive` logic into a function returning Result, to use `?` instead of `return`.

abizjak:
@limemloh I've reworked it now so that it uses, more or less consistently, the Result type for all the internal computations.

This has changed the semantics slightly in some error cases, which previously would have produced both compilation errors, as well as emitted code. Now they simply fail.

PR #12, opened 2020-12-21
Add expect_err_report and expect_none_report for Result and Option

limemloh:


concordium-std/src/impls.rs
```diff
@@ -556,6 +556,17 @@ impl<A, E: fmt::Debug> ExpectReport for Result<A, E> {
     }
 }
 
+impl<A, E: fmt::Debug> ExpectErrReport for Result<A, E> {
```
abizjak, line 559: I think what we want here is that A has Debug, and then in the error message for the Ok case display the value.

concordium-std/src/impls.rs
```diff
@@ -556,6 +556,17 @@ impl<A, E: fmt::Debug> ExpectReport for Result<A, E> {
...
+    type Unwrap = E;
+
+    fn expect_err_report(self, msg: &str) -> Self::Unwrap {
+        match self {
+            Ok(_) => crate::fail!("{}", msg),
```
abizjak, line 564: Here we do
```rust
Ok(v) => crate::fail!("{}: {:?}", msg, v)
```

concordium-std/src/impls.rs
abizjak: And we don't need Debug on E.

abizjak (changes requested):

abizjak (approved):

PR #16, opened 2021-01-21
Adding auction contract example

amaurremi:
If you have a chance I'd really appreciate feedback, especially on whether there are better/more conventional ways to express things in Rust.

### DIFF
--- a/examples/auction/src/lib.rs
+++ b/examples/auction/src/lib.rs
@@ -1,0 +1,349 @@
+use concordium_std::*;
+use std::collections::btree_map::BTreeMap;
+use core::fmt::Debug;
+
+// Implementation of an auction smart contract
+//
+// To bid, participants send GTU using the bid function.
+// The participant with the highest bid wins the auction.
+// Bids are to be placed before the auction end. After that, bids are refused.
+// Only bids that exceed the highest bid are accepted.
+// Bids are placed incrementally, i.e., an account's bid is considered
+// to be the **sum** of all bids.
+// Example: if Alice first bid 1 GTU and then bid 2 GTU, her total
+// bid is 3 GTU. The bidding will only go through if 3 GTU is higher than
+// the currently highest bid.
+//
+// After the auction end, any account can finalize the auction.
+// The auction can be finalized only once.
+// When the auction is finalized, every participant except the
+// winner gets their money back.
+
+#[contract_state(contract = "auction")]
+#[derive(Debug, Serialize, SchemaType, Eq, PartialEq)]
+pub struct State {
+    // Is the auction active?
+    auction_state: AuctionState,
+    // The highest bid so far (stored explicitly so that bidders can quickly see it)
+    highest_bid: Amount,
+    // The sold item (to be displayed to the auction participants), encoded in ASCII
+    item: Vec<u8>,
+    // Expiration time of the auction at which bids will be closed (to be displayed to the auction participants)
+    expiry: Timestamp,
+    // Keeping track of which account bid how much money
+    #[concordium(map_size_length = 2)]
+    bids: BTreeMap<AccountAddress, Amount>,
+}
+
+#[derive(Debug, Serialize, SchemaType, Eq, PartialEq, PartialOrd)]
+pub enum AuctionState {
+    Active,
+    BidsOverWaitingForAuctionFinalization,
+    Sold(AccountAddress), // winning account's address
+}
+
+fn fresh_state(itm: Vec<u8>, exp: Timestamp) -> State {
+    State {
+        auction_state: AuctionState::Active,
+        highest_bid: Amount::zero(),
+        item: itm,
+        expiry: exp,
+        bids: BTreeMap::new(),
+    }
+}
+
+#[derive(Serialize, SchemaType)]
+struct InitParameter {
+    item: Vec<u8>, // item to be sold as a sequence of ASCII codes
+    expiry: Timestamp, // time of the auction end
+}
+
+#[derive(Debug, PartialEq, Eq)]
+enum ReceiveError {
+    ContractSender, // raised if a contract, as opposed to account, tries to bid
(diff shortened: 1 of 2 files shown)


examples/auction/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+        let state_result = auction_init(&ctx);
+        let state = state_result.expect("Contract initialization results in error");
+        assert_eq!(state,
+                   dummy_fresh_state(),
+                   "Auction state should be new after initialization"
+        );
+    }
+
+    #[test]
+    fn test_auction_bid_and_finalize() {
```
limemloh, line 237: I would prefer if this was split into multiple unit tests, since it testing more than one thing, meaning if any of them failed it would be hard to know what exactly failed.

limemloh (changes requested):

examples/auction/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+fn auction_bid<A: HasActions>(
+    ctx: &impl HasReceiveContext,
+    amount: Amount,
+    state: &mut State,
+) -> Result<A, ReceiveError> {
+    ensure!(state.auction_state <= AuctionState::BidsOverWaitingForAuctionFinalization, ReceiveError::AuctionFinalized);
+
+    let slot_time = ctx.metadata().slot_time();
+    if slot_time > state.expiry {
+        state.auction_state = AuctionState::BidsOverWaitingForAuctionFinalization;
```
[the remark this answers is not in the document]
abizjak: This if statement would also be clearer expressed as
```rust
ensure!(slot_time <= state.expiry, ReceiveError::BidsOverWaitingForAuctionFinalization)`
```

examples/auction/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+        bail!(ReceiveError::BidsOverWaitingForAuctionFinalization);
+    }
+
+    let sender_address = match ctx.sender() {
+        Address::Contract(_) => bail!(ReceiveError::ContractSender),
+        Address::Account(account_address) => account_address,
+    };
+    let existing_bid = *state.bids.entry(sender_address).or_insert_with(|| Amount::zero());
```
abizjak, line 100: I would not dereference here, keeping

```rust
let existing_bid_ref = state.bids.entry(sender_address).or_insert_with(|| Amount::zero());
```

examples/auction/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    let new_bid = existing_bid + amount;
```
abizjak, line 102: Then here you can do 

```rust
new_bid = *existing_bid_ref + amount
```

examples/auction/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+    // Ensure that the new bid exceeds the highest bid so far
+    ensure!(new_bid > state.highest_bid, ReceiveError::BidTooLow { bidded: amount, highest_bid: state.highest_bid });
+    state.highest_bid = new_bid;
+    state.bids.insert(sender_address, existing_bid + amount);
```
abizjak, line 106: And here you can just do 
```rust
existing_bid_ref += amount
```

examples/auction/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+    let owner = ctx.owner();
+
+    ensure!(state.auction_state == AuctionState::Active, FinalizeError::AuctionFinalized);
+
+    let balance = ctx.self_balance();
+    if balance == Amount::zero() {
+        Ok(A::accept())
+    } else {
+        let mut return_action = A::simple_transfer(&owner, state.highest_bid);
+        let mut remaining_bids = BTreeMap::new();
```
abizjak, line 132: I would not use a map here since it can have exactly one element.

I think the better thing to do would be to initialize with `None` and then ...

examples/auction/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+        // Return bids that are smaller than highest
+        for (addr, &amnt) in state.bids.iter() {
+            if amnt < state.highest_bid {
+                return_action = return_action.and_then(A::simple_transfer(addr, amnt));
+            } else {
+                state.auction_state = AuctionState::Sold(*addr);
+                remaining_bids.insert(addr, amnt);
```
abizjak, line 139: Here set to `Some (addr, amnt)`

abizjak: You could raise an error if this is already set to `Some`.

abizjak: Alternatively just do
```rust
return_action = return_action.and_then(A::simple_transfer(owner, amnt)
```

examples/auction/src/lib.rs
abizjak: I disagree with this. I like this test. It is testing a lot of failed and successful scenarios in sequence, also testing that things can be done after failed attempts.

Moreover there are concrete error messages attached to each test, which will make it easy to identify them when they fail.

examples/auction/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+        highest_bid: Amount,
+    ) {
+        let res: Result<ActionsTree, _> = auction_bid(ctx, amount, &mut state);
+        res.expect("Bidding should pass");
+        bid_map.insert(account, highest_bid);
+        assert_eq!(*state, dummy_state(highest_bid, bid_map.clone()));
+    }
+
+    #[test]
+    fn test_auction_bid_repeated_bid() {
```
abizjak, line 313: It would be useful to add a line or two briefly summarizing the test (the same applies to other tests).

abizjak (commented):
I think this is very good.

If you could address the minor issues commented on this is a good example of a smart contract to have as an example.

abizjak (approved):

PR #17, opened 2021-02-18
Add a wrapper for contract state testing with low-level contracts.

abizjak:


concordium-std/src/test_infrastructure.rs
```diff
@@ -530,3 +533,222 @@ pub fn report_error(message: &str, filename: &str, line: u32, column: u32) {
...
+/// A borrowed instantiation of `ContractStateTest`.
+pub type ContractStateTestBorrowed<'a> = ContractStateTest<&'a mut Vec<u8>>;
+
+/// An owned variant that can be more convenient for testing since the type
+/// itself owns the data.
+pub type ContractStateTestOwned = ContractStateTest<Vec<u8>>;
+
+#[derive(Debug, PartialEq, Eq)]
+// An error that is raised when operating with `Seek`, `Write`, or `Read` trait
+// methods of the `ContractStateTest` type.
```
limemloh, line 552: Should probably be a doc comment
```suggestion
/// An error that is raised when operating with `Seek`, `Write`, or `Read` trait
/// methods of the `ContractStateTest` type.
```

concordium-std/src/test_infrastructure.rs
```diff
@@ -530,3 +533,222 @@ pub fn report_error(message: &str, filename: &str, line: u32, column: u32) {
...
+                    Ok(x)
+                } else {
+                    Err(Offset)
+                }
+            }
+            SeekFrom::End(x) => {
+                // cannot seek beyond end, nor before beginning
+                if x <= 0 {
+                    let end: u32 = self.cursor.data.as_ref().len().try_into()?;
+                    let minus_x = x.checked_abs().ok_or(Overflow)?;
```
limemloh, line 650: When reading the documentation for [`std::io::Seek`](https://doc.rust-lang.org/std/io/trait.Seek.html), it seems to be *allowed* to seek beyond the end.
Not sure how 'allowed' should be interpreted, whether it means: Results in Ok.

limemloh: The documentation for [SeekFrom](https://doc.rust-lang.org/std/io/enum.SeekFrom.html#variant.End) adds a bit more to this.

limemloh (dismissed):
Looks good, but I have a few comments related to Seek

concordium-std/src/test_infrastructure.rs
abizjak: I know, but this is our Seek trait, and our implementation. It makes it cleaner to not allow this because then we have the invariant that offset is a valid place in the state.

abizjak: And also, it states `A seek beyond the end of a stream is allowed, but behavior is defined by the implementation.`

So we can choose to say "Err" is my interpretation of it.

limemloh (approved):

PR #22, opened 2021-03-01
Returning error codes from init and receive functions

amaurremi:
Note: the changes in the auction smart contract are hard to see because I did a reformatting. But the only "real" changes are that the custom error types derive `Reject` (like in the other example contracts), and that it removes fields from the `BidTooLow` error, because we don't support enums with fields yet.

### DIFF
--- a/concordium-std-derive/src/lib.rs
+++ b/concordium-std-derive/src/lib.rs
@@ -7,8 +7,7 @@ extern crate quote;
 use proc_macro::TokenStream;
 use proc_macro2::Span;
 use quote::ToTokens;
-use std::ops::Neg;
-use std::convert::TryFrom;
+use std::{convert::TryFrom, ops::Neg};
 use syn::{
     parse::Parser, parse_macro_input, punctuated::*, spanned::Spanned, DataEnum, Ident, Meta, Token,
 };
@@ -1264,11 +1263,18 @@ fn reject_derive_worker(input: TokenStream) -> syn::Result<TokenStream> {
     }?;
     let enum_ident = &ast.ident;
 
-    // Ensure that the number of enum variants fits into the number of error codes we can generate.
-    let too_many_variants = format!("Error enum {} cannot have more than {} variants.", enum_ident, RESERVED_ERROR_CODES.neg());
+    // Ensure that the number of enum variants fits into the number of error codes
+    // we can generate.
+    let too_many_variants = format!(
+        "Error enum {} cannot have more than {} variants.",
+        enum_ident,
+        RESERVED_ERROR_CODES.neg()
+    );
     match i32::try_from(enum_data.variants.len()) {
         Ok(n) if n <= RESERVED_ERROR_CODES.neg() => (),
-        _ => Err(syn::Error::new(ast.span(), &too_many_variants))?
+        _ => {
+            return Err(syn::Error::new(ast.span(), &too_many_variants));
+        }
     };
 
     let variant_error_conversions: Vec<_> =
@@ -1276,6 +1282,7 @@ fn reject_derive_worker(input: TokenStream) -> syn::Result<TokenStream> {
 
     let gen = quote! {
         impl From<#enum_ident> for Reject {
+            #[inline]
             fn from(e: #enum_ident) -> Self {
                 Reject { error_code: (e as u32 + 1) }
             }
@@ -1365,6 +1372,7 @@ fn from_error_token_stream(
         .map(|from_error| {
             quote! {
             impl From<#from_error> for #enum_name {
+               #[inline]
                fn from(fe: #from_error) -> Self {
                  #enum_name::#variant_name
                }
--- a/concordium-std-derive/src/lib.rs
+++ b/concordium-std-derive/src/lib.rs
@@ -209,7 +209,7 @@ fn init_worker(attr: TokenStream, item: TokenStream) -> syn::Result<TokenStream>
                         let code = Reject::from(reject).error_code;
                         let with_reserved_offset = code.checked_add(#RESERVED_ERROR_CODES).expect(#exceeded_error_code_limit);
                         -i32::try_from(with_reserved_offset).expect(#exceeded_error_code_limit)
-                    } // TODO (MRA) check that error code is in bound
+                    }
                 }
             }
         }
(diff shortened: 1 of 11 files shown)


concordium-std-derive/src/lib.rs
```diff
@@ -193,12 +195,16 @@ fn init_worker(attr: TokenStream, item: TokenStream) -> syn::Result<TokenStream>
             #[export_name = #wasm_export_fn_name]
             pub extern "C" fn #rust_export_fn_name(#amount_ident: Amount) -> i32 {
                 use concordium_std::{trap, ExternContext, InitContextExtern, ContractState};
+                use std::convert::TryFrom;
```
abizjak, line 198: We can't use `std` directly since `concordium-std` is meant to work with `#[no_std]`.

We usually handle this by re-exporting the relevant definition from `concordium-std`. I think `convert` is already exported, so probably `concordium_std::convert::TryFrom` should work.

abizjak:
> Note: the changes in the auction smart contract are hard to see because I did a reformatting. But the only "real" changes are that the custom error types derive `Reject` (like in the other example contracts), and that it removes fields from the `BidTooLow` error, because we don't support enums with fields yet.

Could you also add the auction contract to the CI here:
- https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/.github/workflows/linter.yml#L24
- https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/.github/workflows/linter.yml#L204

to avoid such issues in the future.

concordium-std-derive/src/lib.rs
```diff
@@ -1202,6 +1222,57 @@ fn schema_type_fields(fields: &syn::Fields) -> syn::Result<proc_macro2::TokenStr
...
+    let enum_name = &ast.ident;
+    let exceeded_error_code_limit = format!(
+        "Error code should not exceed {} (i32::MAX minus reserved error-code offset).",
+        i32::MAX - RESERVED_ERROR_CODES as i32
+    );
+
+    let gen = quote! {
+        impl From<#enum_name> for Reject {
+            fn from(e: #enum_name) -> Self {
+                Reject { error_code: (e as u32).checked_add(#RESERVED_ERROR_CODES).expect(#exceeded_error_code_limit) }
```
abizjak, line 1269: I think for fieldless enums this is a good (and cheap, which is important) option.

For enums with fields I think we'll want to generate an explicit match statement so we can handle sub-enums better as we discussed yesterday.

We need to do some experimentation to see what is most ergonomic, and least error prone for the user.

abizjak (commented):
I think this is on the right track.

It needs some polish, but even in this form it will provide a much better experience on the chain than the "no information" we had before.

amaurremi:
It looks like we currently won't able to annotate enum fields with `#[from = ErrorType]` attributes (see https://github.com/rust-lang/rust/issues/53012). I'm instead going to annotate the whole enum as `#[derive(..., Reject(MyEnumField -> ParseError))]` or something of the like.

amaurremi:
On a second look, derive macro helper attributes are a thing--wonderful.

concordium-std-derive/src/lib.rs
```diff
@@ -1202,6 +1224,154 @@ fn schema_type_fields(fields: &syn::Fields) -> syn::Result<proc_macro2::TokenStr
     }
 }
 
+/// We reserve a number of error codes for custom errors, such as ParseError,
+/// that are provided by concordium-std. The error codes for user-defined error
+/// types will have this number as an offset.
+const RESERVED_ERROR_CODES: u32 = 100;
```
abizjak, line 1230: I think it is going to be better if the "reserved" codes are not -1, -2, ... but that we start `i32::MIN, i32::MIN+1, ...`

This way user defined codes are going to be less scary, and more direct, and things that are kind of "system errors" will have weird looking codes.

concordium-std-derive/src/lib.rs
```diff
@@ -1202,6 +1224,154 @@ fn schema_type_fields(fields: &syn::Fields) -> syn::Result<proc_macro2::TokenStr
...
+
+/// Derive the conversion of enums that represent error types into the Reject
+/// struct which can be used as the error type of init and receive functions.
+/// Creating custom enums for error types can provide meaningful error messages
+/// to the user of the smart contract.
+///
+/// Please note:
+///   - At the moment, we can only derive fieldless enums.
+///   - The error-type enum needs to derive (or implement) the Copy and Clone
+///     traits.
```
abizjak, line 1240: I'm guessing it should be possible to not require Copy. Practically it is not a problem, since fieldless enums are easily copyable, but if we can avoid requiring it we can avoid annoying errors for the users.

concordium-std-derive/src/lib.rs
```diff
@@ -1202,6 +1224,154 @@ fn schema_type_fields(fields: &syn::Fields) -> syn::Result<proc_macro2::TokenStr
...
+
+    let exceeded_error_code_limit = format!(
+        "Error code should not exceed {} (i32::MAX minus reserved error-code offset).",
+        i32::MAX - RESERVED_ERROR_CODES as i32
+    );
+
+    let gen = quote! {
+        impl From<#enum_ident> for Reject {
+            fn from(e: #enum_ident) -> Self {
+                Reject { error_code: (e as u32 + 1).checked_add(#RESERVED_ERROR_CODES).expect(#exceeded_error_code_limit) }
```
abizjak, line 1279: If we make the reserved codes as I suggested then we can avoid this wart here with adding an offset.

abizjak: Maybe it would be worthwhile to also add an `#[inline]` annotation on the generated function.

We should test though what, if any, effect it has on the generated size of the `.wasm` file.

concordium-std-derive/src/lib.rs
amaurremi: I think that's a good idea, will do.

concordium-std-derive/src/lib.rs
amaurremi: For the `error_code` smart contract, it didn't have any effect. Both produced 1,104 bytes.

abizjak (approved):

abizjak (approved):

PR #23, opened 2021-03-06
A well placed hint as to how to call this function

gilescope:


limemloh:
Thank you for taking time to help us!

But I will close this PR, because the comment is not true in the general case.
It is not a bad idea to mention this, but it should be a note next to a more precise doc-string for this perticular function.

PR #25, opened 2021-03-09
CB-1203: Ergonomic send

Bargsteen:
Closes CB-1203.

Makes `HasActions::send` more ergonomic:

- Change the `receive_name` parameter of `HasActions::send` to use `ReceiveName`
  instead of `str`.
- Rename `send` to `send_raw`
- Rename `log_bytes` to `log_raw`
- Add `send`, a wrapper for `HasActions::send_raw`, which automatically
  serializes `parameter` (using `Serial`).

Related:
https://gitlab.com/Concordium/smart-contracts/-/merge_requests/125
https://gitlab.com/Concordium/concordium-client/-/merge_requests/291
https://github.com/Concordium/concordium-contracts-common/pull/14

### DIFF
--- a/concordium-std/CHANGELOG.md
+++ b/concordium-std/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Change the `receive_name` parameter of `HasActions::send` to use `ReceiveName`
   instead of `str`.
 - Rename `send` to `send_raw` in `HasActions`.
+- Rename `log_bytes` to `log_raw` in `HasLogger`.
 - Add `send`, a wrapper for `HasActions::send_raw`, which automatically
   serializes `parameter` (using `Serial`).
 - Allow init and receive methods to return custom error codes that will be displayed to the user
--- a/concordium-std/CHANGELOG.md
+++ b/concordium-std/CHANGELOG.md
@@ -8,6 +8,7 @@
   reverts the change in concordium-std 0.4.1.
 - Change the `receive_name` parameter of `HasActions::send` to use `ReceiveName`
   instead of `str`.
+- Rename `send` to `send_raw` in `HasActions`.
 - Add `simple_send`, a wrapper for `HasActions::send`, which automatically
   serializes `parameter` (using `Serial`).
 - Allow init and receive methods to return custom error codes that will be displayed to the user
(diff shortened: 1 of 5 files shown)


concordium-std/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
   to be consistent with the Write implementation for ContractState.
 - Use little-endian encoding for sender contract addresses in receive contexts. This
   reverts the change in concordium-std 0.4.1.
+- Change the `receive_name` parameter of `HasActions::send` to use `ReceiveName`
+  instead of `str`.
+- Add `simple_send`, a wrapper for `HasActions::send`, which automatically
```
abizjak, line 11: I think we need a better name than `simple_send` so as to not confuse people (there is `simple_transfer` already)

concordium-std/src/impls.rs
```diff
@@ -470,8 +470,13 @@ impl HasActions for Action {
     }
 
     #[inline(always)]
-    fn send(ca: &ContractAddress, receive_name: &str, amount: Amount, parameter: &[u8]) -> Self {
-        let receive_bytes = receive_name.as_bytes();
+    fn send(
+        ca: &ContractAddress,
+        receive_name: &ReceiveName,
```
abizjak, line 475: Since `ReceiveName` is already a reference I don't think we should pass a reference to that here.

concordium-std/src/impls.rs
```diff
@@ -521,6 +526,18 @@ pub fn put_in_memory(input: &[u8]) -> *mut u8 {
     ptr
 }
 
+/// Wrapper for HasActions::Send, which automatically serializes the parameter.
+#[inline(always)]
```
abizjak, line 530: I'm not sure that the inline here is a good idea.

We'd need to do some measurements to see whether it makes sense.

concordium-std/src/impls.rs
```diff
@@ -521,6 +526,18 @@ pub fn put_in_memory(input: &[u8]) -> *mut u8 {
...
+pub fn simple_send<A: HasActions, P: Serial>(
+    ca: &ContractAddress,
+    receive_name: &ReceiveName,
+    amount: Amount,
+    parameter: &P,
+) -> A {
+    let param_bytes = to_bytes(parameter);
+    HasActions::send(ca, receive_name, amount, &param_bytes)
```
abizjak, line 538: I think this should be `A::send` instead, shouldn't it?

abizjak (changes requested):
Apart from the minor comments this looks good..

abizjak (dismissed):

PR #26, opened 2021-03-22
Make derived exports a bit more hygienic

jakobbotsch:
We had an ambiguous type error here due to a user-named type called `Amount`.


abizjak:
Thanks, looks good.

But before we accept external contributions we'll need you to accept a CLA. We are setting up the templates and everything should be ready early next week. I apologize for the delay.

jakobbotsch:
No problem, just let me know when and how :)

abizjak:
> No problem, just let me know when and how :)

@jakobbotsch if you could rebase this branch on top of main, and update the description to be in-line with the following template
https://raw.githubusercontent.com/Concordium/.github/main/.github/pull_request_template.md then we're good to go.

PR #35, opened 2021-06-11
Validate contract and receive names

Bargsteen:
## Purpose

Validate contract and receive names in attributes for the init and receive functions.
Improves the precision in the error locations shown.

## Changes

- Validate contract and receive names.
- Make `concordium-std-derive` depend on `concordium-contracts-common` to use the common validation functions
- Change the signature of `get_attribute_value` to return a `LitStr` so that we can get the correct `Span`.
- Correct the `Span` for errors with missing fields. (It now underlines the whole attribute list instead of just the first field)
- Handle new error cases (from https://github.com/Concordium/concordium-contracts-common/pull/19)

## Checklist

      hard-to-understand areas.


concordium-std-derive/src/lib.rs
```diff
@@ -12,6 +12,8 @@ use syn::{
     parse::Parser, parse_macro_input, punctuated::*, spanned::Spanned, DataEnum, Ident, Meta, Token,
 };
 
+use concordium_contracts_common::*;
+
```
abizjak, line 16: No empty lines before the import, and use fmt to order them.

concordium-std-derive/src/lib.rs
```diff
@@ -179,7 +181,11 @@ fn init_worker(attr: TokenStream, item: TokenStream) -> syn::Result<TokenStream>
 
     let fn_name = &ast.sig.ident;
     let rust_export_fn_name = format_ident!("export_{}", fn_name);
-    let wasm_export_fn_name = format!("init_{}", contract_name);
+    let wasm_export_fn_name = format!("init_{}", contract_name.value());
+
+    ContractName::is_valid_contract_name(&wasm_export_fn_name)
+        .map_err(|e| syn::Error::new(contract_name.span(), e.to_string()))?;
```
abizjak, line 187: I think this would be easier to understand if you used an `if`, e.g., 
```rust
if let Err(e) = ContractName::is_valid... {
  return syn::Error::new(contract_name.span(), e.to_string())
}
```

concordium-std-derive/src/lib.rs
```diff
@@ -376,7 +382,20 @@ fn receive_worker(attr: TokenStream, item: TokenStream) -> syn::Result<TokenStre
 
     let fn_name = &ast.sig.ident;
     let rust_export_fn_name = format_ident!("export_{}", fn_name);
-    let wasm_export_fn_name = format!("{}.{}", contract_name, name);
+    let wasm_export_fn_name = format!("{}.{}", contract_name.value(), name.value());
```
abizjak, line 385: Is there ever a case where you don't use `.value()` on the result of `get_attribute_...`?

If so, I think it is a bad idea to change the return type as you did.

concordium-std/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased changes
+- Handle new cases for NewContractNameError and NewReceiveNameError
```
abizjak, line 4: This needs to be a bit clearer. As it is it really doesn't say anything to the users.

abizjak (changes requested):

concordium-std-derive/src/lib.rs
Bargsteen: No, there isn't. But I made the change so that we could use `span()` on the `LitStr` and thereby show the correct position of the error.

concordium-std-derive/Cargo.toml
```diff
@@ -15,6 +15,11 @@ syn = { version = "1.0.63", features = [ "full", "extra-traits" ] }
 quote = "=1.0.0"
 proc-macro2 = "1.0"
 
+[dependencies.concordium-contracts-common]
+path = "../concordium-contracts-common"
+version = "=0.4"
```
abizjak, line 20: Don't you mean at least `0.4.1` here?

concordium-std-derive/src/lib.rs
```diff
@@ -376,7 +382,27 @@ fn receive_worker(attr: TokenStream, item: TokenStream) -> syn::Result<TokenStre
...
+
+    // Validate the contract name independently to ensure that it doesn't contain a
+    // '.' as this causes a subtle error when receive names are being split.
+    let contract_name_validation =
+        ContractName::is_valid_contract_name(&format!("init_{}", contract_name.value()))
+            .map_err(|e| syn::Error::new(contract_name.span(), e.to_string()));
```
abizjak, line 391: This is minor, but you can just remove `.to_string()` since `::new` takes any value that implements `Display`, and thus calls `to_string` automatically.

concordium-std-derive/src/lib.rs
```diff
@@ -376,7 +382,27 @@ fn receive_worker(attr: TokenStream, item: TokenStream) -> syn::Result<TokenStre
...
+
+    let receive_name_validation = ReceiveName::is_valid_receive_name(&wasm_export_fn_name)
+        .map_err(|e| syn::Error::new(name.span(), e.to_string()));
```
abizjak, line 394: And here.

abizjak (changes requested):

abizjak (approved):

PR #36, opened 2021-06-14
Enable renaming of enum variants and field names

Bargsteen:
## Purpose

Enable renaming of enum variants and field names.
This allows SC writers to use long and descriptive names while coding, but short names on-chain.

You simply annotate each field or variant with `#[concordium(rename = "<name>")]`:
``` rust
enum Model {
    #[concordium(rename = "V")]
    SomeLongVariantName {
        #[concordium(rename = "a")]
        some_long_field_name_a: i32,

        #[concordium(rename = "b")]
        some_long_field_name_b: u32,
    },
}
```

## Changes

- Add `"rename"` to the list of valid concordium attributes
- Replace the variant/field name if another one is provided with the `rename` attribute
- Checks for conflicting definitions caused by the renaming

## Checklist

      hard-to-understand areas.


concordium-std-derive/CHANGELOG.md
```diff
@@ -6,3 +6,4 @@
 
 - Add macros for deriving error implementations.
 - Make derive macros slightly more hygienic.
+- Enable renaming of enum variants and field names.
```
limemloh, line 9: You should probably also mention that this is in regard to schemas.

concordium-std-derive/src/lib.rs
```diff
@@ -1144,7 +1150,20 @@ fn schema_type_derive_worker(input: TokenStream) -> syn::Result<TokenStream> {
...
+                            syn::Lit::Str(renamed_variant_name) => renamed_variant_name.value(),
+                            _ => return Err(syn::Error::new(
+                                renamed_variant_name.span(),
+                                "Rename attribute value must be a string.",
+                            )),
+                        }
+                    } else {
+                        variant.ident.to_string()
+                    };
+
```
limemloh, line 1166: We should also check for name collisions, which I imagine will be possible now.

limemloh (changes requested):
I would love to see the addition of checking for name collisions

concordium-std-derive/CHANGELOG.md
Bargsteen: Done.

concordium-std-derive/src/lib.rs
Bargsteen: Ah yes :) Done.

concordium-std-derive/src/lib.rs
```diff
@@ -1210,10 +1254,48 @@ fn schema_type_field_type(field: &syn::Field) -> syn::Result<proc_macro2::TokenS
...
+                            _ => {
+                                return Err(syn::Error::new(
+                                    renamed_field_name.span(),
+                                    "Rename attribute value must be a string.",
+                                ))
+                            }
+                        }
+                    } else {
+                        (field.ident.clone().unwrap().to_string(), field.ident.span())
+                        // safe since
```
limemloh, line 1277: Seems like a typo or something

limemloh (approved):
Only one minor thing, looks good to me otherwise

concordium-std-derive/src/lib.rs
```diff
@@ -1210,10 +1254,48 @@ fn schema_type_field_type(field: &syn::Field) -> syn::Result<proc_macro2::TokenS
...
+                            }
+                            _ => {
+                                return Err(syn::Error::new(
+                                    renamed_field_name.span(),
+                                    "Rename attribute value must be a string.",
+                                ))
+                            }
+                        }
+                    } else {
+                        (field.ident.clone().unwrap().to_string(), field.ident.span())
```
abizjak, line 1276: Please leave the comment as to why it is safe to unwrap.

concordium-std-derive/src/lib.rs
```diff
@@ -1210,10 +1254,48 @@ fn schema_type_field_type(field: &syn::Field) -> syn::Result<proc_macro2::TokenS
...
+                            syn::Lit::Str(renamed_field_name) => {
+                                (renamed_field_name.value(), renamed_field_name.span())
+                            }
+                            _ => {
+                                return Err(syn::Error::new(
+                                    renamed_field_name.span(),
+                                    "Rename attribute value must be a string.",
+                                ))
+                            }
+                        }
```
abizjak, line 1274: I believe this part finding the rename attribute and checking whether it is the right type could be factored into a helper function, so that it is not duplicated here and in the `type_derive_worker` function.

concordium-std-derive/src/lib.rs
```diff
@@ -1210,10 +1254,48 @@ fn schema_type_field_type(field: &syn::Field) -> syn::Result<proc_macro2::TokenS
...
+                        );
+                        let error_at_second_def = syn::Error::new(
+                            field_span,
+                            format!("the name `{}` is defined multiple times", field_name),
+                        );
+                        // Combine the errors to show both at once
+                        error_at_first_def.combine(error_at_second_def);
+
+                        return Err(error_at_first_def);
+                    }
```
abizjak, line 1297: The same goes for this part for checking duplicates. The error messages look the same.

abizjak (changes requested):

abizjak (approved):

concordium-std-derive/CHANGELOG.md
```diff
@@ -8,3 +8,4 @@
 
 - Add macros for deriving error implementations.
 - Make derive macros slightly more hygienic.
+- Enable renaming of enum variants and field names for the schema.
```
abizjak, line 11: It would be useful to add a bit more detail here, specifically what the new features are (rename attribute, ...)

abizjak (approved):

PR #39, opened 2021-06-17
Ensure ordered

Bargsteen:
## Purpose

Allow the use of `ensure_ordered` without also setting the `size_length`.
Remove the different versions of `size_length` attributes (`map_`, `set_`, `string_`). So now you only have `size_length`.

## Changes

- Add two new traits `SerialCtx` and `DeserialCtx`, which act as wrappers for `Serial` and `Deserial`, but with the contextual information of `size_length` and `ensure_ordered`.
- Simplify `impl_serial_field` and `impl_deserial_field` by using the new (de)serial_ctx functions.
- Remove the different versions of the `size_length` attributes.
- I also switched to the correct `span` for the `size_length` attribute (not directly related to purpose).

## Checklist

      hard-to-understand areas.
- [X] (If necessary) I have updated the CHANGELOG.

### DIFF
--- a/concordium-std/src/impls.rs
+++ b/concordium-std/src/impls.rs
@@ -680,73 +680,85 @@ impl<A: fmt::Debug> ExpectNoneReport for Option<A> {
     }
 }
 
-// fn serial_as_size_length<W: Write>(
-//     len: usize,
-//     size_length: schema::SizeLength,
-//     out: &mut W,
-// ) -> Result<(), W::Err> {
-//     match size_length {
-//         schema::SizeLength::U8 => (len as u8).serial(out),
-//         schema::SizeLength::U16 => (len as u16).serial(out),
-//         schema::SizeLength::U32 => (len as u32).serial(out),
-//         schema::SizeLength::U64 => (len as u64).serial(out),
-//     }
-// }
-
-fn deserial_length(source: &mut impl Read, size_length: u32) -> ParseResult<usize> {
+fn deserial_length(source: &mut impl Read, size_length: schema::SizeLength) -> ParseResult<usize> {
     use core::convert::TryInto;
     let res = match size_length {
-        1 => u8::deserial(source)?.into(),
-        2 => u16::deserial(source)?.into(),
-        4 => u32::deserial(source)?.try_into().map_err(|_| ParseError::default())?,
-        8 => u64::deserial(source)?.try_into().map_err(|_| ParseError::default())?,
-        _ => return Err(Default::default()),
+        schema::SizeLength::U8 => u8::deserial(source)?.into(),
+        schema::SizeLength::U16 => u16::deserial(source)?.into(),
+        schema::SizeLength::U32 => {
+            u32::deserial(source)?.try_into().map_err(|_| ParseError::default())?
+        }
+        schema::SizeLength::U64 => {
+            u64::deserial(source)?.try_into().map_err(|_| ParseError::default())?
+        }
     };
     Ok(res)
 }
 
-fn serial_length<W: Write>(len: usize, size_length: u32, out: &mut W) -> Result<(), W::Err> {
+fn serial_length<W: Write>(
+    len: usize,
+    size_length: schema::SizeLength,
+    out: &mut W,
+) -> Result<(), W::Err> {
     match size_length {
-        1 => (len as u8).serial(out)?,
-        2 => (len as u16).serial(out)?,
-        4 => (len as u32).serial(out)?,
-        8 => (len as u64).serial(out)?,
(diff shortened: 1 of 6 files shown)


concordium-std/src/impls.rs
```diff
@@ -1,7 +1,19 @@
 use crate::{convert, mem, num, prims, prims::*, traits::*, types::*};
+#[cfg(not(feature = "std"))]
+use alloc::{
+    collections::{BTreeMap, BTreeSet},
+    string::String,
+    vec::Vec,
+};
```
abizjak, line 7: This should not be here we already have conditional re-exports in lib.rs, so there is no need for the conditional imports here. These can just be added to `use crate:: ...`.

In general conditional imports should be limited as much as possible to lib.rs.

concordium-std/src/impls.rs
```diff
@@ -1,7 +1,19 @@
...
 use concordium_contracts_common::*;
-
+#[cfg(not(feature = "std"))]
+use core::convert::{TryFrom, TryInto};
 use mem::MaybeUninit;
+#[cfg(feature = "std")]
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    convert::{TryFrom, TryInto},
+};
```
abizjak, line 16: Same here. `convert` is already imported in `lib.rs`, so use that instead.

concordium-std/src/impls.rs
```diff
@@ -678,3 +685,111 @@ impl<A: fmt::Debug> ExpectNoneReport for Option<A> {
         }
     }
 }
+
+impl<K: Serial + Ord> SerialCtx for BTreeSet<K> {
+    fn serial_ctx<W: Write>(
+        &self,
+        size_len: &schema::SizeLength,
```
abizjak, line 692: Since `SizeLength` is a small type it should be taken by value, not by reference.

concordium-std/src/traits.rs
```diff
@@ -220,3 +220,26 @@ pub trait ExpectNoneReport {
     /// [fail](macro.fail.html) with the given message, instead of `panic`.
     fn expect_none_report(self, msg: &str);
 }
+
+/// A wrapper trait for 'Serial' with additional contextual information, namely
```
abizjak, line 224: What does a "wrapper trait" mean?

Additionally, you could explain the "size_length property" in more understandable terms, as, e.g., the number of bytes to use to record the length of the data.

concordium-std/src/traits.rs
```diff
@@ -220,3 +220,26 @@ pub trait ExpectNoneReport {
...
+    /// A wrapper for 'Serial::serial' which uses the `size_length` property.
+    fn serial_ctx<W: Write>(
+        &self,
+        size_length: &schema::SizeLength,
+        out: &mut W,
+    ) -> Result<(), W::Err>;
+}
+
+/// A wrapper trait for 'Deserial' with additional contextual information,
+/// namely the size_length and ensure_ordered properties.
```
abizjak, line 236: Make the comment more understandable here as well.

abizjak (changes requested):

examples/two-step-transfer/src/lib.rs
```diff
@@ -89,7 +89,7 @@ pub struct State {
     // Requests which have not been dropped due to timing out or due to being agreed to yet
     // The request ID, the associated amount, when it times out, who is making the transfer and
     // which account holders support this transfer
-    #[concordium(map_size_length = 2)]
+    #[concordium(size_length = 2)]
```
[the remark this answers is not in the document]
Bargsteen: https://github.com/Concordium/concordium-wasm-smart-contracts/pull/8

concordium-std/src/traits.rs
```diff
@@ -220,3 +220,43 @@ pub trait ExpectNoneReport {
...
+        &self,
+        size_length: schema::SizeLength,
+        out: &mut W,
+    ) -> Result<(), W::Err>;
+}
+
+/// A wrapper trait for 'Deserial' with additional contextual information,
+/// namely the size_length and ensure_ordered properties.
+/// The `DeserialCtx` trait provides a means of reading structures from
+/// byte-sinks (`Read`) using contextual information.
```
abizjak, line 247: I suppose this should be byte-sources, not sinks.

abizjak (dismissed):

concordium-std/src/traits.rs
Bargsteen: Ah okay. I also corrected it in the `Deserial` trait documentation.

abizjak (dismissed):

abizjak (approved):

PR #43, opened 2021-07-08
Add a simple contract to support the memo feature.

### LINKED ISSUE #33: Address clippy warnings about using `panic!` without a string literal.
**Task description**

In some cases for the test infrastructure we use the `panic!` macro in `no_std` mode with a message that is not a string literal, but it is a string formatted using `format`.  This causes clippy to complain and might become an error in the future. We should try to fix this.

abizjak:
## Purpose

Add an example contract that can support a "memo" feature.

Also fixes #33 

## Changes

Add a new example smart contract.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/memo/src/lib.rs
+++ b/examples/memo/src/lib.rs
@@ -3,6 +3,13 @@ use core::fmt::Debug;
 
 /// # Implementation of a smart contract that can receive transfers with a memo
 /// message and forward them to the owner account.
+/// All this contract does is expose a single `receive` method which checks that
+///
+/// - it is being invoked by an account
+/// - the message it is receiving is 32 bytes long
+///
+/// And if both of these are valid it forwards the amount it received to the
+/// owner account.
 
 /// The contract has no state.
 #[derive(Debug, Serialize, SchemaType)]
--- a/examples/memo/src/lib.rs
+++ b/examples/memo/src/lib.rs
@@ -1,5 +1,4 @@
 use concordium_std::*;
-use core::fmt::Debug;
 
 /// # Implementation of a smart contract that can receive transfers with a memo
 /// message and forward them to the owner account.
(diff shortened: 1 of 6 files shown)


examples/memo/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+///
+/// And if both of these are valid it forwards the amount it received to the
+/// owner account.
+
+#[derive(Serialize, SchemaType)]
+/// The contract has no initialization parameters.
+struct InitParameter;
+
+/// Init function that creates a new contract.
+#[init(contract = "memo", parameter = "InitParameter", low_level)]
```
limemloh, line 18: You could save a few bytes by not adding the parameter to the schema, assuming you are embedding the schema.

examples/memo/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+
+/// Receive a transaction with a message. This ensures that the message is 32
+/// bytes and that the sender of the message is an account.
+#[receive(contract = "memo", name = "receive", parameter = "ReceiveParameter", payable, low_level)]
+fn memo_receive<A: HasActions>(
+    ctx: &impl HasReceiveContext,
+    amount: Amount,
+    _state: &mut impl HasContractState,
+) -> ReceiveResult<A> {
+    matches!(ctx.sender(), Address::Account(..));
```
limemloh, line 35: I think you want to use `ensure!` around matches, otherwise this should not do anything

limemloh (changes requested):

examples/memo/src/lib.rs
abizjak: Yes, but I wanted to add it for symmetry.

examples/memo/src/lib.rs
abizjak: Indeed, silly me.

limemloh (approved):

PR #57, opened 2021-08-17
Concordium Token Standard (CIS1)

limemloh:
First see https://github.com/Concordium/concordium-rust-smart-contracts/pull/58

## Purpose

Define standard CIS1 for smart contract tokens on the concordium blockchain and implement some basic examples of common token types in Rust.

## Changes

- Add specification of CIS1.
- Add `concordium-cis1` crate with common types and functions used to implement a CIS1 token.
- Add ERC20-like example of CIS1 (`cis1-wgtu`).
- Add ERC721-like example of CIS1 (`cis1-nft`).
- Add ERC1155-like example of CIS1 (`cis1-multi`).

## Checklist

      hard-to-understand areas.


CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+
+No token level approval/allowance like in ERC20 and ERC721
+----------------------------------------------------------
+
+This standard only specifies address-level operators and no authentication on per token level.
+The main argument is simplicity and to save energy cost on common cases, but other reasons are:
+
+- A token level authentication requires the token smart contract to track more state, which increases the overall energy cost.
+- For token smart contracts with a lot of token types, such as a smart contract with a large collection of NFTs, a token level authentication could become very expensive.
+- For fungible tokens; approval/allowance introduces an attack vector as `described here<https://docs.google.com/document/d/1YLPtQxZu1UAvO9cZ1O2RPXBbT0mooh4DYKjA_jp-RLM/edit>`.
```
abizjak, line 514: A google docs link should not be in the standard. Is there no published paper or PDF or similar?

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+
+.. note::
+
+  The specification does not prevent adding more fine-grained authentication, such as a token level authentication.
+
+Receive hook function
+---------------------
+
+The specification requires a token receive hook to be called on a smart contract receiving tokens, this will in some cases prevent mistakes such as sending tokens to smart contracts, which do not defined behavior for receiving tokens.
```
abizjak, line 523:
```suggestion
The specification requires a token receive hook to be called on a smart contract receiving tokens, this will in some cases prevent mistakes such as sending tokens to smart contracts which do not define behavior for receiving tokens.
```

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+These token could then be lost forever.
+
+The reason for this not being optional is to allow other smart contracts which integrates with a token smart contract to rely on this for functionality.
```
abizjak, line 526:
```suggestion
The reason for this not being optional is to allow other smart contracts which integrate with a token smart contract to rely on this for functionality.
```

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+
+.. warning::
+
+  The smart contract receive hook function could be called by any smart contract and it is up to the integrating contract whether to trust the token contract.
+
+Receive hook function callback argument
+---------------------------------------
+
+The name of the receive hook function called on a smart contract receiving tokens is supplied as part of the parameter.
+This allows for a smart contract to integrating with a token smart contract to have multiple hooks and leave it to the caller to know which hook they want to trigger.
```
abizjak, line 537:
```suggestion
This allows for a smart contract integrating with a token smart contract to have multiple hooks and leave it to the caller to know which hook they want to trigger.
```

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+  The specification does not prevent adding more fine-grained authentication, such as a token level authentication.
+
+Receive hook function
+---------------------
+
+The specification requires a token receive hook to be called on a smart contract receiving tokens, this will in some cases prevent mistakes such as sending tokens to smart contracts, which do not defined behavior for receiving tokens.
+These token could then be lost forever.
+
+The reason for this not being optional is to allow other smart contracts which integrates with a token smart contract to rely on this for functionality.
+An auction smart contract could take bids by token transfers directly.
```
abizjak, line 527: I'm not sure why the auction smart contract suddenly appears.

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+.. warning::
+
+  The smart contract receive hook function could be called by any smart contract and it is up to the integrating contract whether to trust the token contract.
+
+Receive hook function callback argument
+---------------------------------------
+
+The name of the receive hook function called on a smart contract receiving tokens is supplied as part of the parameter.
+This allows for a smart contract to integrating with a token smart contract to have multiple hooks and leave it to the caller to know which hook they want to trigger.
+An auction smart contract could receive the item to auction using one hook and bids on another hook.
```
abizjak, line 538: Same as above, don't really understand the references to auction here.

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+-----------------------
+
+The FA2 token standard found on Tezos, allows for a hook function to be called on a smart contract sending tokens, such that the contract could reject the transfer on some criteria.
+This seems to only make sense, if some operator is transferring tokens from a contract, in which case the sender smart contract might as well contain the logic to transfer the tokens and trigger this directly.
+
+Explicit events for mint and burn
+---------------------------------
+
+In ERC20, ERC721 and ERC1155 they use a transfer event from or to the zero address to indicate mint and burn respectively, but since there are no such thing as the zero address on the Concordium blockchain these events are separate.
+Making it more explicit, instead of special case transfer events.
```
abizjak, line 552:
```suggestion
Making it more explicit instead of special case transfer events.
```

examples/cts1-multi/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies]
+concordium-std = {path = "../../concordium-std", default-features = false}
+concordium-cts1 = {path = "../../concordium-cts1", default-features = false}
+
+[lib]
+crate-type=["cdylib", "rlib"]
+
+[profile.release]
+opt-level = "z"
```
abizjak, line 20: Have you tried `codegen-units=1` as we discussed? Does it improve the size?

examples/cts1-nft/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies]
+concordium-std = {path = "../../concordium-std", default-features = false}
+concordium-cts1 = {path = "../../concordium-cts1", default-features = false}
+
+[lib]
+crate-type=["cdylib", "rlib"]
+
+[profile.release]
+opt-level = "s"
```
abizjak, line 20: Is `s` here better than `z`? Also have you tried codegen-units=1?

examples/cts1-multi/src/lib.rs
```diff
@@ -0,0 +1,783 @@
...
+//! An instance of this smart contract can contain a number of different token
+//! types each identified by a token ID. A token type is then globally
+//! identified by the contract address together with the token ID.
+//!
+//! In this example the contract is initialized with no tokens, and tokens can
+//! be minted through a `mint` contract function, which will only succeed for
+//! the contract owner. No functionality to burn token is defined in this
+//! example.
+//!
+//! Note: The word 'address' is referring to either an account address or a
```
abizjak, line 13:
```suggestion
//! Note: The word 'address' is refers to either an account address or a
```

abizjak (changes requested):

abizjak (changes requested):
I did not go through the example contracts in detail yet, but at a glance they look good.

I also have a general comment about the description of the standard. I think serializations are hard to read as they are. I quite liked the way they specify serializations in the Wasm spec https://webassembly.github.io/spec/core/binary/index.html which would mean roughly having a section defining the types involved and specifying their serialization (from primitive to combined types) and then using the notation introduced in this section whenever you need to specify serialization.

.github/workflows/linter.yml
```diff
@@ -208,6 +214,9 @@ jobs:
           - examples/piggy-bank/part2/Cargo.toml
           - examples/auction/Cargo.toml
           - examples/erc721/Cargo.toml
+          - examples/cts1-wgtu/Cargo.toml
+          - examples/cts1-nft/Cargo.toml
+          - examples/cts1-multi/Cargo.toml
```
abizjak, line 219: Also, please add these examples to the examples/README.md file.

limemloh: Already there

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+- A transfer of some amount of a token type MUST only transfer the exact amount of tokens between balances.
+- A transfer of any amount of a token type to a contract address MUST call receive hook function on the receiving smart contract with a receive hook parameter :ref:`described above<CTS-functions-transfer-receive-hook-parameter>`
+- The contract function MUST reject if a receive hook function called on the contract receiving tokens rejects.
+
+.. _CTS-functions-updateOperator:
+
+``updateOperator``
+^^^^^^^^^^^^^^^^^^
+
+Add or remove an address as operator of the address sending this message.
```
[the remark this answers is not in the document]
limemloh: I am not sure what you mean by it being a two-step process.
Could you elaborate?

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+
+Requirements
+~~~~~~~~~~~~
+
+- The contract function MUST reject if the sender address is the same as the operator address with error :ref:`OPERATOR_IS_SENDER<CTS-rejection-errors>`.
+
+.. note::
+
+  Operators are not set per token ID, and an operator can control any token type of the owner address.
+  This was chosen to require less on the contract implementation and also simplify off-chain integration.
```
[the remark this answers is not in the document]
limemloh: I was reasoning why the operators are not scoped to a per token type.
But I will move this to the rationale section in the end of this document instead.

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+The parameter contains whether to add or remove an operator and the address to add/remove as operator.
+It does not contain the address which are adding/removing the operator as this will be the sender address of the transaction invoking this function.
+
+The parameter is first a byte indicating whether to remove or add an operator, where if the byte is 0 the sender is removing an operator, if the byte is 1 the sender is adding an operator.
+The followed is the operator address :ref:`CTS-Address` to add or remove as operator for the sender.
+
+Requirements
+~~~~~~~~~~~~
+
+- The contract function MUST reject if the sender address is the same as the operator address with error :ref:`OPERATOR_IS_SENDER<CTS-rejection-errors>`.
```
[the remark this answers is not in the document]
limemloh: I guess there is not really a need for this requirement, since being an operator of your own address does not grant you any new permissions

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+
+Logged events
+-------------
+
+The idea of the logged events for this specification is for off-chain applications to be able to track balances and operators without knowledge of the contract specific implementation details.
+For this reason it is important to log events in any custom functionality for the token contract, if these modifies balances or operators.
+
+It MUST be safe for off-chain applications to assume a contract implementing this specification and no events logged have zero tokens and no operators enabled for any address.
+
+.. Other events custom to the contract implementation MUST be safe for the off-application to ignore.
```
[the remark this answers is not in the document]
limemloh: This is a RST comment ment for myself, and it should have been removed.

I'll put a requirement for accumulating the balance from the CTS events only, instead of something like this.

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+The event to log for a transfer of some amount of a token type.
+A contract function which transfers tokens MUST log a transfer event for each of these transfers.
+
+The Transfer event is serialized as: first a byte with the value of 0, followed by the token ID :ref:`CTS-TokenID`, an amount of tokens :ref:`CTS-TokenAmount`, from address :ref:`CTS-Address` and to address :ref:`CTS-Address`.
+
+Mint
+^^^^
+
+An event for minting MUST be logged every time a new token is minted. This also applies when introducing new token types and the initial token types and amounts in a contract.
+Minting a token with a zero amount is valid.
```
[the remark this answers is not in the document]
limemloh: No, the point (which I'll add) is that the event of minting a token type with amount 0 is something you might do to indicate a token type exists, but no amount of token exists yet.

An example of this can be seen in the wrapped GTU example, where initially no amount of wGTU exists, but the token type still does.

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+.. note::
+
+  Be aware of the limit on the number of logs per smart contract function call which currently is 64.
+  A token smart contract function which needs to mint a large number of token types with token metadata might hit this limit.
+
+Burn
+^^^^
+
+An event for burning MUST be logged every time an amount of tokens are burned.
+Burning a zero amount of a token is allowed.
```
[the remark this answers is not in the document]
limemloh: I'll remove the sentence, as I don't have a case for when to burn a zero amount, and this sentence does not provide anything useful.

CTS1.rst
limemloh: This is the link they also use in ERC20 for the describing the attack

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+    - string
+    - An image URI to a large image for displaying the asset.
+  * - ``artifact`` (optional)
+    - URI JSON object
+    - A URI to the token asset.
+  * - ``assets`` (optional)
+    - JSON array of Token metadata JSON objects
+    - Collection of assets.
+  * - ``attributes`` (optional)
+    - JSON array of Attribute JSON objects
```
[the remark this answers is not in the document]
limemloh: It is very common for NFTs to use attributes for adding some extra information besides the actual artifact of the NFT.
On a NFT marketplace they would just be listed next to the artifact.

An example use case could be a Sword NFT for a game, here the attributes could be something like damage, attack speed and weight.

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+    - JSON value type [JSON-Schema]
+    - Description
+  * - ``name`` (optional)
+    - string
+    - The name to display for the token type.
+  * - ``symbol`` (optional)
+    - string
+    - Short text to display for the token type.
+  * - ``decimals`` (optional)
+    - number [``integer``]
```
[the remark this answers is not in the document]
limemloh: No, a negative number is fine, I'll add a bit more on how to use this integer for displaying an amount.

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+An amount of a token type.
+It is represented in WASM as an unsigned integer ``i64`` (8 bytes little endian).
+
+.. _CTS-ReceiveHookName:
+
+``ReceiveHookName``
+^^^^^^^^^^^^^^^^^^^^^^^
+
+A smart contract receive function name.
+It is serialized as: the function name byte length is represented by the first 2 bytes, followed this many bytes for the function name.
```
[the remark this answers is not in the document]
limemloh: I have added the extra requirements

limemloh:
> I think serializations are hard to read as they are.

I have tried to make it more readable by using a format inspired by the one on Wasm

examples/cts1-nft/Cargo.toml
limemloh: yes, "s" is the better option here. 

I have just tried setting codegen-units to 1 and it saves another 1.6 kB

examples/cts1-multi/Cargo.toml
limemloh: I have added codegen-units since it reduces the size

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+  Token IDs can be as small as a single byte (by setting the first byte to the value 0) or as big as 256 bytes leaving more than 10^614 possible token IDs.
+  The token ID could be an encoding of a small text string or some checksum hash, but to save energy it is still recommended to use small token IDs if possible.
+
+.. _CTS-TokenAmount:
+
+``TokenAmount``
+^^^^^^^^^^^^^^^
+
+An amount of a token type.
+It is represented in WASM as an unsigned integer ``i64`` (8 bytes little endian).
```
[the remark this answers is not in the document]
limemloh: This type is never really exposed as a Wasm integer, so I have deleted the wasm type.

CTS1.rst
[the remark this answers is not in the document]
abizjak: I meant that if I (the owner) want to change the operator from A to B I have to do two transactions right?

I am not saying that is wrong, just asking.

CTS1.rst
abizjak: I see, the issue is that it is an editable document.

CTS1.rst
```diff
@@ -0,0 +1,666 @@
...
+
+A smart contract implementing CTS1 MUST export three functions :ref:`CTS-functions-transfer`, :ref:`CTS-functions-updateOperator` and :ref:`CTS-functions-balanceOf` according to the following description:
+
+.. _CTS-functions-transfer:
+
+``transfer``
+^^^^^^^^^^^^
+
+Executes a list of token transfers.
+A transfer is a token ID, an amount of this token ID to be transferred from one address to some other address.
```
abizjak, line 169: This sentence does not read very well. Maybe be clearer "A transfer is a token ID, an amount of token to transfer, and the from and to addresses."

CTS1.rst
```diff
@@ -0,0 +1,666 @@
...
+
+When transferring tokens to a contract address additional information for a receive function hook to trigger is required.
+
+Parameter
+~~~~~~~~~
+
+The parameter is a list of transfers.
+
+It is serialized as: 1 byte representing the number of transfers (``n``) followed by the bytes for this number of transfers (``transfers``).
```
abizjak, line 178: I can't remember whether we discussed this, but is it not perhaps too limiting to mandate 1 byte for length?

If we relax the limits on the parameter size in the future this might be unnecessary.

CTS1.rst
```diff
@@ -0,0 +1,666 @@
...
+``updateOperator``
+^^^^^^^^^^^^^^^^^^
+
+Add or remove an address as operator of the address sending this message.
+
+Parameter
+~~~~~~~~~
+
+The parameter contains whether to add or remove an operator and the address to add/remove as operator.
+It does not contain the address which are adding/removing the operator as this will be the sender of the message invoking this function.
```
abizjak, line 230:
```suggestion
It does not contain the address which is adding/removing the operator as this will be the sender of the message invoking this function.
```

CTS1.rst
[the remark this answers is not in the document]
abizjak: So will you remove this, or do you want to keep it?

Does this appear in other standards?

CTS1.rst
```diff
@@ -0,0 +1,666 @@
...
+~~~~~~~~~~~~
+
+- The contract function MUST reject if the sender address is the same as the operator address with error :ref:`OPERATOR_IS_SENDER<CTS-rejection-errors>`.
+
+.. _CTS-functions-balanceOf:
+
+``balanceOf``
+^^^^^^^^^^^^^
+
+Query balances of a list of addresses and token IDs, the result is then send back to the sender.
```
abizjak, line 250:
```suggestion
Query balances of a list of addresses and token IDs, the result is then sent back to the sender.
```

CTS1.rst
```diff
@@ -0,0 +1,666 @@
...
+
+Decisions and rationale
+=======================
+
+In this section we point out some of the differences from other popular token standards found on other blockchains, and provide reasons for deviating from them in CTS1.
+
+Token ID bytes instead an integer
+---------------------------------
+
+Token standards such as ERC721 and ERC1155 both uses an 256 bit unsigned integer (32 bytes) for the token ID, to support using something like a SHA256 hash for the token ID.
```
abizjak, line 590:
```suggestion
Token standards such as ERC721 and ERC1155 both use an 256 bit unsigned integer (32 bytes) for the token ID, to support using something like a SHA256 hash for the token ID.
```

CTS1.rst
```diff
@@ -0,0 +1,666 @@
...
+But in the case where the token ID have no significance other than a simple identifier, smaller sized token IDs can reduce energy costs.
+This is why we chose to let the first byte indicate the size of the token ID, meaning a token ID can vary between 1 byte and 256 bytes, resulting in more than 10^614 possible token IDs.
+
+Only batched transfers
+----------------------
+
+The specification only has a ``transfer`` smart contract function which takes list of transfer and no function for a single transfer.
+This will result in lower energy cost compared to multiple contract calls and only introduce a small overhead for single transfers.
```
abizjak, line 598:
```suggestion
This will result in lower energy cost compared to multiple contract calls and only introduces a small overhead for single transfers.
```
```

File: CTS1.rst [hunk 28]
```diff
@@ -0,0 +1,666 @@
...
+The reason for not also including a single transfer function, is to have smaller smart contract modules, which in turn leads to saving cost on every function call.
```
abizjak, line 599:
```suggestion
The reason for not also including a single transfer function is to have smaller smart contract modules, which in turn leads to saving cost on every function call.
```

CTS1.rst
```diff
@@ -0,0 +1,666 @@
...
+No sender hook function
+-----------------------
+
+The FA2 token standard found on Tezos, allows for a hook function to be called on a smart contract sending tokens, such that the contract could reject the transfer on some criteria.
+This seems to only make sense, if some operator is transferring tokens from a contract, in which case the sender smart contract might as well contain the logic to transfer the tokens and trigger this directly.
+
+Explicit events for mint and burn
+---------------------------------
+
+In ERC20, ERC721 and ERC1155 they use a transfer event from or to the zero address to indicate mint and burn respectively, but since there are no such thing as the zero address on the Concordium blockchain these events are separate.
```
abizjak, line 653:
```suggestion
ERC20, ERC721 and ERC1155 use a transfer event from or to the zero address to indicate mint and burn respectively, but since there are no such thing as the zero address on the Concordium blockchain these events are separate.
```

CTS1.rst
```diff
@@ -0,0 +1,666 @@
...
+Making it more explicit instead of special case transfer events.
+
+No error code for receive hook rejecting
+----------------------------------------
+
+The specification could include an error code, for the receive hook function to return if rejecting the token transferred (as seen in the FA2 standard on Tezos).
```
abizjak, line 659: Insert a link?

concordium-std/src/impls.rs
```diff
@@ -800,6 +800,17 @@ impl<T: Serial> SerialCtx for &[T] {
...
 }
 
+impl<T: Serial> SerialCtx for Vec<T> {
+    fn serial_ctx<W: Write>(
+        &self,
+        size_len: schema::SizeLength,
+        out: &mut W,
+    ) -> Result<(), W::Err> {
+        let slice: &[T] = &self;
+        slice.serial_ctx(size_len, out)
```
abizjak, line 810:
```suggestion
        self.as_slice().serial_ctx(size_len, out)
```
and remove the `let slice = ...`

CTS1.rst
limemloh: I don't think we have.

But for your information, the smallest possible transfer is 46 bytes, meaning the parameter size would have to be more than 11.7kB to be able to hit the limit of having one byte.

CTS1.rst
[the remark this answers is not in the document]
limemloh: We could have this be a list of updates instead. Which is what tezos' FA2 do as well.
The only reason I did not add this is to keep the parameter more simple.

CTS1.rst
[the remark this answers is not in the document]
limemloh: I'll remove this

CTS1.rst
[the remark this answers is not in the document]
limemloh: I'll make the parameter take a list of updates

CTS1.rst
limemloh: Also since every transfer will result in an event being logged, the 64 log per contract function will also put a limit on this.

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+
+  This type is passed in a parameter for smart contract function calls, be aware of the parameter size limit of 1024 bytes.
+
+.. _CTS-ContractName:
+
+``ContractName``
+^^^^^^^^^^^^^^^^
+
+A smart contract name.
+It is serialized as: the contract name byte length is represented by the first 2 bytes, followed this many bytes for the contract name.
```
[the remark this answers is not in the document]
limemloh: I have added the extra requirements

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+The UpdateOperator event is serialized as: first a byte with the value of 3, followed by a byte which is 0 if an operator is being removed and 1 if an operator is being added, then the owner address updating an operator :ref:`CTS-Address` and an operator address :ref:`CTS-Address` being added or removed.
+
+TokenMetadata
+^^^^^^^^^^^^^
+
+The event to log when setting the metadata url for a token type.
+It consists of a token ID and an URL for the location of the metadata for this token type with an optional SHA256 checksum of the content.
+Logging the TokenMetadata event again with the same token ID, is used to update the metadata location and only the most recently logged token metadata event for certain token id should be used to get the token metadata.
+
+The TokenMetadata event is serialized as: first a byte with the value of 4, followed by the token ID :ref:`CTS-TokenID`, two bytes for the length of the metadata url and then this many bytes for the url to the metadata.
```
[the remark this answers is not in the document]
limemloh: I have added an RFC

CTS1.rst
limemloh: I have removed this

CTS1.rst
```diff
@@ -0,0 +1,564 @@
...
+- The contract function MUST reject if any of the queries fail.
+- A query MUST fail if the token ID is unknown with error: :ref:`INVALID_TOKEN_ID<CTS-rejection-errors>`.
+
+Logged events
+-------------
+
+The idea of the logged events for this specification is for off-chain applications to be able to track balances and operators without knowledge of the contract specific implementation details.
+For this reason it is important to log events in any custom functionality for the token contract, if these modifies balances or operators.
+
+It MUST be safe for off-chain applications to assume a contract implementing this specification and no events logged have zero tokens and no operators enabled for any address.
```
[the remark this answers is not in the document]
limemloh: I have tried to rephrase this

Bargsteen (commented):
Great work!! :+1:

concordium-cts1/src/lib.rs
```diff
@@ -0,0 +1,747 @@
...
+    /// The address receiving the tokens being transferred.
+    pub to:       Receiver,
+    /// Additional data to include in the transfer.
+    /// Can be used for additional arguments.
+    pub data:     AdditionalData,
+}
+
+/// The parameter type for the contract function `transfer`.
+#[derive(Debug, Serialize, SchemaType)]
+pub struct TransferParams<T: IsTokenId>(#[concordium(size_length = 1)] pub Vec<Transfer<T>>);
```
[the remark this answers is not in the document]
limemloh: I have changed the schemaType for transferParams

CTS1.rst
```diff
@@ -0,0 +1,666 @@
...
+The contract name is not allowed to contain ``.``.
+
+It is serialized as: the contract name byte length (``n``) is represented by the first 2 bytes, followed by this many bytes for the contract name (``name``).
+The contract name MUST be 100 bytes or less::
+
+  ContractName ::= (n: Byte²) (name: Byteⁿ)
+
+.. _CTS-AccountAddress:
+
+``AccountAddress``
```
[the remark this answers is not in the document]
limemloh: I have added a warning in the section describing the `transfer` contract function

concordium-cts1/src/lib.rs
```diff
@@ -0,0 +1,780 @@
...
+    fn get_type() -> schema::Type {
+        schema::Type::List(schema::SizeLength::U8, Box::new(schema::Type::U8))
+    }
+}
+
+/// The `TokenIdUnit` is serialized with one byte with the value 0.
+impl Serial for TokenIdUnit {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        out.write_u8(0)?;
+        Ok(())
```
abizjak, line 413: Why not just `out.write_u8(0)`?

concordium-cts1/src/lib.rs
```diff
@@ -0,0 +1,780 @@
...
+    /// Sender is neither the token owner or an operator of the owner for this
+    /// token.
+    Unauthorized,
+    /// Only contracts can send to this function.
+    ContractOnly,
+    /// Custom error
+    Custom(R),
+}
+
+impl<R: Into<Reject>> From<Cts1Error<R>> for Reject {
```
abizjak, line 536: Document error codes that are assigned in a doc comment.

concordium-cts1/src/lib.rs
```diff
@@ -0,0 +1,780 @@
...
+            Cts1Error::Custom(reject) => reject.into().error_code,
+        };
+        Self {
+            error_code,
+        }
+    }
+}
+
+impl<X: From<LogError>> From<LogError> for Cts1Error<X> {
+    fn from(err: LogError) -> Self { Cts1Error::Custom(X::from(err)) }
```
abizjak, line 556: Suggest `#[inline]`

concordium-cts1/src/lib.rs
```diff
@@ -0,0 +1,780 @@
...
+}
+
+impl<X: From<ParseError>> From<ParseError> for Cts1Error<X> {
+    fn from(err: ParseError) -> Self { Cts1Error::Custom(X::from(err)) }
```
abizjak, line 560: Suggest `#[inline]`

abizjak (commented):
I guess the CTS1.md should be removed before this is merged, but other than that I only have minor comments.

abizjak (approved):

abizjak (approved):

PR #61, opened 2021-09-25
Ensure that there are no redundant attributes in annotations, and no duplicates.

### LINKED ISSUE #40: Only allow valid attributes for `#[init]` and `#[receive]`
**Description**
Currently, the procedural macros `#[init]` and `#[receive]` takes a number of attributes, (`contract`, `name`, `parameter` ...), but there are no check for the user to only provide defined attributes.

Say, a user mistypes `parameter`, since this won't produce any compile time error, the user will have to notice the missing function in the schema.

abizjak:
## Purpose

Closes #40 

## Changes

There should be no changes for valid uses of the macros.

In case of incorrect use compilation errors will be issued.

Screenshot from 2021-09-27 10-10-14

There are some remaining improvements that we could do.

Currently if you have something like
```rust
#init(parameter = "foo", parameter)
```
the error will be that the second parameter is "unrecognized", but it could be argued that a more descriptive error would be that parameter is already set.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std-derive/src/lib.rs
+++ b/concordium-std-derive/src/lib.rs
@@ -88,7 +88,7 @@ impl ParsedAttributes {
         self.values.remove(&key)
     }
 
-    /// Remove an attribute return whether it was present.
+    /// Remove an attribute and return whether it was present.
     pub(crate) fn extract_flag(&mut self, key: &str) -> bool {
         // This is not clean, constructing a new identifier with a call_site span.
         // But the only alternative I see is iterating over the map and locating the key
@@ -399,14 +399,14 @@ pub fn init(attr: TokenStream, item: TokenStream) -> TokenStream {
 }
 
 fn init_worker(attr: TokenStream, item: TokenStream) -> syn::Result<TokenStream> {
+    let ast: syn::ItemFn =
+        attach_error(syn::parse(item), "#[init] can only be applied to functions.")?;
+
     let attrs = Punctuated::<Meta, Token![,]>::parse_terminated.parse(attr)?;
 
     let init_attributes = parse_init_attributes(&attrs)?;
 
     let contract_name = init_attributes.contract;
-
-    let ast: syn::ItemFn =
-        attach_error(syn::parse(item), "#[init] can only be applied to functions.")?;
 
     let fn_name = &ast.sig.ident;
     let rust_export_fn_name = format_ident!("export_{}", fn_name);
@@ -592,10 +592,10 @@ pub fn receive(attr: TokenStream, item: TokenStream) -> TokenStream {
 }
 
 fn receive_worker(attr: TokenStream, item: TokenStream) -> syn::Result<TokenStream> {
-    let attrs = Punctuated::<Meta, Token![,]>::parse_terminated.parse(attr)?;
-
     let ast: syn::ItemFn =
         attach_error(syn::parse(item), "#[receive] can only be applied to functions.")?;
+
+    let attrs = Punctuated::<Meta, Token![,]>::parse_terminated.parse(attr)?;
 
     let receive_attributes = parse_receive_attributes(&attrs)?;
 
--- a/concordium-std-derive/src/lib.rs
+++ b/concordium-std-derive/src/lib.rs
@@ -10,7 +10,11 @@ use proc_macro2::Span;
 use quote::ToTokens;
 #[cfg(feature = "build-schema")]
 use std::collections::HashMap;
-use std::{convert::TryFrom, ops::Neg};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    convert::TryFrom,
+    ops::Neg,
+};
 use syn::{
     parse::Parser, parse_macro_input, punctuated::*, spanned::Spanned, DataEnum, Ident, Meta, Token,
 };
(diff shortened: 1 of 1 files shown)


concordium-std-derive/src/lib.rs
```diff
@@ -33,47 +37,275 @@ fn attach_error<A>(mut v: syn::Result<A>, msg: &str) -> syn::Result<A> {
...
+) -> syn::Result<InitAttributes> {
+    let mut attributes = parse_attributes(attrs)?;
+    let contract: syn::LitStr = attributes.extract_value("contract").ok_or_else(|| {
+        syn::Error::new(
+            Span::call_site(),
+            "A name for the contract must be provided, using the 'contract' attribute.\n\nFor \
+             example, #[init(contract = \"my-contract\")]",
+        )
+    })?;
+    let parameter: Option<syn::LitStr> = attributes.extract_value("parameter");
```
limemloh, line 247: Consider moving the string literals for attribute names (`"parameter"`, `"payable"`, `"enable_logger"` and so on) into `const`s.

limemloh (dismissed):
Great work

concordium-std-derive/src/lib.rs
```diff
@@ -33,47 +37,275 @@ fn attach_error<A>(mut v: syn::Result<A>, msg: &str) -> syn::Result<A> {
...
+    /// Rust identifier, otherwise this function will panic.
+    pub(crate) fn extract_value(&mut self, key: &str) -> Option<syn::LitStr> {
+        // This is not clean, constructing a new identifier with a call_site span.
+        // But the only alternative I see is iterating over the map and locating the key
+        // since Ident implements equality comparison with &str.
+        let key = syn::Ident::new(key, Span::call_site());
+        self.values.remove(&key)
+    }
+
+    /// Remove an attribute return whether it was present.
```
Bargsteen, line 91:
```suggestion
    /// Remove an attribute and return whether it was present.
```

concordium-std-derive/src/lib.rs
```diff
@@ -33,47 +37,275 @@ fn attach_error<A>(mut v: syn::Result<A>, msg: &str) -> syn::Result<A> {
...
+        // This is not clean, constructing a new identifier with a call_site span.
+        // But the only alternative I see is iterating over the map and locating the key
+        // since Ident implements equality comparison with &str.
+        let key = syn::Ident::new(key, Span::call_site());
+        self.flags.remove(&key)
+    }
+
+    /// If there are any remaining attributes signal an error. Otherwise return
+    /// Ok(())
+    pub(crate) fn report_all_attributes(self) -> syn::Result<()> {
```
Bargsteen, line 102: Consider renaming it to `report_unrecognized_attributes` (or similar) to clarify its purpose.

concordium-std-derive/src/lib.rs
```diff
@@ -169,13 +401,9 @@ pub fn init(attr: TokenStream, item: TokenStream) -> TokenStream {
...
-            "A name for the contract must be provided, using the contract attribute. For example, \
-             #[init(contract = \"my-contract\")]",
-        )
-    })?;
+    let init_attributes = parse_init_attributes(&attrs)?;
+
+    let contract_name = init_attributes.contract;
 
     let ast: syn::ItemFn =
         attach_error(syn::parse(item), "#[init] can only be applied to functions.")?;
```
Bargsteen, line 409: Consider moving this above the parsing of attributes to avoid unneeded computations and to clear up the error messages.

Bargsteen: Similar to what is done in `receive_worker`.

Bargsteen (dismissed):
Looks great! Just a few minor findings.

concordium-std-derive/src/lib.rs
abizjak: That was in fact the original name of this function :) but then I changed it since the current name is more descriptive.

The `report_unrecognized_attributes` makes sense in the context this is used.

So I would prefer to keep it.

concordium-std-derive/src/lib.rs
abizjak: To be honest I don't really see the advantage of this. What is your reasoning?

concordium-std-derive/src/lib.rs
abizjak: Done.

concordium-std-derive/src/lib.rs
abizjak: Done.

concordium-std-derive/src/lib.rs
limemloh: They appear in multiple places and moving them to a const would make the compiler help ensuring the spelling is the same.

Another minor advantage is that it is easier to ensure we cover every parameter, since they would be defined in the same place, and we would get warnings from unused const.

Bargsteen (dismissed):

concordium-std-derive/src/lib.rs
abizjak: I don't really agree with this reasoning but I have made this change since I want to get this merged.

In my view the code is now harder to read due to an indirection.

Bargsteen (approved):

PR #63, opened 2021-10-18
Update to CIS-1 draft 3: TokenMetadata query and no ContractOnly error

limemloh:
## Purpose

Update cis1 library and examples with the changes in https://github.com/Concordium/concordium-update-proposals/pull/14.

## Changes

For `concordium-cis1`
- Introduce parameter type for new `tokenMetadata` query
- Extend `balanceOf` parameter with contract address for receiving the result.

For CIS-1 examples:
- Send query results to the contract address from the parameter.
- Add CIS1 contract function `tokenMetadata`.
- Remove check for sender being a contract address for queries.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/cis1-multi/src/lib.rs
+++ b/examples/cis1-multi/src/lib.rs
@@ -428,6 +428,36 @@ fn contract_balance_of<A: HasActions>(
     ))
 }
 
+/// Takes a list of queries. Each query is an owner address and some address to
+/// check as an operator of the owner address. It takes a contract address plus
+/// contract function to invoke with the result.
+///
+/// It rejects if:
+/// - It fails to parse the parameter.
+/// - Message sent back with the result rejects.
+#[receive(contract = "CIS1-Multi", name = "operatorOf", parameter = "OperatorOfQueryParams")]
+fn contract_operator_of<A: HasActions>(
+    ctx: &impl HasReceiveContext,
+    state: &mut State,
+) -> ContractResult<A> {
+    // Parse the parameter.
+    let params: OperatorOfQueryParams = ctx.parameter_cursor().get()?;
+    // Build the response.
+    let mut response = Vec::with_capacity(params.queries.len());
+    for query in params.queries {
+        // Query the state for address being an operator of owner.
+        let is_operator = state.is_operator(&query.owner, &query.address);
+        response.push((query, is_operator));
+    }
+    // Send back the response.
+    Ok(send(
+        &params.result_contract,
+        params.result_function.as_ref(),
+        Amount::zero(),
+        &OperatorOfQueryResponse::from(response),
+    ))
+}
+
 /// Parameter type for the CIS-1 function `tokenMetadata` specialized to the
 /// subset of TokenIDs used by this contract.
 type ContractTokenMetadataQueryParams = TokenMetadataQueryParams<ContractTokenId>;
--- a/examples/cis1-multi/src/lib.rs
+++ b/examples/cis1-multi/src/lib.rs
@@ -90,6 +90,8 @@ enum CustomContractError {
     LogMalformed,
     /// Invalid contract name.
     InvalidContractName,
+    /// Only a smart contract can call this function.
+    ContractOnly,
 }
 
 type ContractError = Cis1Error<CustomContractError>;
@@ -128,6 +130,10 @@ impl State {
         *owner_balance += amount;
     }
 
+    /// Check that the token ID currently exists in this contract.
+    #[inline(always)]
+    fn contains_token(&self, token_id: &ContractTokenId) -> bool { self.tokens.contains(token_id) }
+
     /// Get the current balance of a given token id for a given address.
     /// Results in an error if the token id does not exist in the state.
     fn balance(
@@ -135,7 +141,7 @@ impl State {
         token_id: &ContractTokenId,
         address: &Address,
     ) -> ContractResult<TokenAmount> {
-        ensure!(self.tokens.contains(token_id), ContractError::InvalidTokenId);
+        ensure!(self.contains_token(token_id), ContractError::InvalidTokenId);
         let balance = self
             .state
             .get(address)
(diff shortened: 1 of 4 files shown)


examples/cis1-multi/src/lib.rs
```diff
@@ -382,26 +394,21 @@ fn contract_update_operator<A: HasActions>(
     Ok(A::accept())
 }
 
-/// Get the balance of a given token id for a given address and callback
-/// contract function on sender with the result.
-/// Will only succeed if called a smart contracts.
+#[allow(dead_code)]
+type ContractBalanceOfQueryParams = BalanceOfQueryParams<ContractTokenId>;
```
abizjak, line 398: Why is the allow(dead_code) needed? If it is only needed for the schema it would still be better to make it `pub` and avoid the need for `allow(dead_code)`.

abizjak: Also please document the type.

examples/cis1-multi/src/lib.rs
```diff
@@ -413,13 +420,55 @@ fn contract_balance_of<A: HasActions>(
...
-        params.callback.as_ref(),
+        &params.result_contract,
+        params.result_function.as_ref(),
         Amount::zero(),
         &BalanceOfQueryResponse::from(response),
     ))
 }
 
+#[allow(dead_code)]
+type ContractTokenMetadataQueryParams = TokenMetadataQueryParams<ContractTokenId>;
```
abizjak, line 431: Same as above.

examples/cis1-nft/src/lib.rs
```diff
@@ -381,26 +393,21 @@ fn contract_update_operator<A: HasActions>(
     Ok(A::accept())
 }
 
-/// Get the balance of given token IDs and addresses and callback contract
-/// function on sender with the result. Will only succeed if sender is a smart
-/// contract.
+#[allow(dead_code)]
+type ContractBalanceOfQueryParams = BalanceOfQueryParams<ContractTokenId>;
```
abizjak, line 397: Same as above.

examples/cis1-nft/src/lib.rs
```diff
@@ -381,26 +393,21 @@ fn contract_update_operator<A: HasActions>(
...
+
+/// Get the balance of given token IDs and addresses, and it takes a contract
+/// address plus contract function to invoke with the result.
```
abizjak, line 400: This sentence would read better if it was split into two after `addresses`.

examples/cis1-nft/src/lib.rs
```diff
@@ -412,13 +419,55 @@ fn contract_balance_of<A: HasActions>(
...
-        params.callback.as_ref(),
+        &params.result_contract,
+        params.result_function.as_ref(),
         Amount::zero(),
         &BalanceOfQueryResponse::from(response),
     ))
 }
 
+#[allow(dead_code)]
+type ContractTokenMetadataQueryParams = TokenMetadataQueryParams<ContractTokenId>;
```
abizjak, line 430: Same as above.

examples/cis1-nft/src/lib.rs
```diff
@@ -412,13 +419,55 @@ fn contract_balance_of<A: HasActions>(
...
+
+/// Get the token metadata URLs and checksums given a list of token IDs and it
+/// takes a contract address plus contract function to invoke with the result.
```
abizjak, line 433: This sentence would also read better if it was split into two.
```
/// Get the token metadata URLs and checksums given a list of token IDs. It
/// takes a contract address plus a contract function to invoke with the result.
```

examples/cis1-wgtu/src/lib.rs
```diff
@@ -463,29 +463,27 @@ fn contract_update_operator<A: HasActions>(
     Ok(A::accept())
 }
 
-/// Get the balance of a given token id for a given address and callback
-/// contract function on sender with the result.
-/// Will only succeed if called a smart contracts.
+#[allow(dead_code)]
+type ContractBalanceOfQueryParams = BalanceOfQueryParams<ContractTokenId>;
```
abizjak, line 467: Same as above. It just looks weird especially if it is undocumented why.

examples/cis1-wgtu/src/lib.rs
```diff
@@ -499,7 +497,61 @@ fn contract_balance_of<A: HasActions>(
...
+    Ok(send(
+        &result_contract,
+        result_hook.as_ref(),
+        Amount::zero(),
+        &BalanceOfQueryResponse::from(response),
+    ))
+}
+
+#[allow(dead_code)]
+type ContractTokenMetadataQueryParams = TokenMetadataQueryParams<ContractTokenId>;
```
abizjak, line 509: Same as above.

examples/cis1-wgtu/src/lib.rs
```diff
@@ -499,7 +497,61 @@ fn contract_balance_of<A: HasActions>(
...
+
+/// Get the token metadata URLs and checksums given a list of token IDs and it
+/// takes a contract address plus contract function to invoke with the result.
```
abizjak, line 512: Again, this sentence could be improved by splitting.

examples/cis1-wgtu/src/lib.rs
```diff
@@ -499,7 +497,61 @@ fn contract_balance_of<A: HasActions>(
...
+    let result_contract: ContractAddress = cursor.get()?;
+    // Parse the contract function name to call with the result.
+    let result_hook: OwnedReceiveName = cursor.get()?;
+    // Parse the number of queries.
+    let queries_length: u8 = cursor.get()?;
+
+    // Build the response.
+    let mut response = Vec::with_capacity(queries_length.into());
+    for _ in 0..queries_length {
+        let token_id: ContractTokenId = cursor.get()?;
```
abizjak, line 538: This technically means that there can be leftover data in the parameter which might or might not be an issue.

You could potentially check at the end of the loop that you have reached the end of the parameter.

limemloh: I have made an issue #64 which will provide the api necessary for this.

abizjak (approved):

PR #66, opened 2021-12-09
Improve `claim_eq` and `claim_ne` macros

Bargsteen:
## Purpose

Improves the `claim_eq` and `claim_ne` macros:
 - The operands are now only evaluated once.
    - This was a especially a problem when the evaluation had a side-effect.
 - Type inference now works as you would expect (see example of a typical problem):
    ```rust
    // Prior to this PR,
    claim_eq!(my_result, Ok(0));
    // Was turned into:
    claim!(my_result == Ok(0), "left and right are not equal\nleft: {:?} right{:?}", my_result, Ok(0));
    // Result type inferred-^                                 Cannot infer the Result type here --^
    ```

## Changes

- Use `match` in the macros such that the operands are only evaluated once. This also fixes the type inference.
   - Heavily inspired by the `assert_eq!` macro implementation ;) 
- Merge the trailing-comma case into the regular case by using `$(,)?`
- Add a newline after the error message.
- Write an actual error message for `claim_ne!`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std/CHANGELOG.md
+++ b/concordium-std/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased changes
 - Update references to token to match token name (CCD).
+- Improve `claim_eq` and `claim_ne` macros such that:
+  - Arguments are only evaluated once.
+  - Type inference works as you would expect.
 
 ## concordium-std 1.0.0 (2021-10-05)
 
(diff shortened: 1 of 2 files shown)


abizjak (dismissed):
Looks good, but can you also add a line to the changelog?

abizjak (approved):

PR #67, opened 2021-12-10
Fix doc comment so that formatting does not break it.

abizjak:
## Purpose

Fix formatting issues introduced by rustfmt and wrapping.

Results in corrupt table https://docs.rs/concordium-std/latest/concordium_std/#signalling-errors


abizjak:
Screenshot from 2021-12-10 13-49-34
New look

Bargsteen (approved):

PR #70, opened 2021-12-23
Run example tests in CI (and fix a test)

Bargsteen:
## Purpose

Run the tests for our example contracts in our CI to ensure that they work.
Fix the WCCD contract such that its tests pass.

## Changes

- Add a new CI job
- Use the correct length for transfers in the WCCD contract

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -274,3 +274,43 @@ jobs:
           command: test
           # only run the library tests, no doc tests
           args: --manifest-path ${{ matrix.lib-crates }} --target=${{ matrix.target }} --features=${{ matrix.features }} --lib
+
+  test-examples:
+    name: Test examples
+    runs-on: ubuntu-latest
+    needs: rustfmt
+    strategy:
+      matrix:
+        target:
+          - x86_64-unknown-linux-gnu
+
+        crates:
+          - examples/two-step-transfer/Cargo.toml
+          - examples/piggy-bank/part1/Cargo.toml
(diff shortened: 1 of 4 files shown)


.github/workflows/linter.yml
```diff
@@ -274,3 +274,46 @@ jobs:
           command: test
           # only run the library tests, no doc tests
           args: --manifest-path ${{ matrix.lib-crates }} --target=${{ matrix.target }} --features=${{ matrix.features }} --lib
+
+  test-examples:
+    name: Test examples
+    runs-on: ubuntu-latest
+    needs: rustfmt
```
abizjak, line 281: I think this should also require all of the clippy and checks.

abizjak (approved):
In principle this is OK, but the issue is, and part of the reason why it was not done yet, that we really should be running tests with cargo-concordium for this, not tests on x86.

But I have no direct objection to these changes, they just don't address the issue fully.

.github/workflows/linter.yml
Bargsteen: I could add *all* the examples to the `clippy-schema` job (which runs on wasm). Would that be sufficient or what do you mean by "checks"?

abizjak: Sorry, I meant that the test job should only run after the clippy job. So to add to the `needs` field.

Bargsteen: Ah, I see. I'll update it.

abizjak (dismissed):

abizjak (approved):

Issue #77, opened 2022-02-12
Make all doctests runnable in concordium-std/std-derive


abizjak:
**Task description**

Currently doctests (tests that are part of doc-comments) either don't compile or don't run. We'd ideally have them compile and run, so that they can be maintained properly.

Basically `cargo test` should work.

abizjak:
This is the case in version 6 of the library. Cargo test passes.

PR #79, opened 2022-02-15
Applies the changes done in host functions API.

rimbi:
## Purpose

Applies the changes in the host functions/chain API.

## Checklist

- [x ] My code follows the style of this project.
- [ x] The code compiles without warnings.
- [ x] I have performed a self-review of the changes.
- [ x] I have documented my code, in particular the intent of the
      hard-to-understand areas.
- [ x] (If necessary) I have updated the CHANGELOG.

## CLA acceptance

- [x ] I accept the above linked CLA.

### DIFF
--- a/concordium-std/src/impls.rs
+++ b/concordium-std/src/impls.rs
@@ -7,7 +7,7 @@ use crate::{
     marker::PhantomData,
     mem, num,
     num::NonZeroU32,
-    prims::*,
+    prims::{self, *},
     rc::Rc,
     traits::*,
     types::*,
@@ -135,6 +135,58 @@ impl StateEntry {
             current_position: 0,
         }
     }
+
+    /// Returns the key length or an error if the iterator is not found.
+    fn get_key_length(&self) -> Result<u32, ()> {
+        let key_len = unsafe { prims::state_iterator_key_size(self.state_entry_id as u64) };
+        match key_len {
+            u32::MAX => Err(()),
+            n => Ok(n),
+        }
+    }
+
+    /// Returns the next iterator key or an error if the iterator is not found.
+    fn load_key(&self, key_len: u32) -> Result<Vec<u8>, ContractStateError> {
+        let mut key = Vec::with_capacity(key_len as usize);
+        let key_ptr = key.as_mut_ptr();
+        let res = unsafe { prims::state_iterator_key_read(self.state_entry_id as u64, key_ptr, key_len, 0) };
+        match res {
+            u32::MAX => Err(ContractStateError::EntryNotFound),
+            _ => Ok(key),
+        }
+    }
+
+    fn state_entry_read(&mut self, buf: &mut [u8]) -> Result<u32, ContractStateError> {
+        let len = buf.len().try_into().map_err(|_| ContractStateError::SizeTooLarge)?;
+        let num_read =
+            unsafe { prims::state_entry_read(self.state_entry_id as u64, buf.as_mut_ptr(), len, self.current_position) };
+        match num_read {
+            u32::MAX => Err(ContractStateError::EntryNotFound),
+            _ => Ok(num_read)
+        }
+    }
+
+    fn state_entry_write(&mut self, buf: &[u8]) -> Result<u32, ContractStateError> {
+        let len = buf.len().try_into().map_err(|_| ContractStateError::SizeTooLarge)?;
+        if self.current_position.checked_add(len).is_none() {
+            return Err(ContractStateError::SizeTooLarge);
+        }
+        let res = unsafe {
+            prims::state_entry_write(self.state_entry_id as u64, buf.as_ptr(), len, self.current_position)
+        };
+        match res {
+            u32::MAX => Err(ContractStateError::EntryNotFound),
+            _ => Ok(res),
+        }
+    }
+
+    fn state_entry_size(&self) -> Result<u32, ContractStateError> {
+        let res = unsafe { prims::state_entry_size(self.state_entry_id as u64) };
+        match res {
+            u32::MAX => Err(ContractStateError::EntryNotFound),
+            _ => Ok(res),
+        }
+    }
 }
 
 impl HasContractStateEntry for StateEntry {
@@ -147,34 +199,30 @@ impl HasContractStateEntry for StateEntry {
     }
 
     #[inline(always)]
-    fn size(&self) -> u32 { unsafe { entry_state_size(self.state_entry_id) } }
-
-    fn truncate(&mut self, new_size: u32) {
-        let cur_size = self.size();
-        if cur_size > new_size {
-            unsafe { resize_entry_state(self.state_entry_id, new_size) };
-        }
-        if new_size < self.current_position {
-            self.current_position = new_size
-        }
+    fn size(&self) -> Result<u32, Self::Error> {
+        self.state_entry_size().map_err(|_| ())
     }
 
-    fn reserve(&mut self, len: u32) -> bool {
-        let cur_size = unsafe { entry_state_size(self.state_entry_id) };
-        if cur_size < len {
-            let res = unsafe { resize_entry_state(self.state_entry_id, len) };
-            res == 1
-        } else {
-            true
-        }
+    fn get_key(&self) -> Result<Vec<u8>, ()> {
+        let key_len = self.get_key_length()?;
+        self.load_key(key_len).map_err(|_| ())
     }
 
-    fn get_key(&self) -> Vec<u8> {
-        let key_len = unsafe { get_entry_key_length(self.state_entry_id) };
-        let mut key = Vec::with_capacity(key_len as usize);
-        let key_ptr = key.as_mut_ptr();
-        unsafe { load_entry_key(self.state_entry_id, key_ptr, key_len, 0) };
-        key
+    fn get_position(&self) -> u32 {
+        self.current_position
+    }
+
+    fn set_position(&mut self, new_size: u32) {
+        self.current_position = new_size
+    }
+
+    fn resize(&mut self, new_size: u32) -> Result<(), Self::Error> {
+        let res = unsafe { prims::state_entry_resize(self.state_entry_id as u64, new_size) };
+        match res {
+            1 => Ok(()),
+            0 => Err(()),
+            _ => Err(()),
+        }
     }
 }
 
@@ -193,7 +241,7 @@ impl Seek for StateEntry {
                 _ => Err(()),
             },
             End(delta) => {
-                let end = self.size();
+                let end = self.size()?;
                 if delta >= 0 {
                     match u32::try_from(delta)
                         .ok()
@@ -239,77 +287,9 @@ impl Seek for StateEntry {
 
 impl Read for StateEntry {
     fn read(&mut self, buf: &mut [u8]) -> ParseResult<usize> {
-        let len: u32 = {
-            match buf.len().try_into() {
-                Ok(v) => v,
-                _ => return Err(ParseError::default()),
-            }
-        };
-        let num_read = unsafe {
-            load_entry_state(self.state_entry_id, buf.as_mut_ptr(), len, self.current_position)
-        };
+        let num_read = self.state_entry_read(buf).map_err(|_| ParseError::default())?;
         self.current_position += num_read;
         Ok(num_read as usize)
-    }
-
-    /// Read a `u32` in little-endian format. This is optimized to not
-    /// initialize a dummy value before calling an external function.
-    fn read_u64(&mut self) -> ParseResult<u64> {
-        let mut bytes: MaybeUninit<[u8; 8]> = MaybeUninit::uninit();
-        let num_read = unsafe {
-            load_entry_state(
-                self.state_entry_id,
-                bytes.as_mut_ptr() as *mut u8,
-                8,
-                self.current_position,
-            )
-        };
-        self.current_position += num_read;
-        if num_read == 8 {
-            unsafe { Ok(u64::from_le_bytes(bytes.assume_init())) }
-        } else {
-            Err(ParseError::default())
-        }
-    }
-
-    /// Read a `u32` in little-endian format. This is optimized to not
-    /// initialize a dummy value before calling an external function.
-    fn read_u32(&mut self) -> ParseResult<u32> {
-        let mut bytes: MaybeUninit<[u8; 4]> = MaybeUninit::uninit();
-        let num_read = unsafe {
-            load_entry_state(
-                self.state_entry_id,
-                bytes.as_mut_ptr() as *mut u8,
-                4,
-                self.current_position,
-            )
-        };
-        self.current_position += num_read;
-        if num_read == 4 {
-            unsafe { Ok(u32::from_le_bytes(bytes.assume_init())) }
-        } else {
-            Err(ParseError::default())
-        }
-    }
-
-    /// Read a `u8` in little-endian format. This is optimized to not
-    /// initialize a dummy value before calling an external function.
-    fn read_u8(&mut self) -> ParseResult<u8> {
-        let mut bytes: MaybeUninit<[u8; 1]> = MaybeUninit::uninit();
-        let num_read = unsafe {
-            load_entry_state(
-                self.state_entry_id,
-                bytes.as_mut_ptr() as *mut u8,
-                1,
-                self.current_position,
-            )
-        };
-        self.current_position += num_read;
-        if num_read == 1 {
-            unsafe { Ok(bytes.assume_init()[0]) }
-        } else {
-            Err(ParseError::default())
-        }
     }
 }
 
@@ -317,18 +297,7 @@ impl Write for StateEntry {
     type Err = ();
 
     fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Err> {
-        let len: u32 = {
-            match buf.len().try_into() {
-                Ok(v) => v,
-                _ => return Err(()),
-            }
-        };
-        if self.current_position.checked_add(len).is_none() {
-            return Err(());
-        }
-        let num_bytes = unsafe {
-            write_entry_state(self.state_entry_id, buf.as_ptr(), len, self.current_position)
-        };
+        let num_bytes = self.state_entry_write(buf).map_err(|_| ())?;
         self.current_position += num_bytes; // safe because of check above that len + pos is small enough
         Ok(num_bytes as usize)
     }
@@ -518,7 +487,49 @@ const NEXT_COLLECTION_PREFIX_KEY: u64 = 0;
 const GENERIC_MAP_PREFIX: u64 = 1;
 const INITIAL_NEXT_COLLECTION_PREFIX: u64 = 2;
 
-impl HasContractStateLL for ContractStateLL {
+impl HasFunctionsAPIWrapper for ContractStateLL {}
+
+/// Provides higher level wrapper functions for the implementors around the unsafe host functions.
+trait HasFunctionsAPIWrapper {
+
+    #[inline(always)]
+    fn create_entry(&mut self, key: &[u8]) -> Result<StateEntry, ContractStateError> {
+        let key_start = key.as_ptr();
+        let key_len = key.len() as u32; // Wasm usize == 32bit.
+        let entry_id = unsafe { prims::state_create_entry(key_start, key_len) };
+        match entry_id {
+            u64::MAX => Err(ContractStateError::SubtreeLocked),
+            _ => {
+                let entry_id = entry_id.try_into().unwrap_abort();
+                Ok(StateEntry::open(entry_id))
+            }
+        }
+    }
+
+    #[inline(always)]
+    fn get_parameter_size(&self, i: u32) -> Result<u32, ContractStateError> {
+        let size = unsafe { prims::get_parameter_size(i) };
+        if size == -1 {
+            return Err(ContractStateError::ParameterNotFound);
+        }
+        Ok(size as u32)
+    }
+
+    #[inline(always)]
+    fn get_parameter_section(&self, i: u32, buf: &mut [u8], offset: u32) -> Result<u32, ContractStateError> {
+        let len = buf.len().try_into().map_err(|_| ContractStateError::SizeTooLarge)?;
+        let res = unsafe {
+            prims::get_parameter_section(i, buf.as_mut_ptr(), len, offset)
+        };
+        if res == -1 {
+            return Err(ContractStateError::ParameterNotFound);
+        }
+        Ok(res as u32)
+    }
+}
+
+impl HasContractStateLL for ContractStateLL
+{
     type ContractStateData = ();
     type EntryType = StateEntry;
     type IterType = ContractStateIter;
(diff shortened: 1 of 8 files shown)


concordium-std/src/impls.rs
```diff
@@ -575,61 +576,76 @@ fn prepend_generic_map_key<K: Serial>(key: K) -> Vec<u8> {
...
-        let prefix_start = prefix.as_ptr();
-        let prefix_len = prefix.len() as u32; // Wasm usize == 32bit.
-        let iterator_id = unsafe { iterator(prefix_start, prefix_len) };
-        ContractStateIter {
-            iterator_id,
+    fn iterator(
+        &self,
+        prefix: &[u8],
+    ) -> Result<<T as ContractStateApi>::IterType, ContractStateError> {
+        let locked = self.locked_subtrees.borrow().keys().any(|p| {
```
Bargsteen, line 633: The contract should not be checking whether a subtree is locked on its own. 
It should just call `iterator()` on the host and parse the response.

Bargsteen: But the test trie  _should_ perform these kinds of checks so that the semantics are identical to those of the host.

concordium-std/src/impls.rs
```diff
@@ -697,21 +716,18 @@ where
     }
 
     pub fn is_empty(&self) -> bool {
-        if let Some(_) = self.state_ll.borrow().iterator(&self.prefix).next() {
-            false
-        } else {
-            true
-        }
+        let iterator = self.state_ll.borrow().iterator(&self.prefix);
+        iterator.is_err() || iterator.unwrap().next().is_none()
```
Bargsteen, line 720: `iterator(..)` will return an error if you have more than `N` iterators for the same key. 
So if you get `N` iterators for `self.prefix` and then call `is_empty()`, it will return `true` because you will get an error (despite the trie not being empty).

concordium-std/src/test_infrastructure/trie.rs
```diff
@@ -102,15 +74,179 @@ impl StateTrie {
     }
 }
 
-#[derive(Debug)]
+impl ContractStateApi for StateTrie {
```
Bargsteen, line 77: I don't think this makes sense to implement for StateTrie, as it cannot provide a complete host API.

concordium-std/src/impls.rs
```diff
@@ -1606,3 +1623,200 @@ where
         source.read_u64().and_then(|set_prefix| Ok(StateSet::open(Rc::clone(state), set_prefix)))
     }
 }
+
+impl ContractStateApi for () {
```
Bargsteen, line 1627: Given my comment on the `impl ContractStateApi for StateTrie`, I am not sure this makes so much sense to have.
The idea of wrapping the unsafe calls is fine, but if we do that, we should rewrite the other implementations (those not related to state) to also use this API. For example `slot_time()` in `ChainMetaExtern`. But changing those calls should not be part of the new state PR. (It's large enough already ;) )

concordium-std/src/errors.rs
```diff
@@ -0,0 +1,9 @@
+//! This internal module provides the error types used in the crate.
+
+#[derive(Debug)]
+pub enum ContractStateError {
```
Bargsteen, line 4: We already have a number of error types in `types.rs`, so I think this should live there as well 😊 
If you think that the errors should live in a separate error module, then we can move all of them in a separate PR at some point.

Bargsteen (changes requested):
Good work in general. I like that you added tests for the iterators 👍 

There are a few issues, the primary one being that `ContractStateLL` should not keep track of the locked subtrees. Instead, it should rely on the host doing this job.

I'm likely to blame for this. I should have made that clear when we talked about the task. Luckily, you should already have most of the parts needed. They should just be changed so the subtree-lock checks are only performed on the test trie. 😊

concordium-std/src/errors.rs
```diff
@@ -0,0 +1,9 @@
...
+    /// The subtree is locked.
+    SubtreeLocked,
+    /// The iterator does not exist.
```
abizjak, line 7: The name of the variant does not match the comment above it.

concordium-std/src/errors.rs
abizjak: I agree with Kasper. Do not partially move errors here. This belongs in types.rs.

concordium-std/src/impls.rs
```diff
@@ -145,10 +146,10 @@ impl HasContractStateEntry for StateEntry {
     }
 
     fn get_key(&self) -> Vec<u8> {
```
abizjak, line 148: I don't think this method should return a `Vec<u8>`. The principle we have until now is that it is in principle possible to go without an allocator. So our API should as much as practical not rely on an allocator being present.

Instead the buffer should be provided by the caller.

concordium-std/src/impls.rs
```diff
@@ -575,61 +576,76 @@ fn prepend_generic_map_key<K: Serial>(key: K) -> Vec<u8> {
...
-    type ContractStateData = ();
-    type EntryType = StateEntry;
-    type IterType = ContractStateIter;
+impl<T> ContractStateLL<T>
+where
+    T: ContractStateApi,
+{
+    pub fn new(state_api: T) -> Self {
+        Self {
+            locked_subtrees: Default::default(),
```
abizjak, line 585: I don't think the std should keep track of locked trees. THe host already does that.

concordium-std/src/traits.rs
```diff
@@ -321,6 +340,163 @@ pub trait DeserialCtx: Sized {
...
 where
-    S: HasContractStateLL, {
+    S: HasContractStateLL,
+{
     fn deserial_state_ctx<R: Read>(state: &Rc<RefCell<S>>, source: &mut R) -> ParseResult<Self>;
 }
+
+/// A trait for the contract state / host functions API.
+/// It provides abstractions to access the state of the contract, and to invoke host functions.
+pub trait ContractStateApi: Debug {
```
abizjak, line 350: What is the point of this trait? It is just adding a trait that names the functions that are imported. What problem does this solve?

concordium-std/src/types.rs
```diff
@@ -368,7 +375,7 @@ macro_rules! claim_ne {
 ///
 /// # Example
 /// Defining a custom error type
-/// ```rust
+/// ```ignore
```
abizjak, line 378: This should generally not be done. It makes docs less nice to look at (e.g., worse syntax highlighting). 

THe correct way to fix this is to make it compile, but that should not be part of this PR.

Please revert this.

examples/new-state/src/lib.rs
```diff
@@ -167,7 +167,7 @@ mod tests {
     fn test_init() {
         let ctx = InitContextTest::empty();
 
-        let state_ll = Rc::new(RefCell::new(ContractStateLLTest::new()));
+        let state_ll = Rc::new(RefCell::new(ContractStateLLTest::default()));
```
abizjak, line 170: I strongly prefer being explicit instead of implicit, and default is implicit. Is there are particular reason for this change?

examples/new-state/src/lib.rs
```diff
@@ -215,7 +215,7 @@ mod tests {
             Some(Ok(expected_token_count))
         );
 
-        let mut a_set_iter = state_reloaded.another_struct.a_set.iter();
+        let mut a_set_iter = state_reloaded.another_struct.a_set.iter().unwrap();
```
abizjak, line 218: Use `expect` with a meaningful explanation so that if it fails a useful message is displayed.

concordium-std/src/traits.rs
abizjak: Since we did not need it until now, I do not think it is needed now for contract state changes either.

concordium-std/src/traits.rs
```diff
@@ -133,33 +135,50 @@ where
     fn get_key(&self) -> Vec<u8>;
 }
 
-pub trait HasContractStateLL {
-    type ContractStateData;
-    type EntryType: HasContractStateEntry;
-    type IterType: Iterator<Item = Self::EntryType>;
+pub trait HasContractStateLL: Debug {
```
abizjak, line 138: Why the debug supertrait? 

Code size is an issue, so we should not require things in smart contracts that are not crucial.

concordium-std/src/traits.rs
```diff
@@ -2,12 +2,14 @@
 //! This allows setting-up mock objects for testing individual
 //! contract invocations.
 
-use std::{cell::RefCell, rc::Rc};
+use std::{cell::RefCell, fmt::Debug, rc::Rc};
 
 #[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
```
abizjak, line 9: No spaces between imports please. And let `fmt` sort them.

concordium-std/src/traits.rs
```diff
@@ -2,12 +2,14 @@
 //! This allows setting-up mock objects for testing individual
 //! contract invocations.
 
-use std::{cell::RefCell, rc::Rc};
+use std::{cell::RefCell, fmt::Debug, rc::Rc};
```
abizjak, line 6: And here.

concordium-std/src/test_infrastructure/trie.rs
```diff
@@ -254,7 +387,7 @@ impl Node {
 mod tests {
     use concordium_contracts_common::{to_bytes, Deserial, Write};
```
abizjak, line 389: No spaces between imports please. And let `fmt` sort them out.

concordium-std/src/test_infrastructure/trie.rs
abizjak: I agree. `ContractStateApi` trait should be removed.

concordium-std/src/test_infrastructure.rs
```diff
@@ -971,7 +934,8 @@ mod test {
             false
         );
 
-        let mut iter = state.get::<_, StateSet<u8, _>>(my_set_key).unwrap().unwrap().iter();
+        let mut iter =
+            state.get::<_, StateSet<u8, _>>(my_set_key).unwrap().unwrap().iter().unwrap();
```
abizjak, line 938: Please use `expect` and explain why unwrapping is intended. Otherwise the error message will be hard to understand.

concordium-std/src/test_infrastructure.rs
```diff
@@ -884,7 +847,7 @@ mod test {
     fn high_level_insert_get() {
         let expected_value: u64 = 123123123;
         let mut state: ContractStateHL<ContractStateLLTest> =
-            HasContractStateHL::open(Rc::new(RefCell::new(ContractStateLLTest::new())));
+            HasContractStateHL::open(Rc::new(RefCell::new(ContractStateLLTest::default())));
```
abizjak, line 850: I prefer being explicit, so I don't see a need to use default instead of new.

concordium-std/src/test_infrastructure.rs
```diff
@@ -772,6 +725,16 @@ impl Seek for StateEntryTest {
     }
 }
 
+mod trie;
```
abizjak, line 728: I think this is an unusal way to organize modules and is not the way we use generally.

When using nested submodules they are either in the same file, or we use a `mod.rs` file that names the root.

concordium-std/src/test_infrastructure.rs
```diff
@@ -8,7 +8,7 @@
 //!
 //! # Example
 //!
-//! ```rust
+//! ```ignore
```
abizjak, line 11: Please do not ignore. This will make `cargo doc` worse to read.

The correct way to fix this is to make it compile, but there is a separate issue for that #77 and should be done in a separate PR.

So please revert all of these.

concordium-std/src/impls.rs
abizjak: I don't think the idea makes sense. I do not see what value this abstraction would provide. Why is it valuable to provide a function with the same name that just wraps the unsafe? 

The functions are unsafe for a reason, and rely on correct usage, which is not ensured by this implementation. So these functions are in fact very unsafe and can lead to arbitrary memory corruption. For example `get_receive_sender` must have a buffer that is of correct size, which is not ensured here.

abizjak (changes requested):

concordium-std/src/impls.rs
```diff
@@ -638,21 +691,18 @@ where
...
-        }
+        let iterator = self.state_ll.borrow().iterator(&self.prefix);
+        iterator.is_err() || iterator.unwrap().next().is_none()
     }
 
-    pub fn iter(&self) -> StateMapIter<K, V, S> {
-        let state_iter = self.state_ll.borrow().iterator(&self.prefix);
-        StateMapIter {
+    pub fn iter(&self) -> Result<StateMapIter<K, V, S>, ContractStateError> {
+        let state_iter = self.state_ll.borrow().iterator(&self.prefix)?;
```
Bargsteen, line 703: `StateMapIter` should keep a reference to the `StateMap` so that this method can't be called twice without dropping the `StateMapIter`. When that is implemented, we can unwrap this call to `iterator`.

concordium-std/src/impls.rs
```diff
@@ -680,19 +727,23 @@ where
     fn next(&mut self) -> Option<Self::Item> {
         self.state_iter.next().and_then(|mut entry| {
```
Bargsteen, line 728: This is where we should get the key from the `state_iter` (mentioned previously).

I am not sure what the agreement is regarding `state_iterator_key_read`.
From `prims.rs` it seems like it should be called _before_ `next()` to get the correct key.
But if that is the case, then it `state_iterator_key_read` should also be able to return an error, if there is no next. (Or we should peek first). What do you think @abizjak ?

concordium-std/src/impls.rs
```diff
@@ -911,29 +963,26 @@ where
...
     fn read(&mut self, buf: &mut [u8]) -> ParseResult<usize> {
-        let len: u32 = {
-            match buf.len().try_into() {
-                Ok(v) => v,
-                _ => return Err(ParseError::default()),
-            }
-        };
-        let num_read =
-            unsafe { get_parameter_section(0, buf.as_mut_ptr(), len, self.current_position) };
+        let num_read = self.get_parameter_section(0, buf, self.current_position).unwrap_abort();
```
Bargsteen, line 985: Please don't make unrelated changes such as this. It makes it harder and more time-consuming to review. Aim for minimal changes that _only_ relate to the specific PR 😊

concordium-std/src/impls.rs
```diff
@@ -942,36 +991,30 @@ impl Read for ExternParameter {
...
-        let num_read = unsafe {
-            get_parameter_section(self.i.into(), buf.as_mut_ptr(), len, self.current_position)
-        };
-        if num_read >= 0 {
-            self.current_position += num_read as u32;
-            Ok(num_read as usize)
-        } else {
-            Err(ParseError::default())
-        }
+        let num_read = self.get_parameter_section(self.i.into(), buf, self.current_position).map_err(|_| ParseError::default())?;
```
Bargsteen, line 1005: Same here

concordium-std/src/impls.rs
```diff
@@ -942,36 +991,30 @@ impl Read for ExternParameter {
 impl HasParameter for ExternParameter {
     #[inline(always)]
     // parameter 0 always exists so this is correct
-    fn size(&self) -> u32 { unsafe { get_parameter_size(0) as u32 } }
+    fn size(&self) -> u32 {
```
Bargsteen, line 994: And here

concordium-std/src/impls.rs
```diff
@@ -942,36 +991,30 @@ impl Read for ExternParameter {
...
 }
 
+impl HasFunctionsAPIWrapper for CallResponse {}
+
 /// CallResponse can only be constured in this crate. As a result whenever it is
 /// constructed it will point to a valid parameter, which means that
 /// `get_parameter_size` will always return a non-negative value.
 impl HasCallResponse for CallResponse {
-    fn size(&self) -> u32 { unsafe { get_parameter_size(self.i.get()) as u32 } }
+    fn size(&self) -> u32 { self.get_parameter_size(self.i.get()).unwrap_abort() }
```
Bargsteen, line 1017: And here

concordium-std/src/impls.rs
```diff
@@ -778,7 +830,7 @@ where
     S: HasContractStateLL,
 {
     fn store(self, prefix: &[u8], state_ll: Rc<RefCell<S>>) {
-        match state_ll.borrow_mut().entry(prefix) {
+        match state_ll.borrow_mut().entry(prefix).unwrap_abort() {
```
Bargsteen, line 833: We need to make sure this won't cause problems.

concordium-std/src/impls.rs
```diff
@@ -762,13 +813,14 @@ where
         key_with_prefix
     }
 
-    pub fn iter(&self) -> StateSetIter<T, S> {
-        let state_iter = self.state_ll.borrow().iterator(&self.prefix);
-        StateSetIter {
+    pub fn iter(&self) -> Result<StateSetIter<T, S>, ContractStateError> {
```
Bargsteen, line 816: Same idea as with the iter for statemap. It should have a reference to the stateset.

concordium-std/src/test_infrastructure/trie.rs
```diff
@@ -100,20 +81,77 @@ impl StateTrie {
...
+    }
+
+    pub fn delete_iterator(&mut self, iterator: Iter) -> bool {
+        if iterator.clone_count() > 2 {
+            println!("clone count = {}", iterator.clone_count());
+            return true;
+        }
+        println!("2");
+        let initial_count = self.iterators.borrow().len();
+        self.iterators.borrow_mut().retain(|_, iter| iter != &iterator);
```
Bargsteen, line 144: An optimization here would be that the iterator keeps track of its own prefix, so we can use `remove(iterator.key)` instead. But the current solution is probably fine for testing.

Bargsteen (changes requested):

concordium-std/src/impls.rs
```diff
@@ -578,12 +633,7 @@ impl Iterator for ContractStateIter {
     type Item = StateEntry;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let res = unsafe { next(self.iterator_id) };
-        if res < 0 {
-            None
-        } else {
-            Some(StateEntry::open(res as u32))
-        }
+        self.state_iterator_next().unwrap_abort()
```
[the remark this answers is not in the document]
abizjak: That depends a bit on what the intent of this `Iterator` implementation is.

One option is to abort, but this seems like the worst option.
Another option is to just return None, terminate the iterator when encountering an error.

The last option is to have an iterator over `Result<StateEntry, _>`. This would then reflect the fact that this is just a convenience wrapper around `state_iterator_next`.

I would imagine this is only used in low-level implementations, so my preference would be option (3), or remove the implementation of the Iterator trait entirely.

concordium-std/src/impls.rs
```diff
@@ -606,17 +656,17 @@ where
 
     pub fn get(&self, key: K) -> Option<V> {
         let k = self.key_with_map_prefix(&key);
-        self.state_ll.borrow().lookup(&k).and_then(|mut entry| {
-            Some(V::deserial_state_ctx(&self.state_ll, &mut entry).expect_report(
+        self.state_ll.borrow().lookup(&k).map(|mut entry| {
+            V::deserial_state_ctx(&self.state_ll, &mut entry).expect_report(
```
abizjak, line 656: `expect_report` **cannot be used in production code**.

It is only available when testing locally and you **will not be able to build a contract that uses it**. So its use must be removed.

concordium-std/src/impls.rs
abizjak: Since `N=2^16` perhaps we should just abort in that case?

abizjak: Looking at the details of the implementation in the host now, there seems to be no disadvantage in using a u32 for this limit, so I think we can raise it to 2^32, which makes using abort here appropriate.

concordium-std/src/impls.rs
abizjak: Yes, you are right. `state_iterator_key_read` should return an Option indicating whether there is a next value or not.
I'll revise the host implementation and specification.

This also makes it cheaper to use to check whether an iterator is empty (i.e., `is_empty` method above for the state map).

concordium-std/src/impls.rs
```diff
@@ -680,19 +727,23 @@ where
...
-            Some((k, v))
+            if let Ok(key) = key {
+                let mut key_cursor = Cursor {
+                    data:   key,
+                    offset: 8, // Items in a map always start with the set prefix which is 8 bytes.
+                };
+                let k = K::deserial(&mut key_cursor).expect_report(
+                    "Deserializing key failed. State has been incorrectly altered using the low-level \
+                     API.",
+                );
```
abizjak, line 734: As above, expect_report is not to be used in production code, only testing.

concordium-std/src/impls.rs
```diff
@@ -680,19 +727,23 @@ where
...
+                let v = V::deserial_state_ctx(&self.state_ll, &mut entry).expect_report(
+                    "Deserializing value failed. State has been incorrectly altered using the \
+                     low-level API.",
+                );
```
abizjak, line 738: And here.

concordium-std/src/impls.rs
abizjak: Yes.

I would just like you don't get bogged down in this part and that we get something working if the references and lifetimes are not the most straightforward concept.

We can do it in stages.

concordium-std/src/impls.rs
abizjak: I completely agree with this and I am strongly against such changes.

There is already so much subtle stuff going on that we do not need more changes.

concordium-std/src/impls.rs
```diff
@@ -942,36 +991,30 @@ impl Read for ExternParameter {
 impl HasParameter for ExternParameter {
     #[inline(always)]
     // parameter 0 always exists so this is correct
-    fn size(&self) -> u32 { unsafe { get_parameter_size(0) as u32 } }
+    fn size(&self) -> u32 {
+        self.get_parameter_size(0).unwrap_abort()
+    }
 }
 
+impl HasFunctionsAPIWrapper for ExternParameter {}
```
abizjak, line 995: Remove this trait. It does not serve a purpose.

concordium-std/src/impls.rs
```diff
@@ -942,36 +991,30 @@ impl Read for ExternParameter {
...
-        } else {
-            Err(ParseError::default())
-        }
+        let num_read = self.get_parameter_section(self.i.into(), buf, self.current_position).map_err(|_| ParseError::default())?;
+        self.current_position += num_read as u32;
+        Ok(num_read as usize)
     }
 }
 
+impl HasFunctionsAPIWrapper for CallResponse {}
```
abizjak, line 1007: I don't get these implementations. `CallResponse` should only have Read + Seek implementation. That is the only thing you can do with it.

concordium-std/src/impls.rs
```diff
@@ -1563,7 +1606,12 @@ impl<A> UnwrapAbort for Option<A> {
     type Unwrap = A;
 
     #[inline(always)]
-    fn unwrap_abort(self) -> Self::Unwrap { self.unwrap_or_else(|| crate::trap()) }
+    fn unwrap_abort(self) -> Self::Unwrap {
+        match self {
+            Some(x) => x,
+            None => crate::trap(),
+        }
+    }
```
abizjak, line 1610: Why was this change made?

Please avoid unrelated changes unless they are bugs. In which case it is still better to document a bug and fix it in a separate PR.

concordium-std/src/test_infrastructure/trie.rs
```diff
@@ -17,6 +17,16 @@ pub struct StateTrie {
     nodes:         Node,
     next_entry_id: Cell<StateEntryId>,
     entry_map:     RefCell<HashMap<StateEntryId, Vec<Index>>>,
+    iterators:     RefCell<BTreeMap<Vec<usize>, Iter>>,
+}
+
+/// Default constructor for a new state trie.
+/// 
+/// Added after clippy warning.
```
abizjak, line 25: That's not a good comment :) It does not provide value to the reader.

Instead document what the default value is, i.e., the empty state trie.

concordium-std/src/test_infrastructure/trie.rs
```diff
@@ -100,20 +81,77 @@ impl StateTrie {
...
+        self.iterators.borrow_mut().insert(index_prefix, iter.clone());
+        Ok(iter)
+    }
+
+    pub fn delete_iterator(&mut self, iterator: Iter) -> bool {
+        if iterator.clone_count() > 2 {
+            println!("clone count = {}", iterator.clone_count());
+            return true;
+        }
+        println!("2");
```
abizjak: I guess debug print? Should probably be removed.

concordium-std/src/test_infrastructure/trie.rs
```diff
@@ -100,20 +81,77 @@ impl StateTrie {
...
+        }
+        let node = self.nodes.lookup_node(&index_prefix).ok_or(ContractStateError::EntryNotFound)?;
+        let iter = Iter::new(self, index_prefix.clone(), node);
+        self.iterators.borrow_mut().insert(index_prefix, iter.clone());
+        Ok(iter)
+    }
+
+    pub fn delete_iterator(&mut self, iterator: Iter) -> bool {
+        if iterator.clone_count() > 2 {
+            println!("clone count = {}", iterator.clone_count());
```
abizjak: I guess debug print? Should probably be removed.

concordium-std/src/types.rs
```diff
@@ -514,3 +514,21 @@ pub(crate) mod sealed {
...
+    /// The entry does not exist.
+    EntryNotFound,
+    /// The parameter does not exist.
+    ParameterNotFound,
+    /// The specified size is too big.
+    SizeTooLarge,
+    /// Overflow error.
+    Overflow,
+    /// The iterator limit for a prefix has been reached.
+    IteratorLimitForPrefixExceeded,
```
abizjak, line 533: I suppose you also want to add `IteratorNotFound` error here.

concordium-std/src/types.rs
```diff
@@ -514,3 +514,21 @@ pub(crate) mod sealed {
...
+    /// The operation is not allowed.
+    NotAllowed,
```
abizjak, line 535: Where is this used? At a quick glance I could not see it.

examples/new-state/src/lib.rs
```diff
@@ -208,7 +208,7 @@ mod tests {
             Some(expected_token_count)
         );
 
-        let mut a_set_iter = state_reloaded.another_struct.a_set.iter();
+        let mut a_set_iter = state_reloaded.another_struct.a_set.iter().expect("Failed to get iterator");
```
abizjak, line 211: In tests it is better to use `expect_report` since that will make `cargo concordium` report the error. This won't since panics are not reported, the program just terminates.

abizjak (changes requested):

Bargsteen:
Merged this into https://github.com/Concordium/concordium-rust-smart-contracts/tree/new-contract-state and resolved most of the comments.
I'll close this now, but keep the remaining comments as an additional todo list ;)

Issue #73, opened 2022-01-09
New contract state: concordium-std support


abizjak:
**Task description**

This is part of the new contract state feature, see also https://github.com/Concordium/concordium-node/issues/237 

Since the state will look very different in V1 contracts we will need to significantly revise the support for using it. The key features that need to be provided by concordium-std are

- efficient mappings from small objects (e.g., account addresses, token IDs) to (potentially) more complex objects
- efficient sets of small objects (e.g., account addresses, token IDs)
- easy to use state with multiple fields, e.g., a Rust struct as the state of a smart contract
- nested maps and sets should be possible and painless

Some of this will require writing new procedural macros to generate code for users as much as possible, based on their annotations.

As usual, this needs to be provided in two variants, one for use on the chain, and one for unit testing. The latter will involve essentially implementing the same kind of backing radix trie implementation as is used by the node, except the implementation does not have to deal with persistence.

Proposed API for the new contract state
```rust
/// Lookup an entry with the given key. The return value is either
/// u64::MAX if the entry at the given key does not exist, or else
/// the first bit of the result is 0, and the remaining bits
/// are an entry identifier that may be used in subsequent calls.
fn state_lookup_entry(key_start: *const u8, key_length: u32) -> u64;

/// Create an empty entry with the given key. The return value is either u64::MAX if
/// creating the entry failed because of an iterator lock on the part of the
/// tree, or else the first bit is 0, and the remaining bits are an entry identifier
/// that maybe used in subsequent calls.
fn state_create_entry(key_start: *const u8, key_length: u32) -> u64;

/// Delete the entry. Returns one of
/// - 0 if the part of the tree this entry was in is locked
/// - 1 if the entry did not exist
/// - 2 if the entry was deleted as a result of this call.
fn state_delete_entry(entry: u64) -> u32;

/// Delete a prefix in the tree, that is, delete all parts of the tree that have
/// the given key as prefix. Returns
/// - 0 if the tree was locked and thus deletion failed.
/// - 1 if the tree **was not locked**, but the key points to an empty part of the tree
/// - 2 if a part of the tree was successfully deleted
fn state_delete_prefix(key_start: *const u8, key_length: u32) -> u32;

/// Construct an iterator over a part of the tree. This **locks the part of the
/// tree that has the given prefix**. Locking means that no deletions or
/// insertions of entries may occur in that subtree.
/// Returns
/// - all 1 bits if too many iterators already exist with this key
[...]

PR #82, opened 2022-02-16
New contract state

Bargsteen:
## Purpose

Addresses #73.

## Changes

Subtasks:
- [x] Provide a low-level abstraction over the new state host functions
- [x] Provide an ergonomic high-level abstraction over the new state host functions
   - [x] An allocator for allocating maps/set/boxes (with unique prefixes)
   - [x] StateMap
     - [x] Consider whether `get` should return `Option<ParseResult<V>>` or `Option<V>` (the latter should be safe if users only use the high-level API). The latter has been implemented.
     - [x] Make the iterator return a &V. 
        - ~Perhaps by using a `StateBox<V>` underneath.~
        - Problem solved with `StateRef`
     - [x] Implement `iter_mut`
   - [x] StateSet
   - [x] StateBox
       - [x] Basic implementation
       - [x] Should be changed, so it is created by the allocator
       - [x] Make it lazy and use caching.
       - [ ] Add helper methods for `LazyStateBox<Box<T>>` to allow for recursive structures, such as linked lists.
       - [x] Figure out how to implement `Deref`.
   - [ ] Deallocation
      - [ ] Find a way to deallocate the existing box when doing: `state.my_box = allocator.new_box(..)` (however, this use case should be rare)
        - A slightly different use-case is now support due to the `Freeable` trait: `let old_box = state.my_box.take(); old_box.free(); state.my_box = Some(allocator.new_box(..);`
      - [x] Make sure deleting entries in maps/sets with indirections (for example `StateMap<StateBox<u8>>`) also deletes the data pointed at. 
         - ~Adding a `delete` method on `Persistable` might be the right way.~
         - [x] Add `Freeable` trait with a `free` methods and implement for state types that should clean up the state they own.
- [x] Update concordium-std-derive
   - [x] Use the new contract state
   - [x] Add a `mutable` attribute to `receive` (and only call `state.store()` after a mutable receive).
   - [x] Derive `DeserialStateCtx` ~(Should we just hide this fact and make `#[derive(Serialize)]` derive `Serial` + `DeserialStateCtx`?)~
- [x] Update test infrastructure
   - [x] Implement a test trie
   - [x] Update the trie to match the semantic changes in the host (addressed in #79)
   - [x] Branch on 4 bits to match https://github.com/Concordium/concordium-wasm-smart-contracts/pull/38/
- [x] Technical improvements
   - [x] Use `()` (or our own zero-sized type) instead of `Rc<RefCell<()>>` for `ConstractStateLL` so the `borrow().clone()` is optimized away.
   - [x] Ensure the correct relationship (including blanket impls) between `Persistable` and `[De]Serial[StateCtx]`
     - ~Perhaps the right solution is to remove `DeserialStateCtx` and replace it with a blanket impl for `Persistable` for everything that is `Serializable` (+ implementations for statebox, set, and map, which also deletes things they point at).~
- [x] Merge in sync-calls changes
- [x] Bugfixes
   - [x] map and set `.clear` do not behave in the expected way. They delete the collection.
- [x] There is an issue with space leaks when updating/replacing inner boxes, maps and sets. The user has to delete it manually if they replace it. This needs to at least be documented.
- [x] There should be `derive(Deletable)` to enable convenient use of user data types.
- [x] Update examples
- [x] Documentation
- [x] Polish

## Checklist

      hard-to-understand areas.


concordium-std/src/impls.rs
```diff
@@ -251,46 +326,610 @@ impl Write for ContractState {
...
+                                                                 // fail.
+    }
+
+    pub fn try_modify<F, E>(&mut self, f: F) -> Result<(), E>
+    where
+        F: FnOnce(&mut V) -> Result<(), E>, {
+        f(&mut self.value)?;
+        // Rewind state entry. Cannot fail.
+        self.state_entry.seek(SeekFrom::Start(0)).unwrap_abort();
+        self.value.serial(&mut self.state_entry).unwrap_abort(); // Writing to state cannot fail.
```
abizjak, line 463: Same here I believe. We need to make sure there is no leftover.

concordium-std/src/impls.rs
```diff
@@ -251,46 +326,610 @@ impl Write for ContractState {
...
+        }
+    }
+
+    /// Try to get the value with the given key.
+    pub fn get(&self, key: &K) -> Option<V> {
+        let k = self.key_with_map_prefix(&key);
+        self.state_ll.borrow().lookup(&k).and_then(|mut entry| {
+            Some(V::deserial_state_ctx(&self.state_ll, &mut entry).expect_report(
+                "Deserial failed. State has been incorrectly altered using the low-level API.",
+            ))
```
abizjak, line 613: expect_report cannot be used in production code.

concordium-std/src/impls.rs
```diff
@@ -251,46 +326,610 @@ impl Write for ContractState {
...
+        let value_bytes = to_bytes(&value);
+        match self.state_ll.borrow_mut().entry(&key_bytes) {
+            EntryRaw::Vacant(vac) => {
+                let _ = vac.insert(&value_bytes);
+                None
+            }
+            EntryRaw::Occupied(mut occ) => {
+                let old_value = V::deserial_state_ctx(&self.state_ll, occ.get_mut()).expect_report(
+                    "Deserial failed. State has been incorrectly altered using the low-level API.",
+                );
```
abizjak, line 630: expect_report cannot be used in production code.

concordium-std/src/impls.rs
```diff
@@ -251,46 +326,610 @@ impl Write for ContractState {
...
+
+    /// Get an entry for the given key.
+    pub fn entry(&mut self, key: K) -> Entry<K, V, S::EntryType> {
+        let key_bytes = self.key_with_map_prefix(&key);
+        match self.state_ll.borrow_mut().entry(&key_bytes) {
+            EntryRaw::Vacant(vac) => Entry::Vacant(VacantEntry::new(key, vac.state_entry)),
+            EntryRaw::Occupied(mut occ) => {
+                let value = V::deserial_state_ctx(&self.state_ll, occ.get_mut()).expect_report(
+                    "Deserial failed. State has been incorrectly altered using the low-level API.",
+                );
```
abizjak, line 645: expect_report cannot be used in production code.

concordium-std/src/traits.rs
```diff
@@ -114,34 +118,78 @@ pub trait HasReceiveContext<Error: Default = ()>: HasCommonData {
...
+
+    /// Get an iterator over a map in the state.
+    fn iterator(&self, prefix: &[u8]) -> Self::IterType;
+}
+
+pub trait Persistable<S>
+where
+    Self: Sized, {
+    // TODO: Should this return something?
+    fn store(self, prefix: &[u8], state_ll: Rc<RefCell<S>>);
```
abizjak, line 182: As we discussed, it would be good to get rid of this `Rc<RefCell<S>>` for the host backed implementation.

concordium-std/src/types.rs
```diff
@@ -1,11 +1,100 @@
-use crate::num::NonZeroU32;
+use crate::{num::NonZeroU32, HasContractStateEntry, HasContractStateLL};
+use std::{cell::RefCell, marker::PhantomData, rc::Rc};
```
abizjak, line 2: This part generally has to work with and without `std`. 

We re-export some definitions already in `lib.rs.` If these are not please add them there. But I hope we can remove the Rc.

concordium-std/src/types.rs
```diff
@@ -1,11 +1,100 @@
...
+
+#[derive(Debug)]
+pub struct StateMap<K, V, S>
+where
+    S: HasContractStateLL, {
+    pub(crate) _marker_key:   PhantomData<K>,
+    pub(crate) _marker_value: PhantomData<V>,
```
abizjak, line 9: I am not sure whether this will be helpful or detrimental, but just adding it here so we can come back to it.
It might be better to have `PhantomData<*const K>` and the same for `V` instead.

See https://doc.rust-lang.org/std/marker/struct.PhantomData.html#ownership-and-the-drop-check

concordium-std/src/types.rs
```diff
@@ -1,11 +1,100 @@
...
+
+#[derive(Debug)]
+pub struct StateBox<T, S: std::fmt::Debug> {
+    pub(crate) prefix:   StateItemPrefix,
+    pub(crate) state_ll: Rc<RefCell<S>>,
+    pub(crate) _marker:  PhantomData<T>,
+}
+
+#[derive(Debug, Default)]
+pub struct ContractStateLL;
```
abizjak, line 46: This should probably be `doc(hidden)` at the end of the day.

concordium-std/src/types.rs
```diff
@@ -1,11 +1,100 @@
...
 
+pub struct VacantEntryRaw<StateEntryType>
+where
+    StateEntryType: HasContractStateEntry, {
+    pub(crate) state_entry: StateEntryType,
+}
+
+pub struct OccupiedEntryRaw<StateEntryType>
+where
+    StateEntryType: HasContractStateEntry, {
```
abizjak, line 70: Do all of these constraints on the structs provide value?

In my experience they were usually not very useful and tended to clutter things, forcing the constraint even for methods that did not require them.

abizjak (commented):
I have some not-so-important comments now.

I am still looking at the design of the statebox but while it bothers me a bit, I need some more time to think about what exactly I would change. I hope to look at it tonight.

abizjak:
@Bargsteen Regarding the statebox.

I was considering whether something like the following would be a better approach
```rust
pub struct StateBox<T, S: std::fmt::Debug> {
    pub(crate) prefix:   StateItemPrefix,
    pub(crate) state_ll: Rc<RefCell<S>>,
    pub(crate) value: T,
}
```
with the idea being that 
- when serializing we only serialize the prefix
- when deserializing we first load the prefix, and then load the value itself at the prefix. We could do tradeoffs here, doing this part lazily if `value: Box<RefCell<Option<T>>>`, for example, or keeping track of whether it was modified, but this is an optimization.
- the rationale is that it seems that it'd be unusual that you get the box itself, but do not want the value, so just loading the value immediately seems desirable.

The downside for this would be with recursive structures, where `T = ...StateBox<T>` (e.g., a linked list) would mean that loading the head would load the entire list. This however could also be addressed in two ways.
- Introduce another "LazyBox" type if you really want this lazy behaviour. The difference between Box and LazyBox would be that `LazyBox` has a method `.get` that takes a mutable reference to self so that it can modify itself, i.e., `value: Option<T>` for `LazyBox`.
- Or make `value: Box<RefCell<Option<T>>>` and quietly modify it during `.get`.

It would seem to me that option 1 would be reasonable if users mostly use the box to make things "persistable" and not for building recursive structures.

abizjak:
Another thing I am not quite sure of but I have some proposals which seem to simplify things.

Currently there is this generic implementation of Persistable
```rust
impl<T, S> Persistable<S> for T
where
    T: Serial + DeserialStateCtx<S>,
    S: HasContractStateLL,
{
    fn store(self, prefix: &[u8], state_ll: Rc<RefCell<S>>) {
        match state_ll.borrow_mut().entry(prefix) {
            EntryRaw::Vacant(vac) => {
                let _ = vac.insert(&to_bytes(&self));
            }
            EntryRaw::Occupied(mut occ) => occ.insert(&to_bytes(&self)),
        }
    }

    fn load(prefix: &[u8], state_ll: Rc<RefCell<S>>) -> ParseResult<Self> {
        match state_ll.borrow().lookup(prefix) {
            Some(mut entry) => Self::deserial_state_ctx(&state_ll, &mut entry),
            None => Err(ParseError::default()),
        }
    }
}
```

Would it not make more sense to have specific implementations for
```
StateMap
StateSet
StateBox
```

Then you'd have implicit StateBox at the top-level of any state. That would mean changing 
```rust
pub struct ExternHost<State> {
    pub state:     State,
    pub allocator: Allocator<ContractStateLL>,
}
```
to 
```rust
pub struct ExternHost<State> {
    pub state:     StateBox<State>,
    pub allocator: Allocator<ContractStateLL>,
}
```

This would mean the user could have any struct as their top-level state, without worrying about persistable. The state just needs to be serializable. It seems to me that then the design tradeoffs would be when considering performance. For example having

```rust
struct State {
   field1: Vec<u64>,
   field2: u64
}
```

```rust
struct State {
   field1: StateBox<Vec<u64>>,
   field2: u64
}
```
would be about whether `field2` is ever updated without updating `field1`. If so, using the StateBox makes sense and adds value, otherwise it does not.

concordium-std/src/types.rs
Bargsteen: I think they are left there by accident. I'll remove them.

Bargsteen:
> @Bargsteen Regarding the statebox.
> 
> I was considering whether something like the following would be a better approach
> 
> ```rust
> pub struct StateBox<T, S: std::fmt::Debug> {
>     pub(crate) prefix:   StateItemPrefix,
>     pub(crate) state_ll: Rc<RefCell<S>>,
>     pub(crate) value: T,
> }
> ```
> 
> with the idea being that
> 
> * when serializing we only serialize the prefix
> * when deserializing we first load the prefix, and then load the value itself at the prefix. We could do tradeoffs here, doing this part lazily if `value: Box<RefCell<Option<T>>>`, for example, or keeping track of whether it was modified, but this is an optimization.
> * the rationale is that it seems that it'd be unusual that you get the box itself, but do not want the value, so just loading the value immediately seems desirable.
> 
> The downside for this would be with recursive structures, where `T = ...StateBox<T>` (e.g., a linked list) would mean that loading the head would load the entire list. This however could also be addressed in two ways.
> 
> * Introduce another "LazyBox" type if you really want this lazy behaviour. The difference between Box and LazyBox would be that `LazyBox` has a method `.get` that takes a mutable reference to self so that it can modify itself, i.e., `value: Option<T>` for `LazyBox`.
> * Or make `value: Box<RefCell<Option<T>>>` and quietly modify it during `.get`.
> 
> It would seem to me that option 1 would be reasonable if users mostly use the box to make things "persistable" and not for building recursive structures.
> 
> What do you think?

Well, my initial design was pretty much the `LazyStateBox`. I changed it to this new design in https://github.com/Concordium/concordium-rust-smart-contracts/pull/82/commits/f64c468d101ff4e301c7113d61fdc176c8735f20 .
I think we should have both a lazy and strict version because it's not possible to construct recursive structures with the strict (which holds the actual `T`) version due problems with `Size`.

Bargsteen:
> Another thing I am not quite sure of but I have some proposals which seem to simplify things.
> 
> Currently there is this generic implementation of Persistable
> 
> ```rust
> impl<T, S> Persistable<S> for T
> where
>     T: Serial + DeserialStateCtx<S>,
>     S: HasContractStateLL,
> {
>     fn store(self, prefix: &[u8], state_ll: Rc<RefCell<S>>) {
>         match state_ll.borrow_mut().entry(prefix) {
>             EntryRaw::Vacant(vac) => {
>                 let _ = vac.insert(&to_bytes(&self));
>             }
>             EntryRaw::Occupied(mut occ) => occ.insert(&to_bytes(&self)),
>         }
>     }
> 
>     fn load(prefix: &[u8], state_ll: Rc<RefCell<S>>) -> ParseResult<Self> {
>         match state_ll.borrow().lookup(prefix) {
>             Some(mut entry) => Self::deserial_state_ctx(&state_ll, &mut entry),
>             None => Err(ParseError::default()),
>         }
>     }
> }
> ```
> 
> Would it not make more sense to have specific implementations for
> 
> ```
> StateMap
> StateSet
> StateBox
> ```
> 
> Then you'd have implicit StateBox at the top-level of any state. That would mean changing
> 
> ```rust
> pub struct ExternHost<State> {
>     pub state:     State,
>     pub allocator: Allocator<ContractStateLL>,
> }
> ```
> 
> to
> 
> ```rust
> pub struct ExternHost<State> {
>     pub state:     StateBox<State>,
>     pub allocator: Allocator<ContractStateLL>,
> }
> ```
> 
> This would mean the user could have any struct as their top-level state, without worrying about persistable. The state just needs to be serializable. It seems to me that then the design tradeoffs would be when considering performance. For example having
> 
> ```rust
> struct State {
>    field1: Vec<u64>,
>    field2: u64
> }
> ```
> 
> ```rust
> struct State {
>    field1: StateBox<Vec<u64>>,
>    field2: u64
> }
> ```
> 
> would be about whether `field2` is ever updated without updating `field1`. If so, using the StateBox makes sense and adds value, otherwise it does not.

Yes, removing the blanket impl was also my plan. I think that would make sense.

Wrapping the state in a statebox is one option. I also looked into restricting the state type to be of `Persistable` (once the blanket impl is removed). This would allow for a state that is simply a StateMap or StateSet.

One problem with both approaches remains, however. Namely replacing the whole state:
```
host.state() = allocator.new_[box, set, map](..)
```

Regarding the structs, I think your suggestion makes a lot of sense. I recently had the idea that we would allow users to derive `Persistable` for structs if they wanted to be able to update fields individually. If they wouldn't need that, they could simply use a `Serializable` struct wrapped in a box. But your suggestion is simpler to understand and implement.

concordium-std/src/impls.rs
```diff
@@ -516,10 +516,17 @@ where
     pub fn into_key(self) -> K { self.key }
 
     /// Sets the value of the entry with the `VacantEntry`'s key.
-    pub fn insert(mut self, value: V) {
+    pub fn insert(mut self, value: V) -> OccupiedEntry<'a, K, V, StateEntryType> {
         // Writing to state cannot fail.
+        // TODO: This is subtle. Should we normalize to 0 first, and reset the cursor?
```
Bargsteen, line 521: I assume this means whether we should overwrite any existing values with 0s.
I don't think that is necessary, because the only way to get a `VacantEntry` is through `StateMap::entry`, which calls `prims::create_state_entry`, which in turn ensures that we are looking at a blob of bytes.

concordium-std/src/impls.rs
```diff
@@ -970,6 +982,7 @@ where
             offset: 8, // Items in a map always start with the set prefix which is 8 bytes.
         };
         // Unwrapping is safe when only using the high-level API.
+        // TODO: This is inefficient. There's no need to allocate the entry.
```
Bargsteen, line 985: I am not sure I understand this TODO.

Bargsteen: Do you want to make StateRef load the values lazily?

concordium-std/src/impls.rs
```diff
@@ -596,19 +603,21 @@ where
     StateEntryType: HasStateEntry,
 {
     /// Ensures a value is in the entry by inserting the default if empty.
-    pub fn or_insert(self, default: V) {
-        if let Entry::Vacant(vac) = self {
-            vac.insert(default);
+    pub fn or_insert(self, default: V) -> OccupiedEntry<'a, K, V, StateEntryType> {
```
Bargsteen, line 606: This is a nice addition :+1:

concordium-std/src/types.rs
```diff
@@ -142,8 +143,11 @@ impl<'a, V> crate::ops::Deref for StateRef<'a, V> {
 }
 
 #[derive(Debug)]
-/// The [`StateRefMut<_, V, _>`] behaves like the type [`StateBox<V, _>`] in the
-/// way that it can be accessed.
+/// The [`StateRefMut<_, V, _>`] behaves like the type [`StateBox<V, _>`],
+/// but its lifetime is tied to the lifetime of some parent object. In that
+/// sense it is more like `&'mut V` except it does not implement
```
Bargsteen, line 148: Should be `&mut V`

examples/recorder/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+    use super::*;
+    use test_infrastructure::*;
+
+    #[concordium_test]
+    fn test_init() {
+        let ctx = TestInitContext::empty();
+
+        let mut state_builder = TestStateBuilder::new();
+
+        let _ = init(&ctx, &mut state_builder).expect("Init failed");
```
Bargsteen, line 97: `expect_report`

examples/auction/src/lib.rs
```diff
@@ -95,19 +83,29 @@ enum FinalizeError {
...
+fn auction_init<S: HasStateApi>(
+    ctx: &impl HasInitContext,
+    state_builder: &mut StateBuilder<S>,
+) -> InitResult<State<S>> {
     let parameter: InitParameter = ctx.parameter_cursor().get()?;
-    Ok(((), fresh_state(parameter.item, parameter.expiry)))
+    let state = State {
+        auction_state: AuctionState::NotSoldYet,
+        highest_bid:   Amount::zero(),
+        item:          parameter.item,
```
Bargsteen, line 94: I think we should make the `item` a `String` now that we're at it.

abizjak: Changed.

concordium-std/src/impls.rs
```diff
@@ -251,46 +391,1124 @@ impl Write for ContractState {
...
+        }
+    }
+}
+
+impl<'a, V, S> StateRefMut<'a, V, S>
+where
+    V: Serial + DeserialWithState<S>,
+    S: HasStateApi,
+{
+    pub fn get(&self) -> &V {
```
Bargsteen, line 1139: Docs.

concordium-std/src/impls.rs
```diff
@@ -251,46 +391,1124 @@ impl Write for ContractState {
...
+            v
+        } else {
+            let entry = unsafe { &mut *self.entry.get() };
+            entry.move_to_start();
+            let value = V::deserial_with_state(&self.state_api, entry).unwrap_abort();
+            lv.insert(value)
+        }
+    }
+
+    pub fn get_mut(&mut self) -> &mut V {
```
Bargsteen, line 1151: Docs.

concordium-std/src/types.rs
```diff
@@ -1,28 +1,561 @@
...
-/// A type representing the constract state bytes.
-#[derive(Default)]
-pub struct ContractState {
+#[derive(Debug)]
+/// A high-level map based on the low-level key-value store, which is the
+/// interface provided by the chain.
+///
+/// In most situations, this collection should be preferred over
+/// [`BTreeMap`][btm] and [`HashMap`][hm] since it will be more efficient to
+/// lookup and update since costs to lookup and update will be grow very slowly
```
Bargsteen, line 9:
```suggestion
/// lookup and update since costs to lookup and update will grow very slowly
```

concordium-std/src/types.rs
```diff
@@ -1,28 +1,561 @@
...
+    pub(crate) _lifetime_marker: PhantomData<&'a mut (K, V)>,
+}
+
+#[derive(Debug)]
+/// A high-level set of _flat_ values based on the low-level key-value store,
+/// which is the interface provided by the chain.
+///
+/// In most situations, this collection should be preferred over
+/// [`BTreeSet`][bts] and [`HashSet`][hs] since it will be more efficient to
+/// lookup and update since costs to lookup and update will be grow very slowly
```
Bargsteen, line 189:
```suggestion
/// lookup and update since costs to lookup and update will grow very slowly
```

concordium-std/src/impls.rs
```diff
@@ -251,46 +391,1124 @@ impl Write for ContractState {
...
+        Ok(entry)
+    }
+
+    /// Sets the value of the entry with the `VacantEntryRaw`’s key.
+    /// This differs from
+    /// [`insert`](Self::insert) in that it automatically serializes
+    /// the provided value. [`insert`](Self::insert) should be preferred
+    /// for values that can be directly converted to byte arrays, e.g., any
+    /// value that implements [`AsRef<[u8]>](AsRef).
+    pub fn insert_serial<V: Serial>(
```
Bargsteen, line 433: We use `x` for the auto-serializing version and `x_raw` for the byte version elsewhere. Perhaps we should rename this to match?

concordium-std/src/types.rs
```diff
@@ -1,28 +1,561 @@
...
+/// ### Low-level state `S`
+///
+/// The type parameter `S` is extra compared to usual Rust collections. As
+/// mentioned above it specifies the [low-level state
+/// implementation](crate::HasStateApi). This library provides two such
+/// implementations. The "external" one, which is the implementation supported
+/// by external host functions provided by the chain, and a
+/// [test](crate::test_infrastructure::TestStateApi) one. The latter one is
+/// useful for testing since it provides an implementation that is easier to
+/// construct, executed, and inspect during unit testing.
```
Bargsteen, line 65:
```suggestion
/// construct, execute, and inspect during unit testing.
```

concordium-std/src/impls.rs
abizjak: Yeah, makes sense.

concordium-std/src/impls.rs
```diff
@@ -1037,3 +2479,134 @@ impl DeserialCtx for String {
...
+    fn delete(mut self) {
+        // Delete all values pointed at by the statemap. This is necessary if `V` is a
+        // StateBox/StateMap.
+        for (_, value) in self.iter() {
+            value.value.delete()
+        }
+
+        // Then delete the map itself.
+        // Unwrapping is safe when only using the high-level API.
+        self.state_api.delete_prefix(&self.prefix).unwrap_abort();
```
rimbi, line 2610:
```suggestion
    K: Serialize,
    V: Serial + DeserialWithState<S> + Deletable,
{
    fn delete(mut self) {
        self.clear();
```

rimbi: nitpick: I am not sure about the implications of the changes in the trait bounds but it's worth checking. The compiler did not complain.

concordium-std/src/prims.rs
```diff
@@ -63,15 +63,124 @@ extern "C" {
...
+    ///   identifier that can be passed to any of the entry methods.
+    pub(crate) fn state_iterator_next(iterator: u64) -> u64;
+
+    /// Delete the iterator, unlocking the subtree. Returns
+    /// - u64::MAX if the iterator does not exist.
+    /// - 0 if the iterator was already deleted
+    /// - 1 if the iterator was successfully deleted as a result of this call.
+    pub(crate) fn state_iterator_delete(iterator: u64) -> u32;
+
+    /// Get the length of the key that the iterator is currently pointing at.
```
rimbi, line 124: size or length?

concordium-std/src/prims.rs
```diff
@@ -89,6 +198,13 @@ extern "C" {
     /// Owner of the contract, AccountAddress.
     pub(crate) fn get_receive_owner(start: *mut u8);
 
+    /// Get the size of the entrypoint that was named.
```
rimbi, line 201:
```suggestion
    /// Get the size of the entrypoint name.
```

rimbi (commented):
All look very good. I have not found any major issues. Most of my comments are nitpicks. Please feel free to ignore them.
I have not had enough time to focus on the architecture, unfortunately.

concordium-std/src/impls.rs
```diff
@@ -213,8 +335,17 @@ impl Read for ContractState {
     /// initialize a dummy value before calling an external function.
     fn read_u32(&mut self) -> ParseResult<u32> {
```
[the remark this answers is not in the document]
Bargsteen: Due to optimization.

concordium-std/src/impls.rs
```diff
@@ -1,30 +1,37 @@
...
 impl convert::From<ParseError> for Reject {
     #[inline(always)]
     fn from(_: ParseError) -> Self {
         unsafe { num::NonZeroI32::new_unchecked(i32::MIN + 2) }.into()
     }
 }
 
-/// Full is mapped to i32::MIN+3, Malformed is mapped to i32::MIN+4.
+/// Full is mapped to i32::MIN + 3,
+/// Malformed is mapped to i32::MIN + 4.
```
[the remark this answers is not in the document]
abizjak: There actually is, because this makes it show up in documentation on docs.rs.
Screenshot from 2022-05-01 16-46-21

concordium-std/src/impls.rs
```diff
@@ -79,18 +86,71 @@ impl From<NewReceiveNameError> for Reject {
...
-/// low-level interface by manually using the [Write] trait's interface.
-impl Write for ReturnValue {
+/// AmountTooLarge is i32::MIN + 13,
+/// MissingAccount is i32::MIN + 14.
+impl From<TransferError> for Reject {
+    #[inline(always)]
+    fn from(te: TransferError) -> Self {
+        match te {
+            TransferError::AmountTooLarge => unsafe {
+                crate::num::NonZeroI32::new_unchecked(i32::MIN + 13).into()
```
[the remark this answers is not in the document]
abizjak: This is not a safe implementation, so we are not doing this. It is extremely unsafe to call this `from` with `0`. It breaks all compiler invariants about the Reject type.

concordium-std/src/impls.rs
```diff
@@ -213,8 +335,17 @@ impl Read for ContractState {
...
+        let num_read = unsafe {
+            prims::state_entry_read(
+                self.state_entry_id,
+                bytes.as_mut_ptr() as *mut u8,
+                4,
+                self.current_position,
+            )
+        };
+        if num_read == u32::MAX {
+            return Err(ParseError::default()); // Entry did not exist.
```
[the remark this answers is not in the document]
abizjak: Because code size and performance is so important, ParseError is a zero sized type. So there is exactly one value, the default one.

concordium-std/src/impls.rs
```diff
@@ -251,46 +391,1129 @@ impl Write for ContractState {
...
+                state_api:        self.state_api.clone(),
+                _lifetime_marker: PhantomData,
+            },
+            Err(StateError::SubtreeWithPrefixNotFound) => StateMapIterMut {
+                state_iter:       None,
+                state_api:        self.state_api.clone(),
+                _lifetime_marker: PhantomData,
+            },
+            _ => crate::trap(),
+        }
```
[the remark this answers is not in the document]
abizjak: I am not in favor of this. If we get any other error than SubTreeWithPrefixNotFound that is an invariant violation (i.e., somebody used a low-level API). We do not want to be silent about this. We want to fail hard.

concordium-std/src/impls.rs
```diff
@@ -251,46 +391,1129 @@ impl Write for ContractState {
...
+                state_api:        self.state_api.clone(),
+                _lifetime_marker: PhantomData,
+            },
+            Err(StateError::SubtreeWithPrefixNotFound) => StateMapIter {
+                state_iter:       None,
+                state_api:        self.state_api.clone(),
+                _lifetime_marker: PhantomData,
+            },
+            _ => crate::trap(),
+        }
```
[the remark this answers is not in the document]
abizjak: I am not in favor of this since it will hide any bugs or incorrect uses of the library. We want to fail hard in case we get an unexpected error, since that is either a bug in the library, or the user using low-level state.

concordium-std/src/impls.rs
```diff
@@ -251,46 +391,1129 @@ impl Write for ContractState {
...
+        {
+            if *modified {
+                entry.move_to_start();
+                value.serial(entry).unwrap_abort();
+            }
+        }
+    }
+}
+
+fn get_with_inner<'a, T: Serial + DeserialWithState<S>, S: HasStateApi>(
```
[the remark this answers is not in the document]
abizjak: Added

concordium-std/src/impls.rs
```diff
@@ -251,46 +391,1129 @@ impl Write for ContractState {
...
+                value,
+                ..
+            } => (entry, value),
+            StateBoxInner::Reference {
+                ..
+            } => {
+                // We just set it to loaded
+                unsafe { crate::hint::unreachable_unchecked() }
+            }
         }
```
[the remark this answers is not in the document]
abizjak: I personally think this is less clear and the control flow is very convoluted. You have also removed any explanation as to why the last statement is not reachable.

concordium-std/src/impls.rs
abizjak: I removed the todo.

abizjak (approved):

PR #94, opened 2022-04-03
Add DerefMut to OccupiedEntry and StateRefMut.

abizjak:
- Update the auction example to use it.

This is not clear-cut since it does have a performance penalty which might be significant.
However it makes for much easier use of nested data structures, without nested closures.

It is unclear whether we want this approach or not since it has a performance penalty, and is additionally really subtle.
In particular it is not entirely clear where rust guarantees that drop will run. I believe it should always run unless there are panics or users make weird loops with circular Rc's or some such. 

## Purpose

Improve ergonomics of map use.

## Changes

As the title states, DerefMut instances were added together with the accompanying `Drop` instances that make sure data is properly stored.

### DIFF
--- a/concordium-std/src/impls.rs
+++ b/concordium-std/src/impls.rs
@@ -560,6 +560,7 @@ where
         OccupiedEntry {
             key: self.key,
             value,
+            modified: false,
             state_entry,
             _lifetime_marker: self._lifetime_marker,
         }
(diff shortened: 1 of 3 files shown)


abizjak:
@chrmatt @EmilLa1 @Bargsteen Opinions?

EmilLa1:
I'm in doubt regarding this.
 
I think the subtleties (regarding cost) can be quite confusing for the user - and if not understood correctly it can lead to poorly written contracts with possible higher costs than ideal. But this could be the case anyway when using the `State`.

 
I think however that this will be very useful when using nested State(Map/Set)s as the nested closure's can be a bit daunting currently. 


I think the latter outweighs the former with proper documentation/examples.

chrmatt:
As said previously, I find it much easier to use with this functionality. The auction contract is a good example.

Regarding the subtle performance penalties, I don't quite understand where and why they occur. One thing seems that the state of something is stored again if DerefMut has been called, even if the value is not actually changed. I believe this is fine if documented clearly.

abizjak:
Thanks @chrmatt @EmilLa1  I'll go ahead with this after adding the same to the StateBox, and more documentation.

EmilLa1 (commented):
lgtm

PR #111, opened 2022-05-27
invoker variable should not be used for authentication in smar…

### LINKED ISSUE #110: tx.origin/invoker variable should not be used for authentication in smart contracts
**Description**
Solidity uses the variable `tx.origin` (the equivalent variable is `invoker` on the Concordium chain) to denote the address that initiated a tx.

Smart contract best practice guides advise NOT to use `tx.origin/invoker` for authentication/authorization in smart contracts because phishing attacks are possible with such a setup.
Additional reading [here](https://hackernoon.com/hacking-solidity-contracts-using-txorigin-for-authorization-are-vulnerable-to-phishing) and [here](https://consensys.github.io/smart-contract-best-practices/development-recommendations/solidity-specific/tx-origin/) 

Suggested changes to following [line](https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/examples/proxy/src/lib.rs#L86):
`claim_eq!(ctx.owner(), ctx.sender());`

Alternative suggestion:
I know that these smart contracts are not meant for production but I suspect new smart contract developers to kick start their projects based on these examples nonetheless. If we want to keep the `invoker` variable at this line then I would at least add a warning comment above it so people are aware of it.

Please remove the label "question" if it is ok that I work on it or close the issue to reject the suggested changes.

DOBEN:
## Purpose

`Invoker` variable should not be used for authentication in smart contracts due to phishing attacks.

closes #110 

## Changes

- FIXED: invoker variable should not be used for authentication in smart contracts

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/proxy/src/lib.rs
+++ b/examples/proxy/src/lib.rs
@@ -83,7 +83,7 @@ fn receive_reconfigure<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
     host: &mut impl HasHost<State, StateApiType = S>,
 ) -> ReceiveResult<()> {
-    claim_eq!(ctx.owner(), ctx.invoker());
+    ensure!(ctx.sender().matches_account(&ctx.owner()));
     *host.state_mut() = ctx.parameter_cursor().get()?;
     Ok(())
 }


abizjak (approved):
The change looks good.

However can you please use the PR template with the checklist and corresponding sections (removing the license stuff since you are an employee) in the future.

Issue #112, opened 2022-05-30
Returning funds immediately when outbidded by another bid (this will prevent having to use loops that can run out of gas)


DOBEN:
**Description**
`Loops` can run out of gas in smart contracts, therefore they are used very rarely in smart contracts (only when we can ensure that we are in control of the size of the object that the loop is iterating through).

Example when loops can be used safely:
If the owner of a smart contract adds addresses to a list, then it is perfectly fine to iterate via a for loop over this list since we can assume that the owner of the smart contract is a smart contract developer and knowledgable enough to not insert too many addresses in the list. 

Example when loops should be avoided:
If a general user calls our smart contract and this action causes an e.g. address to be inserted in a list then we don't have any control of the size of the list anymore. If our smart contract protocol becomes super popular the list can grow significantly. In addition, competitor protocols might on purpose call this function to blow up our list so that any loop that iterates through the list in our smart contract is now reverting (parts of our protocol become inexecutable).

Suggestion for the [auction example](https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/examples/auction/src/lib.rs):
Since the payout function of the auction example iterates via a for loop over an object that was grown in size by user actions, we should safeguard that the size of the object does not grow too large. This is to prevent the smart contract to be unable to pay out its rewards (causing the funds to be stuck in the smart contract).

1. STEP: Adding a variable `MAX_PARTICIPANTS` :

    **const MAX_PARTICIPANTS: u32 = 100;**

2. STEP: Adding a counter of how many addresses/users/bids already participated in our auction:

     /// Number of bids in the StateMap bids
    **number_of_bids: u32**
 
 3. STEP: Checkinging in the `bid` receive funciton that we are not inserting too many addresses
        **ensure!(MAX_PARTICIPANTS >= state.viewable_state.number_of_bids,
        BidError::MaxParticipantsReached);**
        
Please remove the label "question" if it is ok that I work on it or close the issue to reject the suggested changes.

abizjak:
I am not sure what if anything should be done here.

On the one hand this auction is not really useful in the current scope. It is a toy example so from that perspective I do not wish to complicate it more. It would make sense to maybe add a comment to this effect though.

[...]

PR #113, opened 2022-05-30
returning funds immediately when outbidded by another bid

DOBEN:
## Purpose

Closing #112

## Changes

- Update auction example

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/auction/src/lib.rs
+++ b/examples/auction/src/lib.rs
@@ -66,13 +66,13 @@ struct InitParameter {
 enum BidError {
     /// Raised when a contract tries to bid; Only accounts
     /// are allowed to bid.
-    ContractSender,
+    OnlyAccount,
     /// Raised when new bid amount is lower than current highest bid
     BidTooLow,
     /// Raised when bid is placed after auction end time passed
-    BidsOverWaitingForAuctionFinalization,
+    BidTooLate,
     /// Raised when bid is placed after auction has been finalized
-    AuctionFinalized,
+    AuctionAlreadyFinalized,
 }
 
 /// `finalize` function errors
@@ -81,7 +81,7 @@ enum FinalizeError {
     /// Raised when finalizing an auction before auction end time passed
     AuctionStillActive,
     /// Raised when finalizing an auction that is already finalized
-    AuctionFinalized,
+    AuctionAlreadyFinalized,
 }
 
 /// Init function that creates a new auction
@@ -112,15 +112,15 @@ fn auction_bid<S: HasStateApi>(
     let balance = host.self_balance();
     let state = host.state_mut();
     // Ensure the auction has not been finalized yet
-    ensure!(state.auction_state == AuctionState::NotSoldYet, BidError::AuctionFinalized);
+    ensure!(state.auction_state == AuctionState::NotSoldYet, BidError::AuctionAlreadyFinalized);
 
     let slot_time = ctx.metadata().slot_time();
     // Ensure the auction has not ended yet
-    ensure!(slot_time <= state.end, BidError::BidsOverWaitingForAuctionFinalization);
+    ensure!(slot_time <= state.end, BidError::BidTooLate);
 
     // Ensure that only accounts can place a bid
     let sender_address = match ctx.sender() {
-        Address::Contract(_) => bail!(BidError::ContractSender),
+        Address::Contract(_) => bail!(BidError::OnlyAccount),
         Address::Account(account_address) => account_address,
     };
 
@@ -133,6 +133,9 @@ fn auction_bid<S: HasStateApi>(
         // `account_address` exists since it was recorded when it placed a bid.
         // If an `account_address` exists, and the contract has the funds then the
         // transfer will always succeed.
+        // Please consider using a pull-over-push pattern when expanding this smart
+        // contract to allow smart contract instances to participate in the auction as
+        // well. https://consensys.github.io/smart-contract-best-practices/attacks/denial-of-service/
         let _ = host.invoke_transfer(&account_address, balance);
     }
     Ok(())
@@ -157,7 +160,10 @@ fn auction_finalize<S: HasStateApi>(
 ) -> Result<(), FinalizeError> {
     let state = host.state();
     // Ensure the auction has not been finalized yet
-    ensure!(state.auction_state == AuctionState::NotSoldYet, FinalizeError::AuctionFinalized);
+    ensure!(
+        state.auction_state == AuctionState::NotSoldYet,
+        FinalizeError::AuctionAlreadyFinalized
+    );
 
     let slot_time = ctx.metadata().slot_time();
     // Ensure the auction has ended already
@@ -334,7 +340,7 @@ mod tests {
         let finres3 = auction_finalize(&ctx5, &mut host);
         expect_error(
             finres3,
-            FinalizeError::AuctionFinalized,
+            FinalizeError::AuctionAlreadyFinalized,
             "Finalizing the auction a second time should fail",
         );
 
@@ -342,7 +348,7 @@ mod tests {
         let res4 = auction_bid(&bob_ctx, &mut host, big_amount);
         expect_error(
             res4,
-            BidError::AuctionFinalized,
+            BidError::AuctionAlreadyFinalized,
             "Bidding should fail because the auction is finalized",
         );
     }
--- a/examples/auction/src/lib.rs
+++ b/examples/auction/src/lib.rs
@@ -112,7 +112,7 @@ fn auction_bid<S: HasStateApi>(
     let balance = host.self_balance();
     let state = host.state_mut();
     // Ensure the auction has not been finalized yet
-    ensure!(state.auction_state == AuctionState::NotSoldYet, BidError::AuctionAlreadyFinalized);
+    ensure_eq!(state.auction_state, AuctionState::NotSoldYet, BidError::AuctionAlreadyFinalized);
 
     let slot_time = ctx.metadata().slot_time();
     // Ensure the auction has not ended yet
@@ -136,18 +136,18 @@ fn auction_bid<S: HasStateApi>(
         // Please consider using a pull-over-push pattern when expanding this smart
         // contract to allow smart contract instances to participate in the auction as
         // well. https://consensys.github.io/smart-contract-best-practices/attacks/denial-of-service/
-        let _ = host.invoke_transfer(&account_address, balance);
+        host.invoke_transfer(&account_address, balance).unwrap_abort();
     }
     Ok(())
 }
 
 /// View function that returns the content of the state
 #[receive(contract = "auction", name = "view", return_value = "State")]
-fn view<S: HasStateApi>(
-    _ctx: &impl HasReceiveContext,
-    host: &impl HasHost<State, StateApiType = S>,
-) -> ReceiveResult<State> {
-    Ok(host.state().clone())
+fn view<'a, 'b, S: HasStateApi>(
+    _ctx: &'a impl HasReceiveContext,
+    host: &'b impl HasHost<State, StateApiType = S>,
+) -> ReceiveResult<&'b State> {
+    Ok(host.state())
 }
 
 /// Receive function used to finalize the auction. It sends the highest bid (the
@@ -160,8 +160,9 @@ fn auction_finalize<S: HasStateApi>(
 ) -> Result<(), FinalizeError> {
     let state = host.state();
     // Ensure the auction has not been finalized yet
-    ensure!(
-        state.auction_state == AuctionState::NotSoldYet,
+    ensure_eq!(
+        state.auction_state,
+        AuctionState::NotSoldYet,
         FinalizeError::AuctionAlreadyFinalized
     );
 
@@ -180,7 +181,7 @@ fn auction_finalize<S: HasStateApi>(
         // `owner` exists since it deployed the smart contract instance.
         // If an account exists, and the contract has the funds then the
         // transfer will always succeed.
-        let _ = host.invoke_transfer(&owner, balance);
+        host.invoke_transfer(&owner, balance).unwrap_abort();
     }
     Ok(())
 }
--- a/examples/auction/src/lib.rs
+++ b/examples/auction/src/lib.rs
@@ -1,24 +1,23 @@
+//! # Implementation of an auction smart contract
+//!
+//! Account addresses can call the bid function to participate in the auction.
+//! An account address has to send some CCD when calling the bid function.
+//! This CCD amount has to exceed the current highest bid to be accepted by the
+//! smart contract.
+//!
+//! The smart contract keeps track of the current highest bidder as well as
+//! the CCD amount of the highest bid. The CCD balance of the smart contract
+//! represents the highest bid. When a new highest bid is accepted by the smart
+//! contract, the smart contract refunds the old highest bidder.
+//!
+//! Bids have to be placed before the auction ends. The participant with the
+//! highest bid (the last bidder) wins the auction.
+//!
+//! After the auction ends, any account can finalize the auction. The owner of
+//! the smart contract instance receives the highest bid (the balance of this
+//! contract) when the auction is finalized. This can be done only once.
 use concordium_std::*;
 use core::fmt::Debug;
-
-/// # Implementation of an auction smart contract
-///
-/// Users can call the bid function to participate in the auction.
-/// A user has to send some CCD when calling the bid function.
-/// This CCD amount has to exceed the current highest bid to be accepted by the
-/// smart contract.
-///
-/// The smart contract keeps track of the current highest bidder as well as
-/// the CCD amount of the highest bid. The CCD balance of this smart contract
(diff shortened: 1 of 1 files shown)


examples/auction/src/lib.rs
```diff
@@ -166,35 +177,24 @@ fn auction_finalize<S: HasStateApi>(
...
-        for (addr, amnt) in state.bids.iter() {
-            if *amnt < state.viewable_state.highest_bid {
-                if host.invoke_transfer(&*addr, *amnt).is_err() {
-                    bail!(FinalizeError::BidMapError);
+        } else {
+            match state.viewable_state.highest_bidder {
+                Some(account_address) => {
+                    host.state_mut().viewable_state.auction_state =
+                        AuctionState::Sold(account_address);
+                    Ok(())
```
abizjak, line 194: I would restructure this into something like
```rust
if let Some(account_address) = state.viewable_state.highest_bidder.take() {
    host.state_mut().viewable_state.auction_state = AuctionState::Sold(account_address);
    if host.invoke_transfer(&owner, balance).is_err() {
        ...
    }
}
```
so that you don't attempt a transfer when no bids are made. 

At the same time you can also get rid of the outer `if balance == zero` check since if there is a bid the amount must be non-zero.

examples/auction/src/lib.rs
```diff
@@ -130,13 +133,21 @@ fn auction_bid<S: HasStateApi>(
...
+        Some(account_address) => {
+            state.viewable_state.highest_bidder = Some(sender_address);
+            if host.invoke_transfer(&account_address, balance).is_err() {
+                bail!(BidError::RefundError);
+            }
+        }
+        None => {
+            state.viewable_state.highest_bidder = Some(sender_address);
+        }
+    }
```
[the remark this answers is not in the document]
DOBEN: addressed

examples/auction/src/lib.rs
DOBEN: addressed

abizjak:
> Suggestion: Can we please use `//` instead of `///` when commenting?
> 
> I replaced all `///` with `//` because the `fmt` tool can automatically format and break the lines based on the linewidth so that our block comments look much nicer compared to if we use `///` and manually do it.

This is not a matter of style. One are doc comments that would appear in generated documentation. `//` do not show up in generated documentation.

Please refrain from doing such changes, and revert the comments back to doc comments.

DOBEN:
Question for clarification:
Sending ETHER to an address can only fail on the Ethereum chain when the address is a smart contract and it implements no payable `fallback` function or no `receive` function. I would like to know if that is the same on the Concordium chain or if there are any other scenarios when a CCD transfer can fail.

I saw that you are using this pattern a lot:

if host.invoke_transfer(&*addr, *amnt).is_err() {
**bail!(FinalizeError::BidMapError);**
}

but I would rather not revert if any refund of CCD fails since it could be abused once people modify this smart contract and allow smart contract addresses to take part in the auction as well.

Currently, if a smart contract reverts when receiving CCD refunds then the funds are staying in the auction contract (and the next bidder is lucky since it receives more CCD). I think this is a better behavior than reverting.

examples/auction/src/lib.rs
```diff
@@ -3,206 +3,191 @@ use core::fmt::Debug;
...
-    // Ensure that the new bid exceeds the highest bid so far
-    ensure!(*bid_to_update > state.viewable_state.highest_bid, BidError::BidTooLow);
 
-    state.viewable_state.highest_bid = *bid_to_update;
+    // Ensure that the new bid exceeds the highest bid so far
+    ensure!(amount > balance, BidError::BidTooLow);
 
+    if let Some(account_address) = state.viewable_state.highest_bidder.replace(sender_address) {
+        // Refunding old highest bidder
+        if host.invoke_transfer(&account_address, balance).is_err() {}
```
limemloh, line 144: Could you add a comment with why not to reject the contract in this case

examples/auction/src/lib.rs
```diff
@@ -3,206 +3,191 @@ use core::fmt::Debug;
...
-    host: &impl HasHost<State<S>, StateApiType = S>,
+    host: &impl HasHost<State, StateApiType = S>,
 ) -> ReceiveResult<ViewableState> {
     Ok(host.state().viewable_state.clone())
 }
 
-/// Receive function used to finalize the auction, returning all bids to their
-/// senders, except for the winning bid
+/// Receive function used to finalize the auction. It sends the highest bid (the
+/// current balance of this smart contract) to the owner of this smart contract.
```
limemloh, line 159:
```suggestion
/// current balance of this smart contract) to the owner of the smart contract instance.
```

examples/auction/src/lib.rs
```diff
@@ -3,206 +3,191 @@ use core::fmt::Debug;
...
 #[receive(contract = "auction", name = "finalize", mutable)]
 fn auction_finalize<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
-    host: &mut impl HasHost<State<S>, StateApiType = S>,
+    host: &mut impl HasHost<State, StateApiType = S>,
 ) -> Result<(), FinalizeError> {
     let state = host.state();
+    // Ensure that auction has not been finalized yet
```
limemloh, line 166:
```suggestion
    // Ensure the auction has not been finalized yet
```

examples/auction/src/lib.rs
```diff
@@ -3,206 +3,191 @@ use core::fmt::Debug;
...
+    ensure!(slot_time > state.viewable_state.end, FinalizeError::AuctionStillActive);
+
+    if let Some(account_address) = state.viewable_state.highest_bidder {
+        // Marking the highest bid (the last bidder) as winner of the auction
+        host.state_mut().viewable_state.auction_state = AuctionState::Sold(account_address);
+        let owner = ctx.owner();
+        let balance = host.self_balance();
+        // Sending the highest bid (the balance of this contract) to the owner of this
+        // contract
+        if host.invoke_transfer(&owner, balance).is_err() {}
```
limemloh, line 183: Again, add some comment to why not to deal reject in case of an error.
I would probably have rejected the contract in this case.

examples/auction/src/lib.rs
```diff
@@ -275,46 +260,47 @@ mod tests {
     /// Test a sequence of bids and finalizations:
```
limemloh, line 260: IMO this unit test should be split into multiple or at least provide an error description to `claim_eq!`. 
If this unit test fails no useful error message is given.

examples/auction/src/lib.rs
```diff
@@ -3,206 +3,191 @@ use core::fmt::Debug;
 
 /// # Implementation of an auction smart contract
```
limemloh, line 4: The contract description should probably be a module level doc comment, which is done using 
```
//! 
```
for every line instead of `///`

examples/auction/src/lib.rs
```diff
@@ -3,206 +3,191 @@ use core::fmt::Debug;
...
 ///
-/// To bid, participants send CCD using the bid function.
-/// The participant with the highest bid wins the auction.
-/// Bids are to be placed before the auction end. After that, bids are refused.
-/// Only bids that exceed the highest bid are accepted.
-/// Bids are placed incrementally, i.e., an account's bid is considered
-/// to be the **sum** of all bids.
+/// Users can call the bid function to participate in the auction.
```
limemloh, line 6: In our documentation we are not really refering to the contract caller as _user_, often we use address, account or contract, where address is used when it could be either a contract or an account.

In this case the auction only allow bids from accounts, so maybe use account instead?

limemloh (changes requested):

examples/auction/src/lib.rs
```diff
@@ -3,206 +3,191 @@ use core::fmt::Debug;
...
+pub struct State {
     /// The part of the state that can be viewed
     viewable_state: ViewableState,
-    /// Keeping track of which account bid how much money
-    bids:           StateMap<AccountAddress, Amount, S>,
 }
 
-/// The part of the state to be viewed using `concordium-client contract invoke`
+/// This part of the state can be viewed by querying the node
+/// with the command `concordium-client contract show`.
```
Bargsteen, line 46: `concordium-client contract show` does not include the state/model for v1 contracts. The idea, instead, is to use `contract invoke` to invoke entrypoints that return values to be viewed.

examples/auction/src/lib.rs
DOBEN: I provided some error descriptions to the `claim_eq!` now. Let me know if you think I should divide the tests further into smaller pieces.

examples/auction/src/lib.rs
```diff
@@ -219,10 +212,10 @@ mod tests {
         assert_eq!(actual, err);
```
limemloh, line 212: I know you did not add this, but would you please change this to use `claim_eq!` instead, so the test [fails properly when targeting wasm](https://developer.concordium.software/en/mainnet/smart-contracts/tutorials/piggy-bank/testing.html#compiling-and-running-tests-in-wasm).
And the above line to use [`expect_err_report`](https://docs.rs/concordium-std/latest/concordium_std/trait.ExpectErrReport.html#tymethod.expect_err_report) for the same reason

examples/auction/src/lib.rs
limemloh: I am fine with this

limemloh (commented):
Looks good to me

limemloh (approved):

examples/auction/src/lib.rs
```diff
@@ -1,208 +1,201 @@
+//! # Implementation of an auction smart contract
+//!
+//! Account addresses can call the bid function to participate in the auction.
```
abizjak, line 3: I think the terminology we use is a bit different. We don't say that addresses call functions. We generally use "invoke" and it is not the address that does this, but an account. An address is a pointer to an account.

examples/auction/src/lib.rs
```diff
@@ -1,208 +1,201 @@
...
+//! An account address has to send some CCD when calling the bid function.
```
abizjak, line 4: Same here, an account is not the same as an address.

examples/auction/src/lib.rs
```diff
@@ -1,208 +1,201 @@
...
-pub struct State<S> {
+/// This state can be viewed by querying the node with the command
+/// `concordium-client contract invoke` using the `view` function as entrypoint.
+#[derive(Debug, Serial, Deserial)]
+pub struct State {
     /// The part of the state that can be viewed
     viewable_state: ViewableState,
-    /// Keeping track of which account bid how much money
-    bids:           StateMap<AccountAddress, Amount, S>,
 }
```
abizjak, line 42: State is the same as viewable state now, so it would be better to remove the indirection altogether and just have State.

examples/auction/src/lib.rs
```diff
@@ -1,208 +1,201 @@
...
+    // Ensure that the new bid exceeds the highest bid so far
+    ensure!(amount > balance, BidError::BidTooLow);
+
+    if let Some(account_address) = state.viewable_state.highest_bidder.replace(sender_address) {
+        // Refunding old highest bidder;
+        // This transfer (given enough NRG of course) always succeeds because the
+        // `account_address` exists since it was recorded when it placed a bid.
+        // If an `account_address` exists, and the contract has the funds then the
+        // transfer will always succeed.
+        if host.invoke_transfer(&account_address, balance).is_err() {}
```
abizjak, line 147: This is a strange way of doing it. if you want to ignore the error then the more canonical way would be
```rust
let _ = host.invoke_transfer(...)
```
I don't think I've ever seen a dummy if statement like you have it.

However I don't like silently ignoring errors (even if they should not happen). It is better to do
```
host.invoke_transfer(...).unwrap_abort()
```
which will terminate the contract in case the impossible happens.

abizjak (changes requested):

examples/auction/src/lib.rs
DOBEN: If I understand correctly “host.invoke_transfer(...).unwrap_abort()” is equivalent to “bail!(...)”. Meaning it will revert the tx in case the CCD transfer fails, isn’t it?

I am worried about [this scenario](https://consensys.github.io/smart-contract-best-practices/attacks/denial-of-service/). I suspect that people will use our auction example one day and change it to allow smart contracts to take part in the auction (in that case it would be a problem if we revert the “host.invoke_transfer(...)” on an error). The ideal solution would be a pull-over-push pattern but I don’t want to complicate the auction example even further, therefore I changed it to this option now. Please let me know if that is ok, or if you want to proceed differently.

let _ = host.invoke_transfer(...)

examples/auction/src/lib.rs
DOBEN: I updated the sentence and I support the use of `invoke` for storage mutative functions in general.

Suggestion:
I have seen other projects (e.g. Hyperledger Fabric) using `invoke` for txs that execute a storage mutative function on chain while using `query` for non-storage mutative functions that are only `simulated` locally (e.g. the `view` function in this auction example). Concordium uses the word `invoke` also in the context of `simulating` a function execution locally:
`Invoking an instance is not a transaction and it does not change the state of a contract.`
We could improve on the use of `invoke` to avoid confusion.

DOBEN:
I would like to understand better the order of execution of the Rust smart contract code with the following syntax:
 
```
if let Some(account_address) = state.highest_bidder.replace(sender_address) {
      let _ = host.invoke_transfer(&account_address, balance);
}
```

I assume the following order:
1. Step `account_address` is set to `old highest_bidder`.
2. Step `state.highest_bidder` is set to `sender_address`.
3. Step `balance` is sent to `account_address`.

I just want to confirm it is not this order (which needs to be handled for re-entrance attacks in case we use this syntax sending funds to smart contracts one day.
1. Step `account_address` is set to `old highest_bidder`.
2. Step `balance` is sent to `account_address`.
3. Step `state.highest_bidder` is set to `sender_address`.

examples/auction/src/lib.rs
```diff
@@ -1,212 +1,197 @@
...
-#[receive(contract = "auction", name = "view", return_value = "ViewableState")]
+/// View function that returns the content of the state
+#[receive(contract = "auction", name = "view", return_value = "State")]
 fn view<S: HasStateApi>(
     _ctx: &impl HasReceiveContext,
-    host: &impl HasHost<State<S>, StateApiType = S>,
-) -> ReceiveResult<ViewableState> {
-    Ok(host.state().viewable_state.clone())
+    host: &impl HasHost<State, StateApiType = S>,
+) -> ReceiveResult<State> {
```
abizjak, line 149: In principle returning `&State` here should work (although you might need a few lifetime annotations) and would improve performance. I'm not sure we have an implementation of Serial for `&A` though. If not, it would be useful to have it in cases like this. So an implementation like

```rust
impl <'a, A: Serial> Serial for &'a A {

}
```

examples/auction/src/lib.rs
```diff
@@ -1,212 +1,197 @@
...
-    host: &mut impl HasHost<State<S>, StateApiType = S>,
+    host: &mut impl HasHost<State, StateApiType = S>,
 ) -> Result<(), FinalizeError> {
     let state = host.state();
+    // Ensure the auction has not been finalized yet
     ensure!(
-        state.viewable_state.auction_state == AuctionState::NotSoldYet,
-        FinalizeError::AuctionFinalized
+        state.auction_state == AuctionState::NotSoldYet,
+        FinalizeError::AuctionAlreadyFinalized
```
abizjak, line 165: Recommend you use `ensure_eq` instead. It is a better macro to use when testing equality. The error messages during testing are more useful.

abizjak (commented):
I have one further comment, and two nitpicks. 

The general issue of silently failing is rather important to me and I believe we should always panic/abort if an impossible happens, rather than silently ignoring an issue. This is the pattern we've used in the development of the libraries.

abizjak (approved):

Bargsteen:
> I would like to understand better the order of execution of the Rust smart contract code with the following syntax:
> 
> ```
> if let Some(account_address) = state.highest_bidder.replace(sender_address) {
>       let _ = host.invoke_transfer(&account_address, balance);
> }
> ```
> 
> I assume the following order:
> 
>     1. Step `account_address` is set to `old highest_bidder`.
> 
>     2. Step `state.highest_bidder` is set to `sender_address`.
> 
>     3. Step `balance` is sent to `account_address`.
> 
> 
> I just want to confirm it is not this order (which needs to be handled for re-entrance attacks in case we use this syntax sending funds to smart contracts one day).
> 
>     1. Step `account_address` is set to `old highest_bidder`.
> 
>     2. Step `balance` is sent to `account_address`.
> 
>     3. Step `state.highest_bidder` is set to `sender_address`.

The first ordering you mention is almost correct.
What happens is:
1. `state.highest_bidder` is set to `Some(sender_address)` with [Option::replace](https://doc.rust-lang.org/std/option/enum.Option.html#method.replace) which returns `-> Option<AccountAddress>`.
2. `account_address` is set to the output of the `replace` function, i.e. `old highest_bidder` 
3. `balance` is sent to `account_address` (if an previous bidder existed)

PR #128, opened 2022-06-28
upgradable wCCD token 

### LINKED ISSUE #129: upgradable wCCD token
**Task description**

Most smart contract protocols have some common features such as pausing and upgrading. This task will add the following features to the wCCD token:

**Sub-tasks**

- [ ] Adding an admin address that is authorized to do maintenance tasks in the smart contract. https://github.com/Concordium/concordium-rust-smart-contracts/pull/128

- [ ] Adding the capability to pause contract by the admin address. https://github.com/Concordium/concordium-rust-smart-contracts/pull/128

- [ ] Adding pattern to upgrade the implementation contract. https://github.com/Concordium/concordium-rust-smart-contracts/pull/128

DOBEN:
## Purpose

Implementing an upgradable wCCD token 

closes #129 

## Changes

- [x] Adding admin addresses that are authorized to do maintenance tasks on the smart contracts.

- [x] Adding the capability to pause the contract by the admin address.
- [x] Adding a pattern to deploy and link up the `proxy`, `implementation` and `state` contracts.
- [x] Adding a pattern to upgrade the implementation contract.
- [x] Adding a pattern that all wccd functions (e.g. `wrap`, `unwrap`, `transfer`, `updateOperator`, `balanceOf`, `operatorOf`, `supports` and `tokenMetadata`) have to be invoked on the `proxy` contract.
- [x] Adding a pattern to store ccd funds on proxy.
- [x] Adding events/logs to proxy
- [x] URL can be set by admin.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/cis2-wccd-upgradable/src/lib.rs
+++ b/examples/cis2-wccd-upgradable/src/lib.rs
@@ -156,25 +156,43 @@ struct AddressState<S> {
 #[derive(Serial, DeserialWithState)]
 #[concordium(state_parameter = "S")]
 struct State<S> {
-    /// Address of the w_ccd proxy contract.
-    proxy_address:          Option<ContractAddress>,
-    /// Address of the w_ccd implementation contract.
-    implementation_address: Option<ContractAddress>,
+    /// Addresses of the protocol
+    protocol_addresses: ProtocolAddressesState,
     /// The state of the one token.
-    token:                  StateMap<Address, AddressState<S>, S>,
+    token:              StateMap<Address, AddressState<S>, S>,
     /// Contract is paused/unpaused.
-    paused:                 bool,
+    paused:             bool,
+}
+
+#[derive(Serialize, PartialEq)]
+enum ProtocolAddressesState {
+    UnInitialized,
+    Initialized {
+        /// Address of the w_ccd proxy contract.
+        proxy_address:          ContractAddress,
+        /// Address of the w_ccd implementation contract.
+        implementation_address: ContractAddress,
+    },
 }
 
 /// The `implementation` contract state.
 #[derive(Serial, Deserial, SchemaType)]
 struct StateImplementation {
     /// The admin address can pause/unpause the contract.
-    admin:         Address,
-    /// Address of the w_ccd proxy contract.
-    proxy_address: Option<ContractAddress>,
-    /// Address of the w_ccd state contract.
-    state_address: Option<ContractAddress>,
+    admin:              Address,
+    /// Addresses of the protocol
+    protocol_addresses: ProtocolAddressesImplementation,
+}
+
+#[derive(SchemaType, Serialize, PartialEq)]
+enum ProtocolAddressesImplementation {
+    UnInitialized,
+    Initialized {
+        /// Address of the w_ccd proxy contract.
+        proxy_address: ContractAddress,
+        /// Address of the w_ccd state contract.
+        state_address: ContractAddress,
+    },
 }
 
 /// The `proxy` contract state.
@@ -361,6 +379,8 @@ struct ReturnBasicState {
     proxy_address:          ContractAddress,
     /// Address of the w_ccd implementation contract.
     implementation_address: ContractAddress,
+    /// Contract is paused/unpaused.
+    paused:                 bool,
 }
 
 /// Update struct.
@@ -391,7 +411,7 @@ enum CustomContractError {
     /// Contract already initialized.
     AlreadyInitialized,
     /// Contract not initialized.
-    NotInitialized,
+    UnInitialized,
     /// Only implementation contract.
     OnlyImplementation,
     /// Only proxy contract.
@@ -439,10 +459,9 @@ impl<S: HasStateApi> State<S> {
     fn new(state_builder: &mut StateBuilder<S>) -> Self {
         // Setup state.
         State {
-            proxy_address:          None,
-            implementation_address: None,
-            token:                  state_builder.new_map(),
-            paused:                 false,
+            protocol_addresses: ProtocolAddressesState::UnInitialized,
+            token:              state_builder.new_map(),
+            paused:             false,
         }
     }
 }
@@ -456,8 +475,7 @@ impl StateImplementation {
         // Setup state.
         StateImplementation {
             admin,
-            proxy_address: None,
-            state_address: None,
+            protocol_addresses: ProtocolAddressesImplementation::UnInitialized,
         }
     }
 
@@ -474,10 +492,15 @@ impl StateImplementation {
         // Setup parameter.
         let parameter_bytes = to_bytes(owner);
 
-        let state_address = host
-            .state()
-            .state_address
-            .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
+        let state_address = if let ProtocolAddressesImplementation::Initialized {
+            proxy_address: _,
+            state_address,
+        } = host.state().protocol_addresses
+        {
+            state_address
+        } else {
+            panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+        };
 
         let balance = host.invoke_contract_raw_read_only(
             &state_address,
@@ -510,10 +533,15 @@ impl StateImplementation {
             address: *address,
         });
 
-        let state_address = host
-            .state()
-            .state_address
-            .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
+        let state_address = if let ProtocolAddressesImplementation::Initialized {
+            proxy_address: _,
+            state_address,
+        } = host.state().protocol_addresses
+        {
+            state_address
+        } else {
+            panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+        };
 
         let is_operator = host.invoke_contract_raw_read_only(
             &state_address,
@@ -541,7 +569,7 @@ fn contract_proxy_log_mint_event<S: HasStateApi>(
     logger: &mut impl HasLogger,
 ) -> ContractResult<()> {
     // Only implementation can log event.
-    only_implementation(Some(host.state().implementation_address), ctx.sender())?;
+    only_implementation(host.state().implementation_address, ctx.sender())?;
 
     let params: MintEventParams = ctx.parameter_cursor().get()?;
 
@@ -563,7 +591,7 @@ fn contract_proxy_log_burn_event<S: HasStateApi>(
     logger: &mut impl HasLogger,
 ) -> ContractResult<()> {
     // Only implementation can log event.
-    only_implementation(Some(host.state().implementation_address), ctx.sender())?;
+    only_implementation(host.state().implementation_address, ctx.sender())?;
 
     let params: BurnEventParams = ctx.parameter_cursor().get()?;
 
@@ -585,7 +613,7 @@ fn contract_proxy_log_transfer_event<S: HasStateApi>(
     logger: &mut impl HasLogger,
 ) -> ContractResult<()> {
     // Only implementation can log event.
-    only_implementation(Some(host.state().implementation_address), ctx.sender())?;
+    only_implementation(host.state().implementation_address, ctx.sender())?;
 
     let params: TransferEventParams = ctx.parameter_cursor().get()?;
 
@@ -608,7 +636,7 @@ fn contract_proxy_log_update_operator_event<S: HasStateApi>(
     logger: &mut impl HasLogger,
 ) -> ContractResult<()> {
     // Only implementation can log event.
-    only_implementation(Some(host.state().implementation_address), ctx.sender())?;
+    only_implementation(host.state().implementation_address, ctx.sender())?;
 
     let params: UpdateOperatorEventParams = ctx.parameter_cursor().get()?;
 
@@ -689,14 +717,20 @@ fn contract_state_initialize<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
     host: &mut impl HasHost<State<S>, StateApiType = S>,
 ) -> ContractResult<()> {
-    ensure!(
-        host.state().proxy_address.is_none() && host.state().implementation_address.is_none(),
+    ensure_eq!(
+        host.state().protocol_addresses,
+        ProtocolAddressesState::UnInitialized,
         concordium_cis2::Cis2Error::Custom(CustomContractError::AlreadyInitialized)
     );
+
     // Set proxy and implementation addresses.
     let params: InitializeStateParams = ctx.parameter_cursor().get()?;
-    host.state_mut().proxy_address = Some(params.proxy_address);
-    host.state_mut().implementation_address = Some(params.implementation_address);
+
+    host.state_mut().protocol_addresses = ProtocolAddressesState::Initialized {
+        proxy_address:          params.proxy_address,
+        implementation_address: params.implementation_address,
+    };
+
     Ok(())
 }
 
@@ -713,14 +747,20 @@ fn contract_initialize<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
     host: &mut impl HasHost<StateImplementation, StateApiType = S>,
 ) -> ContractResult<()> {
-    ensure!(
-        host.state().proxy_address.is_none() && host.state().state_address.is_none(),
+    ensure_eq!(
+        host.state().protocol_addresses,
+        ProtocolAddressesImplementation::UnInitialized,
         concordium_cis2::Cis2Error::Custom(CustomContractError::AlreadyInitialized)
     );
+
     // Set proxy and storage addresses.
     let params: InitializeImplementationParams = ctx.parameter_cursor().get()?;
-    host.state_mut().proxy_address = Some(params.proxy_address);
-    host.state_mut().state_address = Some(params.state_address);
+
+    host.state_mut().protocol_addresses = ProtocolAddressesImplementation::Initialized {
+        proxy_address: params.proxy_address,
+        state_address: params.state_address,
+    };
+
     Ok(())
 }
 
@@ -842,11 +882,9 @@ fn receive_fallback<S: HasStateApi>(
 // or the proxy.
 
 fn only_implementation(
-    implementation: Option<ContractAddress>,
+    implementation_address: ContractAddress,
     sender: Address,
 ) -> ContractResult<()> {
-    let implementation_address = implementation
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
     ensure!(
         sender.matches_contract(&implementation_address),
         concordium_cis2::Cis2Error::Custom(CustomContractError::OnlyImplementation)
@@ -855,9 +893,7 @@ fn only_implementation(
     Ok(())
 }
 
-fn only_proxy(proxy: Option<ContractAddress>, sender: Address) -> ContractResult<()> {
-    let proxy_address =
-        proxy.ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
+fn only_proxy(proxy_address: ContractAddress, sender: Address) -> ContractResult<()> {
     ensure!(
         sender.matches_contract(&proxy_address),
         concordium_cis2::Cis2Error::Custom(CustomContractError::OnlyProxy)
@@ -874,8 +910,18 @@ fn contract_state_set_paused<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
     host: &mut impl HasHost<State<S>, StateApiType = S>,
 ) -> ContractResult<()> {
+    let implementation_address = if let ProtocolAddressesState::Initialized {
+        proxy_address: _,
+        implementation_address,
+    } = host.state().protocol_addresses
+    {
+        implementation_address
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
+
     // Only implementation can set state.
-    only_implementation(host.state().implementation_address, ctx.sender())?;
+    only_implementation(implementation_address, ctx.sender())?;
 
     // Set paused.
     let params: SetPausedParams = ctx.parameter_cursor().get()?;
@@ -896,12 +942,27 @@ fn contract_state_set_implementation_address<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
     host: &mut impl HasHost<State<S>, StateApiType = S>,
 ) -> ContractResult<()> {
+    let proxy_address = if let ProtocolAddressesState::Initialized {
+        proxy_address,
+        implementation_address: _,
+    } = host.state().protocol_addresses
+    {
+        proxy_address
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
+
     // Only proxy can update the implementation address.
-    only_proxy(host.state().proxy_address, ctx.sender())?;
+    only_proxy(proxy_address, ctx.sender())?;
 
     // Set implementation address.
     let params: SetImplementationAddressParams = ctx.parameter_cursor().get()?;
-    host.state_mut().implementation_address = Some(params.implementation_address);
+
+    host.state_mut().protocol_addresses = ProtocolAddressesState::Initialized {
+        proxy_address,
+        implementation_address: params.implementation_address,
+    };
+
     Ok(())
 }
 
@@ -916,8 +977,18 @@ fn contract_state_set_balance<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
     host: &mut impl HasHost<State<S>, StateApiType = S>,
 ) -> ContractResult<()> {
+    let implementation_address = if let ProtocolAddressesState::Initialized {
+        proxy_address: _,
+        implementation_address,
+    } = host.state().protocol_addresses
+    {
+        implementation_address
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
+
     // Only implementation can set state.
-    only_implementation(host.state().implementation_address, ctx.sender())?;
+    only_implementation(implementation_address, ctx.sender())?;
 
     let (state, state_builder) = host.state_and_builder();
 
@@ -948,8 +1019,18 @@ fn contract_state_set_operator<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
     host: &mut impl HasHost<State<S>, StateApiType = S>,
 ) -> ContractResult<()> {
+    let implementation_address = if let ProtocolAddressesState::Initialized {
+        proxy_address: _,
+        implementation_address,
+    } = host.state().protocol_addresses
+    {
+        implementation_address
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
+
     // Only implementation can set state.
-    only_implementation(host.state().implementation_address, ctx.sender())?;
+    only_implementation(implementation_address, ctx.sender())?;
 
     // Set balance.
     let params: SetIsOperatorParams = ctx.parameter_cursor().get()?;
@@ -1060,7 +1141,7 @@ fn contract_proxy_transfer_ccd<S: HasStateApi>(
     host: &mut impl HasHost<StateProxy, StateApiType = S>,
 ) -> ContractResult<()> {
     // Only implementation can access ccds on proxy.
-    only_implementation(Some(host.state().implementation_address), ctx.sender())?;
+    only_implementation(host.state().implementation_address, ctx.sender())?;
 
     let params: TransferCCDParams = ctx.parameter_cursor().get()?;
 
@@ -1133,59 +1214,55 @@ fn contract_state_view<S: HasStateApi>(
     _ctx: &impl HasReceiveContext,
     host: &impl HasHost<State<S>, StateApiType = S>,
 ) -> ContractResult<ReturnBasicState> {
-    let proxy_address = host
-        .state()
-        .proxy_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
-
-    let implementation_address = host
-        .state()
-        .implementation_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
+    let (proxy_address, implementation_address) = if let ProtocolAddressesState::Initialized {
+        proxy_address,
+        implementation_address,
+    } = host.state().protocol_addresses
+    {
+        (proxy_address, implementation_address)
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
 
     let state = ReturnBasicState {
         proxy_address,
         implementation_address,
+        paused: host.state().paused,
     };
     Ok(state)
 }
 
 /// Function to view state of contract.
 #[receive(contract = "CIS2-wCCD", name = "view", return_value = "StateImplementation")]
-fn contract_implementation_view<S: HasStateApi>(
-    _ctx: &impl HasReceiveContext,
-    host: &impl HasHost<StateImplementation, StateApiType = S>,
-) -> ContractResult<StateImplementation> {
-    let state_implementation = StateImplementation {
-        admin:         host.state().admin,
-        proxy_address: host.state().proxy_address,
-        state_address: host.state().state_address,
-    };
-    Ok(state_implementation)
+fn contract_implementation_view<'a, 'b, S: HasStateApi>(
+    _ctx: &'b impl HasReceiveContext,
+    host: &'a impl HasHost<StateImplementation, StateApiType = S>,
+) -> ContractResult<&'a StateImplementation> {
+    Ok(host.state())
 }
 
 /// Function to view state of contract.
 #[receive(contract = "CIS2-wCCD-Proxy", name = "view", return_value = "StateProxy")]
-fn contract_proxy_view<S: HasStateApi>(
-    _ctx: &impl HasReceiveContext,
-    host: &impl HasHost<StateProxy, StateApiType = S>,
-) -> ContractResult<StateProxy> {
-    let state_proxy = StateProxy {
-        admin:                  host.state().admin,
-        implementation_address: host.state().implementation_address,
-        state_address:          host.state().state_address,
-    };
-    Ok(state_proxy)
+fn contract_proxy_view<'a, 'b, S: HasStateApi>(
+    _ctx: &'b impl HasReceiveContext,
+    host: &'a impl HasHost<StateProxy, StateApiType = S>,
+) -> ContractResult<&'a StateProxy> {
+    Ok(host.state())
 }
 
 /// Helper function to ensure contract is not paused.
 fn when_not_paused<S>(
     host: &mut impl HasHost<StateImplementation, StateApiType = S>,
 ) -> ContractResult<()> {
-    let state_address = host
-        .state()
-        .state_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
+    let state_address = if let ProtocolAddressesImplementation::Initialized {
+        proxy_address: _,
+        state_address,
+    } = host.state().protocol_addresses
+    {
+        state_address
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
 
     let paused = host.invoke_contract_raw_read_only(
         &state_address,
@@ -1213,8 +1290,18 @@ fn contract_wrap<S: HasStateApi>(
     host: &mut impl HasHost<StateImplementation, StateApiType = S>,
     amount: Amount,
 ) -> ContractResult<()> {
+    let (proxy_address, state_address) = if let ProtocolAddressesImplementation::Initialized {
+        proxy_address,
+        state_address,
+    } = host.state().protocol_addresses
+    {
+        (proxy_address, state_address)
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
+
     // Can be only called through the fallback function on the proxy.
-    only_proxy(host.state().proxy_address, ctx.sender())?;
+    only_proxy(proxy_address, ctx.sender())?;
 
     // Check that contract is not paused.
     when_not_paused(host)?;
@@ -1228,11 +1315,6 @@ fn contract_wrap<S: HasStateApi>(
     let sender = input.sender;
 
     let receive_address = input.params.to.address();
-
-    let proxy_address = host
-        .state()
-        .proxy_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::ProxyInvokeError))?;
 
     // Proxy holds CCD funds. CCD is sent to proxy.
     host.invoke_contract_raw(
@@ -1251,11 +1333,6 @@ fn contract_wrap<S: HasStateApi>(
     };
 
     let parameter_bytes = to_bytes(&set_balance_params);
-
-    let state_address = host
-        .state()
-        .state_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
 
     // Update state.
     host.invoke_contract_raw(
@@ -1322,8 +1399,18 @@ fn contract_unwrap<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
     host: &mut impl HasHost<StateImplementation, StateApiType = S>,
 ) -> ContractResult<()> {
+    let (proxy_address, state_address) = if let ProtocolAddressesImplementation::Initialized {
+        proxy_address,
+        state_address,
+    } = host.state().protocol_addresses
+    {
+        (proxy_address, state_address)
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
+
     // Can be only called through the fallback function on the proxy.
-    only_proxy(host.state().proxy_address, ctx.sender())?;
+    only_proxy(proxy_address, ctx.sender())?;
 
     // Check that contract is not paused.
     when_not_paused(host)?;
@@ -1360,11 +1447,6 @@ fn contract_unwrap<S: HasStateApi>(
 
     let parameter_bytes = to_bytes(&set_balance_params);
 
-    let state_address = host
-        .state()
-        .state_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
-
     host.invoke_contract_raw(
         &state_address,
         Parameter(&parameter_bytes),
@@ -1377,11 +1459,6 @@ fn contract_unwrap<S: HasStateApi>(
     };
 
     let parameter_bytes = to_bytes(&params2);
-
-    let proxy_address = host
-        .state()
-        .proxy_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
 
     // Access/Request CCD funds from proxy.
     host.invoke_contract_raw(
@@ -1445,8 +1522,18 @@ fn contract_transfer<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
     host: &mut impl HasHost<StateImplementation, StateApiType = S>,
 ) -> ContractResult<()> {
+    let (proxy_address, state_address) = if let ProtocolAddressesImplementation::Initialized {
+        proxy_address,
+        state_address,
+    } = host.state().protocol_addresses
+    {
+        (proxy_address, state_address)
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
+
     // Can be only called through the fallback function on the proxy.
-    only_proxy(host.state().proxy_address, ctx.sender())?;
+    only_proxy(proxy_address, ctx.sender())?;
 
     // Check that contract is not paused.
     when_not_paused(host)?;
@@ -1455,16 +1542,6 @@ fn contract_transfer<S: HasStateApi>(
     let input: TransferParameterWithSender = ctx.parameter_cursor().get()?;
 
     let TransferParams(transfers) = input.params;
-
-    let state_address = host
-        .state()
-        .state_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
-
-    let proxy_address = host
-        .state()
-        .proxy_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
 
     for Transfer {
         token_id,
@@ -1574,8 +1651,18 @@ fn contract_update_operator<S: HasStateApi>(
     ctx: &impl HasReceiveContext,
     host: &mut impl HasHost<StateImplementation, StateApiType = S>,
 ) -> ContractResult<()> {
+    let (proxy_address, state_address) = if let ProtocolAddressesImplementation::Initialized {
+        proxy_address,
+        state_address,
+    } = host.state().protocol_addresses
+    {
+        (proxy_address, state_address)
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
+
     // Can be only called through the fallback function on the proxy.
-    only_proxy(host.state().proxy_address, ctx.sender())?;
+    only_proxy(proxy_address, ctx.sender())?;
 
     // Check that contract is not paused.
     when_not_paused(host)?;
@@ -1584,16 +1671,6 @@ fn contract_update_operator<S: HasStateApi>(
     let input: UpdateOperatorParamsWithSender = ctx.parameter_cursor().get()?;
 
     let UpdateOperatorParams(params) = input.params;
-
-    let state_address = host
-        .state()
-        .state_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
-
-    let proxy_address = host
-        .state()
-        .proxy_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
 
     for param in params {
         // Get the sender who invoked this contract function.
@@ -1740,10 +1817,15 @@ fn contract_pause<S: HasStateApi>(
         paused: true,
     });
 
-    let state_address = host
-        .state()
-        .state_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
+    let state_address = if let ProtocolAddressesImplementation::Initialized {
+        proxy_address: _,
+        state_address,
+    } = host.state().protocol_addresses
+    {
+        state_address
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
 
     host.invoke_contract_raw(
         &state_address,
@@ -1768,10 +1850,15 @@ fn contract_un_pause<S: HasStateApi>(
         paused: false,
     });
 
-    let state_address = host
-        .state()
-        .state_address
-        .ok_or(concordium_cis2::Cis2Error::Custom(CustomContractError::NotInitialized))?;
+    let state_address = if let ProtocolAddressesImplementation::Initialized {
+        proxy_address: _,
+        state_address,
+    } = host.state().protocol_addresses
+    {
+        state_address
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
+    };
 
     host.invoke_contract_raw(
         &state_address,
@@ -2023,13 +2110,10 @@ mod tests {
         // Check the result
         let state = result.expect_report("Contract w_ccd state initialization failed");
 
-        // Check the state
-        claim_eq!(state.proxy_address, None, "Proxy address should not be initialized");
-        claim_eq!(
-            state.implementation_address,
-            None,
-            "Implementation address should not be initialized"
-        );
+        if let ProtocolAddressesState::UnInitialized = state.protocol_addresses {
+        } else {
+            claim!(false, "Implementation address should not be initialized");
+        };
 
         let mut host = TestHost::new(state, builder);
 
@@ -2052,12 +2136,21 @@ mod tests {
         claim!(result.is_ok(), "Results in rejection");
 
         // Check the state.
-        claim_eq!(host.state().proxy_address, Some(PROXY), "Proxy address should now be set");
-        claim_eq!(
-            host.state().implementation_address,
-            Some(IMPLEMENTATION),
-            "Implementation address should now be set"
-        );
+        if let ProtocolAddressesState::Initialized {
+            implementation_address,
+            proxy_address,
+        } = host.state().protocol_addresses
+        {
+            claim_eq!(
+                implementation_address,
+                IMPLEMENTATION,
+                "Implementation address should now be set"
+            );
+
+            claim_eq!(proxy_address, PROXY, "Proxy address should now be set");
+        } else {
+            claim!(false, "Implementation address should now be set");
+        };
 
         // Can not initialize again.
         let result: ContractResult<()> = contract_state_initialize(&ctx, &mut host);
--- a/examples/cis2-wccd-upgradable/src/lib.rs
+++ b/examples/cis2-wccd-upgradable/src/lib.rs
@@ -28,10 +28,11 @@
 //! The protocol consists of three smart contracts (`proxy`, `implementation`,
 //! and `state`). All wccd functions (e.g. `wrap`, `unwrap`, `transfer`,
 //! `updateOperator`, `balanceOf`, `operatorOf` and `tokenMetadata`) have to be
-//! invoked on the `proxy` contract. The `proxy` will set the `current_sender`
-//! on the `implementation` before invoking these functions. Invoking the wccd
-//! functions directly on the `implementation` without going through the
-//! `proxy` fallback function will revert.
+//! invoked on the `proxy` contract. The `proxy` will append the
+//! `current_sender` to the input parameters so that the `implementation` can
+//! retrieve this information. Invoking the wccd functions directly on the
+//! `implementation` without going through the `proxy` fallback function will
+//! revert.
 
 //! The admin address on the `proxy` can upgrade the protocol with
 //! a new `implementation` contract. The admin address on the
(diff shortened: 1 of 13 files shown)


examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+struct StateImplementation<S> {
+    /// The admin address can pause/unpause the contract.
+    admin:         Address,
+    /// Address of the w_ccd proxy contract.
+    proxy_address: Option<ContractAddress>,
+    /// Address of the w_ccd state contract.
+    state_address: Option<ContractAddress>,
+    /// FALSE if proxy and state addresses have not been set yet.
+    /// TRUE if proxy and state addresses have been set.
+    initialized:   bool,
```
limemloh, line 70: Whether the state contract is initialized can be derived from the above options.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+    data: AdditionalData,
+}
+
+/// The parameter type for the state contract function `initialize`.
+#[derive(Serialize, SchemaType)]
+struct InitializeStateParams {
+    /// Address of the w_ccd proxy contract.
+    proxy_address:          Option<ContractAddress>,
+    /// Address of the w_ccd implementation contract.
+    implementation_address: Option<ContractAddress>,
```
limemloh, line 114: Are these suppose to be options?
Would it make sense to initialize with any of these as None?

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+    host: &mut impl HasHost<StateImplementation<S>, StateApiType = S>,
+    amount: Amount,
+    logger: &mut impl HasLogger,
+) -> ContractResult<()> {
+    let slot_time = ctx.metadata().slot_time();
+
+    match host.state().state_address {
+        None => bail!(concordium_cis2::Cis2Error::Custom(OnlyState)),
+
+        Some(state_address) => {
```
limemloh, line 522: You can avoid the match statement and extra indentation by using [`ok_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else):
```rust
    let state_address = host.state().state_address.ok_or_else(|| concordium_cis2::Cis2Error::Custom(OnlyState))?;
```

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+                    EntrypointName::new_unchecked("get_unpause_time"),
+                    Amount::zero(),
+                )?
+                .1;
+
+            let unpause_time = if let Some(mut unpause_time) = unpause_time {
+                unpause_time.get()?
+            } else {
+                return Err(concordium_cis2::Cis2Error::Custom(StateInvokeError));
+            };
```
limemloh, line 536: using `ok_or_else` would simplify this as well

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+    ctx: &impl HasReceiveContext,
+    host: &mut impl HasHost<StateImplementation<S>, StateApiType = S>,
+) -> ContractResult<()> {
+    // Check that only the current admin can pause.
+    ensure_eq!(ctx.sender(), host.state().admin, ContractError::Unauthorized);
+    // Parse the parameter.
+    let unpause_time = ctx.parameter_cursor().get()?;
+    let slot_time = ctx.metadata().slot_time();
+    // Update unpause_time.
+    match slot_time.checked_add(unpause_time) {
```
limemloh, line 865: Again it would simplify using `ok_or_else`

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,1470 @@
+//! upgradable wCCD token
+
+#![cfg_attr(not(feature = "std"), no_std)]
+use crate::CustomContractError::{ContractPaused, OutOfBound};
+use concordium_cis2::*;
+use concordium_std::*;
+use core::fmt::Debug;
```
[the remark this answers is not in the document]
abizjak: I think `concordium_std` re-exports `fmt` so it should be imported from there so that it works both with and without std.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+    fn balance(
+        &self,
+        token_id: &ContractTokenId,
+        address: &Address,
+    ) -> ContractResult<ContractTokenAmount> {
+        ensure_eq!(token_id, &TOKEN_ID_WCCD, ContractError::InvalidTokenId);
+        Ok(self.token.get(address).map(|s| s.balance).unwrap_or_else(|| 0u64.into()))
+    }
+
+    /// Check is an address is an operator of a specific owner address.
```
abizjak, line 245:
```suggestion
    /// Check if an address is an operator of a specific owner address.
```

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+    /// Results in an error if the token id does not exist in the state.
+    fn is_operator(&self, address: &Address, owner: &Address) -> bool {
+        self.token
+            .get(owner)
+            .map(|address_state| address_state.operators.contains(address))
+            .unwrap_or(false)
```
abizjak, line 251:
```suggestion
            .map_or(false, |address_state| address_state.operators.contains(address))
```

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+
+        Ok(())
+    }
+}
+
+// Contract functions
+
+/// Initialize the state contract instance with no initial tokens.
+/// Logs a `Mint` event for the single token id with no amounts.
+/// TODO: move event logs to proxy
```
abizjak, line 353: I'm not sure about the status of the todos in this PR.

Is this something you want to handle now, or is it deliberately set for later?

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+)]
+fn contract_state_initialize<S: HasStateApi>(
+    ctx: &impl HasReceiveContext,
+    host: &mut impl HasHost<State<S>, StateApiType = S>,
+) -> ContractResult<()> {
+    ensure!(!host.state().initialized, concordium_cis2::Cis2Error::Custom(AlreadyInitialized));
+    // Set proxy and implementation addresses.
+    let params: InitializeStateParams = ctx.parameter_cursor().get()?;
+    host.state_mut().proxy_address = params.proxy_address;
+    host.state_mut().implementation_address = params.implementation_address;
```
abizjak, line 399: Does it ever make sense to initilize this with `None`?

If not, I recommend removing the boolean flag alltogether and using `proxy_address.is_some()` as "is initialized" status.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+fn only_implementation(
+    implementation: Option<ContractAddress>,
+    sender: Address,
+) -> ContractResult<()> {
+    match implementation {
+        Some(implementation_address) => {
+            ensure_eq!(
+                sender,
+                Address::Contract(implementation_address),
+                concordium_cis2::Cis2Error::Custom(OnlyImplementation)
```
abizjak, line 416: I think there is a `matches_contract` function to simplify this, so

```
ensure!(sender.matches_contract(implementation_address, concordium_cis2::Cis2Error::Custom(OnlyImplementation))
```

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+    host: &mut impl HasHost<State<S>, StateApiType = S>,
+) -> ContractResult<()> {
+    // Only implementation can set state.
+    only_implementation(host.state().implementation_address, ctx.sender())?;
+
+    // Set unpause_time.
+    let unpause_time = ctx.parameter_cursor().get()?;
+    host.state_mut().unpause_time = unpause_time;
+    Ok(())
+}
```
abizjak, line 437: I guess here we don't want to have any opinions about a valid unpause time, e.g., that it be in the future?

abizjak: I'm not objecting to that, but it probably makes sense to add a bit of documentation for rationale.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+        metadata_url: MetadataUrl {
+            url:  String::from(TOKEN_METADATA_URL),
+            hash: None,
+        },
+    }))?;
+
+    Ok(state)
+}
+
+/// Initialzes the implementation of the w_ccd contract with the proxy and the
```
abizjak, line 479:
```suggestion
/// Initializes the implementation of the w_ccd contract with the proxy and the
```

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+)]
+fn contract_initialize<S: HasStateApi>(
+    ctx: &impl HasReceiveContext,
+    host: &mut impl HasHost<StateImplementation<S>, StateApiType = S>,
+) -> ContractResult<()> {
+    ensure!(!host.state().initialized, concordium_cis2::Cis2Error::Custom(AlreadyInitialized));
+    // Set proxy and storage addresses.
+    let params: InitializeImplementationParams = ctx.parameter_cursor().get()?;
+    host.state_mut().proxy_address = params.proxy_address;
+    host.state_mut().state_address = params.state_address;
```
abizjak, line 495: When does it make sense to initialize this with `None`?

I would have expected that once you are initializing you mean to actually do it, so the right hand side should always be a valid address.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+    // Set the w_ccd_implementation contract as initialized.
+    host.state_mut().initialized = true;
```
abizjak, line 497: Same as for the other contract. You could just use `proxy_address.is_some()` to mean "initialized".

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+
+            // Update the state.
+            state.mint(&TOKEN_ID_WCCD, amount.micro_ccd.into(), &receive_address, state_builder)?;
+
+            // Log the newly minted tokens.
+            logger.log(&Cis2Event::Mint(MintEvent {
+                token_id: TOKEN_ID_WCCD,
+                amount:   ContractTokenAmount::from(amount.micro_ccd),
+                owner:    sender,
+            }))?;
```
abizjak, line 556: I think I'm missing something, but why do we first log that the owner is the sender, and then log the transfer.

Why don't we just log that upon minting they are owned by the `receive_address`.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+fn contract_unwrap<S: HasStateApi>(
+    ctx: &impl HasReceiveContext,
+    host: &mut impl HasHost<StateImplementation<S>, StateApiType = S>,
+    logger: &mut impl HasLogger,
+) -> ContractResult<()> {
+    let slot_time = ctx.metadata().slot_time();
+
+    match host.state().state_address {
+        None => bail!(concordium_cis2::Cis2Error::Custom(OnlyState)),
+
```
abizjak, line 608: As above, this might be easier to read  (i.e., less nesting) if you used

```rust
let state_address = host.state().state_address.ok_or(concordium_cis2::Cis2Error::Custom(OnlyState))?
...
```

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+        Some(state_address) => {
+            let unpause_time = host
+                .invoke_contract_raw(
+                    &state_address,
+                    Parameter(&[]),
+                    EntrypointName::new_unchecked("get_unpause_time"),
+                    Amount::zero(),
+                )?
+                .1;
```
abizjak, line 530: You should probably use the `read_only` variant of the invoke_contract functions for queries such as this.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+
+            let unpause_time = if let Some(mut unpause_time) = unpause_time {
+                unpause_time.get()?
+            } else {
+                return Err(concordium_cis2::Cis2Error::Custom(StateInvokeError));
```
abizjak, line 535: I recommend commenting a bit on what the logic here is (e.g., that you are expecting a response since the state contract must be a V1 contract).

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+                    &address,
+                    &parameter,
+                    function.as_entrypoint_name(),
+                    Amount::zero(),
+                )
+                .unwrap_abort();
+                Ok(())
+            } else {
+                Ok(())
+            }
```
abizjak, line 586:
```suggestion
            }
            Ok(())
```

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+            let sender = ctx.sender();
+            let state = host.state_mut();
+            ensure!(
+                sender == params.owner || state.is_operator(&sender, &params.owner),
+                ContractError::Unauthorized
+            );
+
+            // Update the state.
+            state.burn(&TOKEN_ID_WCCD, params.amount, &params.owner)?;
+
```
abizjak, line 637: Presumably this burning will be part of the state contract, right?

When that is the case, would it not make more sense that the state contract itself checks whether it is paused or not?

You save a call and some complexity.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+                    EntrypointName::new_unchecked("get_unpause_time"),
+                    Amount::zero(),
+                )?
+                .1;
+
+            let unpause_time = if let Some(mut unpause_time) = unpause_time {
+                unpause_time.get()?
+            } else {
+                return Err(concordium_cis2::Cis2Error::Custom(StateInvokeError));
+            };
```
abizjak, line 714: This repeats a lot, so it would probably make sense to have a function for it if it stays like this.

examples/cis2-wccd-upgrade/src/lib.rs
abizjak, line 865: So the unpause_time is actually a relative time, but it is stored as an absolute time in the state.

Can you please describe the rationale for this choice? It also needs to be documented well.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+fn contract_un_pause<S: HasStateApi>(
+    ctx: &impl HasReceiveContext,
+    host: &mut impl HasHost<StateImplementation<S>, StateApiType = S>,
+) -> ContractResult<()> {
+    // Check that only the current admin can un_pause.
+    ensure_eq!(ctx.sender(), host.state().admin, ContractError::Unauthorized);
+
+    let slot_time = ctx.metadata().slot_time();
+    // Update unpause_time to current slot time.
+    host.state_mut().unpause_time = slot_time;
```
abizjak, line 886: Why not 0?

abizjak (commented):
I think overall it's looking good, but I am a bit confused by the state of it.

Are the TODOs meant to be resolved now?

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+    /// Address of the w_ccd implementation contract.
+    implementation_address: Option<ContractAddress>,
+    /// FALSE if proxy and implementation addresses have not been set yet.
+    /// TRUE if proxy and implementation addresses have been set.
+    initialized:            bool,
+    /// The state of the one token.
+    token:                  StateMap<Address, AddressState<S>, S>,
+    /// Timestamp when contract is unpaused.
+    unpause_time:           Timestamp,
+}
```
Bargsteen, line 56: I think something like this would make more sense 😊 
```
enum State<S> {
    Uninitialized,
    Initialized {
      proxy_address: ContractAddress,
      implementation_address: ContractAddress,
      token: StateMap<Address, AddressState<S>, S>,
      unpause_time: Timestamp,
     }
}

Comment (Bargsteen) [reply examples/cis2-wccd-upgrade/src/lib.rs:56]: The same idea can be applied to `StateImplementation` the proxy state.

File: examples/cis2-wccd-upgrade/src/lib.rs [hunk 25]
```diff
@@ -0,0 +1,2030 @@
...
+    proxy_address:          Option<ContractAddress>,
+    /// Address of the w_ccd implementation contract.
+    implementation_address: Option<ContractAddress>,
+    /// FALSE if proxy and implementation addresses have not been set yet.
+    /// TRUE if proxy and implementation addresses have been set.
+    initialized:            bool,
+    /// The state of the one token.
+    token:                  StateMap<Address, AddressState<S>, S>,
+    /// Timestamp when contract is unpaused.
+    unpause_time:           Timestamp,
```
Bargsteen, line 55: I see that you suggest requiring multiple signatures for changing the implementation and only a single signature for pausing/unpausing, but I don't see what use-case the `unpause_time` enables.
A single person can pause the contract, but more team members are needed to do the update and then a single person can unpause. That is fine. But what happens if they don't manage to fix and update the implementation contract before the timer runs out? Then the buggy contract would just resume being active. 🤔 
Could you explain your rationale? 😊

Bargsteen (commented):
Looks like a good start.
As a general comment:
Please be explicit about what you want a review of when requesting reviews for WIP PRs 😊 You could fx write it in the PR description.

examples/cis2-wccd-upgrade/src/lib.rs
limemloh: Well, that would also be a solution.

If we chose to log a mint event directly to the receiving address, the logs would not show the information of the sender address who actually paid for this.

examples/cis2-wccd-upgrade/src/lib.rs
DOBEN: yes, they are initialized with `None` when the `init` function is called on the smart contract. We don't know the `implementation` or the `proxy` addresses during the call of the `init` function.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,1470 @@
...
+    /// The address which are currently enabled as operators for this token and
+    /// this address.
+    operators: StateSet<Address, S>,
+}
+
+/// The contract state,
+#[derive(Serial, DeserialWithState)]
+#[concordium(state_parameter = "S")]
+struct State<S> {
+    /// The state the one token.
```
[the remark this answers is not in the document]
DOBEN: I also fixed it in the wccd contract.

examples/cis2-wccd-upgrade/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+[dependencies]
+concordium-std = {path = "../../concordium-std", default-features = false}
+concordium-cis2 = {path = "../../concordium-cis2", default-features = false}
+
+[lib]
+crate-type=["cdylib", "rlib"]
+
+[profile.release]
+opt-level = "s"
+codegen-units = 1
```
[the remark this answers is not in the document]
DOBEN: I added it to the same 4 pipelines that the `wccd contract` is part of.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+    ) -> ContractResult<()> {
+        ensure_eq!(token_id, &TOKEN_ID_WCCD, ContractError::InvalidTokenId);
+        let mut owner_state = self.token.entry(*owner).or_insert_with(|| AddressState {
+            balance:   0u64.into(),
+            operators: state_builder.new_set(),
+        });
+
+        owner_state.balance += amount;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
DOBEN: I will do. I will need to write all getter/setter functions in the state contract first before I can remove that part.

examples/cis2-wccd-upgrade/src/lib.rs
DOBEN: I also fixed it in the wccd contract.

examples/cis2-wccd-upgrade/src/lib.rs
DOBEN: I also fixed it in the wccd contract.

examples/cis2-wccd-upgrade/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
...
+    fn is_operator(&self, address: &Address, owner: &Address) -> bool {
+        self.token
+            .get(owner)
+            .map(|address_state| address_state.operators.contains(address))
+            .unwrap_or(false)
+    }
+
+    /// Update the state with a transfer.
+    /// Results in an error if the token id does not exist in the state or if
+    /// the from address have insufficient tokens to do the transfer.
```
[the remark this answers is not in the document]
DOBEN: I also fixed it in the wccd contract.

examples/cis2-wccd-upgrade/src/lib.rs
DOBEN: I also fixed it in the wccd contract.

examples/cis2-wccd-upgrade/src/lib.rs
DOBEN: My plan is to keep all the logic (including this burn function and any pausing/unpausing function) in the implementation contract. That way we can upgrade the logic.

examples/cis2-wccd-upgrade/src/lib.rs
DOBEN: Yes, that is equally possible. I don't have a strong opinion on it. I changed it to 0.

examples/cis2-wccd-upgrade/src/lib.rs
DOBEN: The `unpausing` in the future is meant as a `staging environment`. It makes sense for more complex protocols where the admin address sets several values in a smart contract. We can switch to a simple `toggle` that pauses/unpauses this contract for simplicity (it seems you are all more in favor of a simple toggle based on all the comments I get). Currently, I would like to focus on getting the proxy, implementation, and state contracts linked up correctly before starting to rename variable names/types.

examples/cis2-wccd-upgrade/src/lib.rs
Bargsteen: It might make sense for `State` to have `Option<_>`, but I don't think it makes sense for the parameter to have the same. I would assume that when you invoke `contract_state_initialize` you _must_ provide both the proxy address and the implementation address.

So something like this:
```
#[derive(Serialize, SchemaType)]
struct InitializeStateParams {
    /// Address of the w_ccd proxy contract.
    proxy_address:          ContractAddress,
    /// Address of the w_ccd implementation contract.
    implementation_address: ContractAddress,
}
```

```
    let params: InitializeStateParams = ctx.parameter_cursor().get()?;
    host.state_mut().proxy_address = Some(params.proxy_address);
    host.state_mut().implementation_address = Some(params.implementation_address);
```

Bargsteen: And that would also be the case for `InitializeImplementationParams` 😊

abizjak (commented):
I think this looks overall quite good, although I have not managed to get through all the details of state management between the 3 contracts. I have the following high-level comments

- It would be useful to have a more detailed overview of the 3 contract interactions, who holds what, and what the flow of data and tokens is.
- There was some discussion about the need to have the "unpause" time as opposed to just a flag of "locked/unlocked". It seemed to me like agreement was not fully reached. Could you provide rationale in comments in the code for the choice of the timestamp solution?

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+type ContractTokenId = TokenIdUnit;
+
+/// Contract token amount type.
+/// Since this contract is wrapping the CCD and the CCD can be represented as a
+/// u64, we can specialize the token amount to u64 reducing module size and cost
+/// of arithmetics.
+type ContractTokenAmount = TokenAmountU64;
+
+/// Needed for the custom serial instance, which doesn't include the `Option`
+/// tag and the length of the vector.
```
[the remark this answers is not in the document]
DOBEN: I took this part of the code from the “proxy” example that I found in the repo. The return value from the invoke in the fallback function (“host.invoke_contract_raw().1”) is an "Option" and it is converted into the “RawReturnValue”. If I understand correctly then we need this conversion from “return_value” into “RawReturnValue” because the fallback function needs to specify what type it returns. “ReceiveResult<RawReturnValue>” is a newly defined data type that has an unlimited vector length so we can pass anything from “return_value” into “RawReturnValue”.
I am not exactly sure what the `Option` tag refers to in the comment so.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+    /// Address of the w_ccd state contract.
+    state_address:  Option<ContractAddress>,
+    /// Fallback function on the proxy will set this value
+    /// before invoking a function on the implementation.
+    current_sender: Address,
+}
+
+/// The `proxy` contract state.
+#[derive(Serial, Deserial, SchemaType)]
+struct StateProxy {
```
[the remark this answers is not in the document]
DOBEN: I tried it out but removing it will cause an error during the schema generation. `StateImplementation` and `StateProxy` are used as return values for the `view` function.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+#[derive(Serialize, SchemaType)]
+struct UnwrapParams {
+    /// The amount of tokens to unwrap.
+    amount:   ContractTokenAmount,
+    /// The owner of the tokens.
+    owner:    Address,
+    /// The address to receive these unwrapped CCD.
+    receiver: Receiver,
+    /// Some additional bytes to include in the transfer.
+    data:     AdditionalData,
```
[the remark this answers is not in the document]
DOBEN: Updated the two comments about `additionaData` to the following:

- If the `Receiver` is a contract the unwrapped CCD together with these additional data bytes are sent to the function entrypoint specified in the `Receiver`.
- Some additional bytes that are used in the `OnReceivingCis2` hook. Only if the `Receiver` is a contract the receive hook function is executed.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+    /// Address of the w_ccd implementation contract.
+    implementation_address: Option<ContractAddress>,
+}
+
+/// The return type for the state contract function `isOperator`.
+#[derive(Serialize, SchemaType)]
+struct ReturnIsOperator {
+    /// Boolean that specifies if an address is an operator or not.
+    is_operator: bool,
+}
```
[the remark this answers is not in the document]
DOBEN: I removed it and just use now `return_value = "bool"` in the `isOperator` function.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+            proxy_address:          None,
+            implementation_address: None,
+            token:                  state_builder.new_map(),
+            unpause_time:           Timestamp::from_timestamp_millis(0),
+        }
+    }
+}
+
+impl StateImplementation {
+    /// Creates a new state with no one owning any tokens by default.
```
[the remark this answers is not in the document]
DOBEN: updated comment

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+}
+
+/// The parameter type for the state contract function `initialize`.
+#[derive(Serialize, SchemaType)]
+struct InitializeStateParams {
+    /// Address of the w_ccd proxy contract.
+    proxy_address:          Option<ContractAddress>,
+    /// Address of the w_ccd implementation contract.
+    implementation_address: Option<ContractAddress>,
+}
```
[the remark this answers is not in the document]
DOBEN: yes, I think it is better to not use options here. I removed it and just use `ContractAddress`.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+
+/// The parameter type for the implementation contract function `initialize`.
+#[derive(Serialize, SchemaType)]
+struct InitializeImplementationParams {
+    /// Address of the w_ccd proxy contract.
+    proxy_address: Option<ContractAddress>,
+    /// Address of the w_ccd state contract.
+    state_address: Option<ContractAddress>,
+}
```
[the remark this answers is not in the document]
DOBEN: yes, I think it is better to not use options here. I removed it and just use ContractAddress.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+    if let Receiver::Contract(address, function) = params.to {
+        let parameter = OnReceivingCis2Params {
+            token_id: TOKEN_ID_WCCD,
+            amount:   ContractTokenAmount::from(amount.micro_ccd),
+            from:     sender,
+            data:     params.data,
+        };
+        host.invoke_contract(&address, &parameter, function.as_entrypoint_name(), Amount::zero())
+            .unwrap_abort();
+    }
```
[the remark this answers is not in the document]
Bargsteen: That is also what I recall.

Bargsteen (commented):
This looks quite good to me. I agree with Ales about the overview and unpause time. And I still think we should consider using this pattern for all the states: `enum State_X { Uninitialized, Initialized { proxy_address: ContractAddress, ... }`.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+    // Get the instantiater of this contract instance.
+    let invoker = Address::Account(ctx.init_origin());
+    // Construct the initial contract state.
+    let state = StateImplementation::new(invoker, invoker);
+    // Log event for the newly minted token.
+    logger.log(&Cis2Event::Mint(MintEvent {
+        token_id: TOKEN_ID_WCCD,
+        amount:   ContractTokenAmount::from(0u64),
+        owner:    invoker,
+    }))?;
```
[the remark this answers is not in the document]
limemloh: It is not required by the standard to do this, but the purpose here is to announce the existence of the token and the token id.
The total supply at this point is 0, but the token does exist.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+    let entrypoint = ctx.named_entrypoint();
+    let implementation = host.state().implementation_address;
+    let mut parameter_buffer = vec![0; ctx.parameter_cursor().size() as usize];
+    ctx.parameter_cursor().read_exact(&mut parameter_buffer)?;
+
+    // The `proxy` will set the `current_sender` on the `implementation` before
+    // forwarding the invoke unaltered to the implementation contract.
+    let parameter_bytes = to_bytes(&SetCurrentSenderParams {
+        sender: ctx.sender(),
+    });
```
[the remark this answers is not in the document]
DOBEN: My initial proposal was to append the `sender` to the `parameter_buffer`. Looking at the code now I was not exactly sure how to append this buffer in Rust especially because we need to append it by an `Address` which has a different format/size when it is a `Contract` address compared to when it is an `Account`. Maybe that is not a problem at all to append it by an `Address` but I haven’t tried it yet.

The current approach has an advantage in terms of schema generation. I am able to use the same schema type from the implementation and add it to the fallback function on the proxy. E.g. I use “WrapParams” on the fallback function on the proxy when I want to execute the “wrap” function in the protocol.

Appending an address to the buffer seems cheaper while the other approach might make schema integration easier. I assume, we need to discuss this a bit and then choose the better approach.

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
Bargsteen: The example proxy contract is built so it also works with v0 contracts. Since v0 contracts don't have a return value, the option is needed.
I expect that we will only use v1 contracts in this case, so we can remove the option.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+    /// Address of the w_ccd implementation contract.
+    implementation_address: Option<ContractAddress>,
+    /// The state of the one token.
+    token:                  StateMap<Address, AddressState<S>, S>,
+    /// Timestamp when contract is unpaused.
+    unpause_time:           Timestamp,
+}
+
+/// The `implementation` contract state.
+#[derive(Serial, Deserial, SchemaType)]
```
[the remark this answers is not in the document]
Bargsteen: Then I guess it depends on the purpose of those `view` functions. When will they be used and by who? Perhaps it makes more sense to create new types used for purpose of viewing.

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
DOBEN: Would it be better to use the following event here:

Cis2TokenDeployed{
token_address: ContractAddress
token_id: TOKEN_ID_WCCD,
owner:    invoker,
}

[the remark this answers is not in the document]
limemloh: You mean introduce an event for logging the existence of a token, and another for minting of this token?
It would make it a bit more explicit, but for tokens which are not starting out with a total supply of 0, it would mean having to log more events and I don't think it would provide much value.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+fn contract_implementation_recieve_ccd<S: HasStateApi>(
+    _ctx: &impl HasReceiveContext,
+    _host: &impl HasHost<StateImplementation, StateApiType = S>,
+    _amount: Amount,
+) -> ContractResult<()> {
+    Ok(())
+}
+
+/// Function to receive CCD on the proxy.
+#[receive(contract = "CIS2-wCCD-Proxy", name = "receiveCCD", payable)]
```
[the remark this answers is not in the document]
DOBEN: Yes, all CCDs are held by the proxy contract. Nonetheless, during the `unwrap` function, we need to get access to some CCD amount from the proxy. Currently, we request in the `unwrap` function that the proxy should send some CCD amount to the implementation which then goes through more complex logic to deal with sending the CCD to the receiver address (which can be an account or a smart contract address) as well as logging required events.

The proxy logic is kept as simple as possible because we cannot upgrade the logic. That is why it only sends CCD to the implementation and then let the implementation code more complex logic.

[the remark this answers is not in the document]
DOBEN: added a comment

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
DOBEN: Yes, if minting is happening during the deployment of a token then we need to emit a `MintEvent` in addition to the `Cis2TokenDeployedEvent`.

The advantage of adding a separate `Cis2TokenDeployedEven` is that off-chain indexers can insert the `token_address` in their off-chain database easily without having to skim to `MintEvents` to find a `Mint` event that is associated with a new token and then have to query to token_address or look it up from the tx.receipt data.

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
DOBEN: The view functions are there for manual checking of the state of the contracts. E.g. you might want to check if after the setup all contracts are correctly linked up. The function will be used mainly by us but also by some interested users who want to explore the protocol. 

Not sure if we want to add even more types. There are already a lot of different types in the contract and since the `StateProxy` and `StateImplementation` types are working fine as a return parameter for the `view` function, I am not so convinced that we need a new type here.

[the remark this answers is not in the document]
Bargsteen: Ok, then I think it's fine to keep it as is.

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
DOBEN: How can you specify that this invoke is to a `version1` contract `host.invoke_contract_raw()?.1`? Currently, the `return_value` is an option and that is why we need to deal with the case that `None` is returned by this invoke. The `RawReturnValue` deals with both cases that `None` is returned or that some `return_value` is returned.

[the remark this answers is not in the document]
Bargsteen: You can't specify it directly, but v1 contracts will _always_ return a value (although it might be `()`), so if you unwrap the return value, we can get rid of the option. (I've sent more details directly to you).

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+/// This function is meant for recovering CCD tokens in the future. In case some
+/// CCD tokens are accidentally sent to the `state` contract we can upgrade the
+/// `implementation` contract to access the CCD via this function if necessary.
+#[receive(
+    contract = "CIS2-wCCD-State",
+    parameter = "TransferCCDParams",
+    name = "transferCCD",
+    mutable
+)]
+fn contract_state_transfer_ccd<S: HasStateApi>(
```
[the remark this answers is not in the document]
DOBEN: removed this function after confirmation that it is not possible to force CCD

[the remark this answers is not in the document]
Bargsteen: As for the edge cases:
1. We don't have a self-destruct host function.
2. I assume that this means that the contract receives the rewards for baking a block. Which we don't allow.
3. You can easily precalculate its address, but you cannot transfer CCD to it before it has been initialised.

So CCD can only get onto a smart contract if it has at least one `payable` entrypoint.

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
DOBEN: I added this comment above the the `mint` event:

// Log a `Mint` event for the single token id with 0 amount. Minting an amount
// of 0 is used to signal that a new token was deployed. Invoking the mint event
// with 0 amount in the `wrap` function is impossible.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3721 @@
...
+    )?;
+
+    let owner = receive_address;
+
+    let set_balance_params = SetBalanceParams {
+        owner,
+        amount: amount.micro_ccd.into(),
+        update: Update::Add,
+    };
+
```
limemloh, line 1431:
```suggestion
    let set_balance_params = SetBalanceParams {
        owner: receive_address,
        amount: amount.micro_ccd.into(),
        update: Update::Add,
    };

```

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3721 @@
...
+    amount: Amount,
+) -> ContractResult<()> {
+    let (proxy_address, state_address) = if let ProtocolAddressesImplementation::Initialized {
+        proxy_address,
+        state_address,
+    } = host.state().protocol_addresses
+    {
+        (proxy_address, state_address)
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
```
limemloh, line 1396: use bail

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3721 @@
...
+    };
```
limemloh, line 1397: Consider adding a `getProtocolAddresses` method on `StateImplementation`, which returns this pair and bails if not initialized

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3721 @@
...
+    host: &mut impl HasHost<StateImplementation, StateApiType = S>,
+) -> ContractResult<()> {
+    let state_address = if let ProtocolAddressesImplementation::Initialized {
+        proxy_address: _,
+        state_address,
+    } = host.state().protocol_addresses
+    {
+        state_address
+    } else {
+        panic!("{:?}", concordium_cis2::Cis2Error::Custom(CustomContractError::UnInitialized));
```
limemloh: use bail

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3721 @@
...
+
+/// Function to receive CCD on the proxy. All CCDs are held by the
+/// proxy contract.
+#[receive(contract = "CIS2-wCCD-Proxy", name = "receiveCCD", payable)]
+fn contract_proxy_recieve_ccd<S: HasStateApi>(
+    _ctx: &impl HasReceiveContext,
+    _host: &impl HasHost<StateProxy, StateApiType = S>,
+    _amount: Amount,
+) -> ContractResult<()> {
+    Ok(())
```
limemloh, line 1233: It would probably be a good idea to fail when not called by the implementation contract.
Since we want a 1:1 with wCCD and the amount of CCD in this contract.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3721 @@
...
+
+    let input: WrapParamsWithSender = ctx.parameter_cursor().get()?;
+
+    // Get the sender who invoked this contract function.
+    let sender = input.sender;
+
+    let receive_address = input.params.to.address();
+
+    // Proxy holds CCD funds. CCD is sent to proxy.
+    host.invoke_contract_raw(
```
limemloh, line 1417: Instead of moving the CCD back and forth between proxy and implementation, it might be enough to just pass the information back removing the need to call back the proxy with the CCD.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3721 @@
...
+                    .is_operator(&sender, &from, host)
+                    .map_or(false, |is_operator| is_operator),
+            ContractError::Unauthorized
+        );
+        let to_address = to.address();
+        // Update the contract state
+
+        if amount == 0u64.into() {
+            return Ok(());
+        }
```
limemloh, line 1668: 1. This is returning the entire transfer function call, meaning the rest of the loop is never executed after a zero amount transfer.

2. According to the CIS-2:  _A transfer of a token amount zero MUST be executed as a normal transfer_.
So using `continue` here is right either. It is fine to skip calling the state contract in this case, but it will still need to log a transfer event and call the hook.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3721 @@
...
+/// It rejects if:
+/// - It fails to parse the parameter.
+/// - Any of the transfers fail to be executed, which could be if:
+///     - The `token_id` does not exist.
+///     - The sender is not the owner of the token, or an operator for this
+///       specific `token_id` and `from` address.
+///     - The token is not owned by the `from`.
+/// - Fails to log event.
+/// - Any of the receive hook function calls rejects.
+#[receive(contract = "CIS2-wCCD", name = "transfer", mutable)]
```
limemloh, line 1618: I don't think this needs to be mutable

limemloh (changes requested):

examples/cis2-wccd-upgradable/src/lib.rs
DOBEN: The CCD amount is passed by the fallback function on the proxy to the implementation. The implementation has the logic to check how much WCCD should be minted (based on the amount of CCD it receives in the `wrap` function call). This logic has to be executed before the WCCD can be sent to the proxy (sent to be stored on the proxy). The CCD should be stored on the proxy in case we upgrade the implementation.

Can you please explain how you can pass the information back or improve the pattern based on your above comment?

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -541,7 +569,7 @@ fn contract_proxy_log_mint_event<S: HasStateApi>(
     logger: &mut impl HasLogger,
```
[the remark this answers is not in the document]
DOBEN: I get an error when removing `mutable`

examples/cis2-wccd-upgradable/src/lib.rs
DOBEN: Added restrictions that the two `receiveCCD` functions can only be called by `implementation` or `proxy` respectively.

examples/cis2-wccd-upgradable/src/lib.rs
DOBEN: Transfer events are now logged even if the amount transferred is 0.

examples/cis2-wccd-upgradable/src/lib.rs
DOBEN: I get an error removing the `mutable`.

examples/cis2-wccd-upgradable/src/lib.rs
limemloh: I was thinking adding the amount of tokens (just like the sender address), such that the information of how much was transferred is forwarded to the implementation, but not the actual CCDs.
But I am not sure whether it will simplify anything.

DOBEN: Since the fallback function will be used by all state-mutative functions, we would also append the "amount of tokens" to the "transfer", "unwrap", and "updateOperator" functions which do not need that information. I am not sure what is the best (most efficient/ simplest) way of handling it.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3512 @@
...
+    /// New admin address.
+    new_admin: Address,
+}
+
+/// NewImplementationEvent.
+#[derive(Serial)]
+struct NewImplementationEvent {
+    /// New implementation address.
+    new_implementation: ContractAddress,
+}
```
limemloh, line 269: You probably want to merge these types into an enum, since this would put a tag byte in front of the serialization.
Now, without the tag, it is difficult to distinguish the events provided just the bytes.

examples/cis2-wccd-upgradable/src/lib.rs
limemloh: ahh right, the invoking here needs the mutability, nevermind then

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3537 @@
...
+        .map_err(|r| {
+            if let CallContractError::LogicReject {
+                reason,
+                ..
+            } = r
+            {
+                Reject::new(reason).unwrap_abort()
+            } else {
+                r.into()
+            }
```
limemloh, line 839: For now this is fine, but I don't think it will be necessary after https://github.com/Concordium/concordium-rust-smart-contracts/pull/145 gets merged

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3537 @@
...
+        Receiver::Account(address) => host.invoke_transfer(&address, unwrapped_amount)?,
+        Receiver::Contract(address, function) => {
+            host.invoke_contract(
+                &address,
+                &input.params.data,
+                function.as_entrypoint_name(),
+                unwrapped_amount,
+            )?;
+        }
+    }
```
limemloh, line 1470: Just a thought:
If the `transferCCD` function took a receiver in the parameter, it could send the CCD directly instead of the implementation having to do it.
The main benefit is that users will receive the unwrapped CCD from the Proxy contracts instead of the implementation contract.

DOBEN: The proxy transfers the CCD now directly to the receiver address.

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
abizjak: Ok. In the ideal world that would be a separate type, but in this case that does not seem the most sensible option. So we keep it :)

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3181 @@
...
+    Ok(host.self_balance())
+}
+
+/// Function to view state of contract.
+#[receive(contract = "CIS2-wCCD-State", name = "view", return_value = "ReturnBasicState")]
+fn contract_state_view<S: HasStateApi>(
+    _ctx: &impl HasReceiveContext,
+    host: &impl HasHost<State<S>, StateApiType = S>,
+) -> ContractResult<ReturnBasicState> {
+    let state = ReturnBasicState {
```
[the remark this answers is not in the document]
Bargsteen: We needed an `uninitialized` error anyhow. So this is good now ;)

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+
+    let (_proxy_address, state_address) = get_protocol_addresses_from_implementation(host)?;
+
+    // Build the response.
+    let mut response = Vec::with_capacity(params.queries.len());
+    for std_id in params.queries {
+        if SUPPORTS_STANDARDS.contains(&std_id.as_standard_identifier()) {
+            response.push(SupportResult::Support);
+        } else {
+            response.push(host.state().have_implementors(&state_address, &std_id, host).unwrap());
```
limemloh, line 2224: You should fail with a proper error here

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+                from:     sender,
+                data:     input.params.data,
+            };
+            host.invoke_contract(
+                &address,
+                &parameter,
+                function.as_entrypoint_name(),
+                Amount::zero(),
+            )
+            .unwrap_abort();
```
limemloh, line 1567: I would prefer a proper error here

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+
+    // Get the sender who invoked this contract function.
+    let sender = input.sender;
+
+    ensure!(
+        sender == input.params.owner
+            || host
+                .state()
+                .is_operator(&state_address, &sender, &input.params.owner, host)
+                .map_or(false, |is_operator| is_operator),
```
limemloh, line 1601: If the state contract fails to respond, this will result in `Unauthorized`. It would be better to return the error.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+        data,
+    } in transfers
+    {
+        // Authenticate the sender for this transfer
+        ensure!(
+            from == sender
+                || host
+                    .state()
+                    .is_operator(&state_address, &sender, &from, host)
+                    .map_or(false, |is_operator| is_operator),
```
limemloh, line 1713: Same as other comment

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+    let (_proxy_address, state_address) = get_protocol_addresses_from_implementation(host)?;
+
+    // Build the response.
+    let mut response = Vec::with_capacity(params.queries.len());
+    for query in params.queries {
+        // Query the state if the `address` being an `operator` of `owner`.
+        let is_operator = host
+            .state()
+            .is_operator(&state_address, &query.address, &query.owner, host)
+            .map_or(false, |is_operator| is_operator);
```
limemloh, line 2190: Same as other comment related to is_operator

limemloh (changes requested):
I have a bunch of code style comments, but also a few "real" comments.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+    // with 0 amount in the `wrap` function is impossible.
+    logger.log(&Cis2Event::Mint(MintEvent {
+        token_id: TOKEN_ID_WCCD,
+        amount:   ContractTokenAmount::from(0u64),
+        owner:    ctx.sender(),
+    }))?;
+
+    let url = host.invoke_contract_read_only(
+        &state_address,
+        &Parameter(&[]),
```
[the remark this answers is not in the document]
DOBEN: I get an error when removing the parameter because the type cannot be inferred anymore. I kept the old version.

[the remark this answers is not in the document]
Bargsteen: Yea, the Parameter is now a wrapper of a byte array. So `Parameter(&[])` is correct. I believe the change was needed for something Ales was doing.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+
+    let receive_address = input.params.to.address();
+
+    // Proxy holds CCD funds. CCD is sent to proxy.
+    host.invoke_contract_raw(
+        &proxy_address,
+        Parameter(&[]),
+        EntrypointName::new_unchecked("receiveCCD"),
+        amount,
+    )?;
```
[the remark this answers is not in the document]
Bargsteen: I think the `_raw` version might be slightly faster, as it won't try to serialize the slice.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+        proxy_address:          ctx.self_address(),
+        implementation_address: host.state().implementation_address,
+    });
+
+    host.invoke_contract_raw(
+        &state_address,
+        Parameter(&parameter_bytes),
+        EntrypointName::new_unchecked("initialize"),
+        Amount::zero(),
+    )?;
```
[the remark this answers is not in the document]
Bargsteen: Yep, `invoke_contract` will call the `to_bytes` for you 😊

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+    name = "viewBalance",
+    return_value = "Amount",
+    error = "ContractError"
+)]
+fn contract_proxy_view_balance<S: HasStateApi>(
+    _ctx: &impl HasReceiveContext,
+    host: &impl HasHost<StateProxy, StateApiType = S>,
+) -> ContractResult<Amount> {
+    Ok(host.self_balance())
+}
```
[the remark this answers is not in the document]
DOBEN: These functions are not used internally and if you want to know the balance of a smart contract off-chain, you can use ``concordium-client``. 

But I see an use case that smart contracts can query the balance of any other smart contract within the Rust code (e.g. `(smartContractAddress).balance` this is currently not natively supported but on the smart contract wish list for future development). 

Since it is not available at the moment, I added a view function in case a third-party smart contract needs this information.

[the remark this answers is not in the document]
limemloh: I don't see why other smart contracts would need this. Especially not for the internal contracts such as implementation and state. 
I think we should be prioritizing having a smaller module instead.

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
limemloh: Right, Serial for a slice would put the length of 0, what I should have suggested was to use unit `&()` instead.

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
limemloh: yes, that is probably true for this case, since `to_bytes` used in `invoke_contract` is allocating a vec.

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
DOBEN: Changed it back to  `Parameter(&[])`

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+
+    // Build the response.
+    let mut response = Vec::with_capacity(params.queries.len());
+    for token_id in params.queries {
+        // Check the token exists.
+        ensure_eq!(token_id, TOKEN_ID_WCCD, ContractError::InvalidTokenId);
+
+        let url = host.invoke_contract_read_only(
+            &state_address,
+            &Parameter(&[]),
```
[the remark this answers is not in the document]
DOBEN: kept it as comment above

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3785 @@
...
+}
+
+/// Helper function to ensure contract is not paused.
+fn when_not_paused<S>(
+    state_address: &ContractAddress,
+    host: &mut impl HasHost<StateImplementation, StateApiType = S>,
+) -> ContractResult<()> {
+    let paused = host.invoke_contract_read_only(
+        state_address,
+        &Parameter(&[]),
```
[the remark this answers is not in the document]
DOBEN: kept it as discussed in a comment above

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
DOBEN: I don't know any concrete example of why a third-party smart contract would want to query the balance of the wCCD smart contract.

Since we could do a protocol update one day and add this `(smartContractAddress).balance` feature, I can remove the above addresses, and we prioritize the module size now.

examples/cis2-wccd-upgradable/src/lib.rs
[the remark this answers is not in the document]
DOBEN: kept it unchanged

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3674 @@
...
+                &address,
+                &OnReceivingCis2Params {
+                    token_id: TOKEN_ID_WCCD,
+                    amount:   ContractTokenAmount::from(amount.micro_ccd),
+                    from:     sender,
+                    data:     input.params.data,
+                },
+                function.as_entrypoint_name(),
+                Amount::zero(),
+            )?;
```
limemloh, line 1508: I think we should move the invoking of receive hook to the proxy, since this acts as the token contract to the outside.
The smart contract receiving this, would see the implementation contract as the sender and might assume the sender is the token contract.

examples/cis2-wccd-upgradable/src/lib.rs
```diff
@@ -0,0 +1,3674 @@
...
+                    &SetIsOperatorParams {
+                        owner:    sender,
+                        operator: param.operator,
+                        update:   Update::Remove,
+                    },
+                    EntrypointName::new_unchecked("setIsOperator"),
+                    Amount::zero(),
+                )?;
+            }
+        };
```
limemloh, line 1759:
```suggestion
        let update = match param.update {
            OperatorUpdate::Add => Update::Add,
            OperatorUpdate::Remove => Update::Remove,
        };
        host.invoke_contract(
            &state_address,
            &SetIsOperatorParams {
                owner: sender,
                operator: param.operator,
                update,
            },
            EntrypointName::new_unchecked("setIsOperator"),
            Amount::zero(),
        )?;

```

examples/cis2-wccd-upgradable/src/lib.rs
DOBEN: moved to proxy

Bargsteen:
I guess we should close this and thank Doris for all of her hard work proving why we really needed the native upgradability :blush:

abizjak:
Closing this, but keeping the branch alive in case we need to come back to some details here.

PR #136, opened 2022-07-11
fixed swapped around variables in operatorOf function

### LINKED ISSUE #135: `address` and `owner` variables are swapped around in the `OperatorOf` functions of the cis2 tokens.
**Bug Description**

When querying the operator then the `owner` and `address` variables are swapped around. This means that the ‘OperatorOf’ function returns a wrong value. 

fn is_operator(&self, address: &Address, owner: &Address) -> bool
https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/examples/cis2-wccd/src/lib.rs#L161

let is_operator = host.state().is_operator(&query.owner, &query.address);
https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/examples/cis2-wccd/src/lib.rs#L580

DOBEN:
## Purpose
When querying the operator then the `owner` and `address` variables were swapped around. This means that the ‘OperatorOf’ function returned a wrong value. This is fixed now.

closes #135

## Changes

- Fixed cis2-nft, cis2-multi, and cis2-wccd smart contract

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/cis2-multi/src/lib.rs
+++ b/examples/cis2-multi/src/lib.rs
@@ -546,7 +546,7 @@ fn contract_operator_of<S: HasStateApi>(
     let mut response = Vec::with_capacity(params.queries.len());
     for query in params.queries {
         // Query the state for address being an operator of owner.
-        let is_operator = host.state().is_operator(&query.owner, &query.address);
+        let is_operator = host.state().is_operator(&query.address, &query.owner);
         response.push(is_operator);
     }
     let result = OperatorOfQueryResponse::from(response);
--- a/examples/cis2-nft/src/lib.rs
+++ b/examples/cis2-nft/src/lib.rs
@@ -500,7 +500,7 @@ fn contract_operator_of<S: HasStateApi>(
     let mut response = Vec::with_capacity(params.queries.len());
     for query in params.queries {
         // Query the state for address being an operator of owner.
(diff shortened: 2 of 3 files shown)


Bargsteen (changes requested):
Please consider adding a test for each contract to ensure that it works as expected (and that it will do so in the future :blush: )

abizjak (approved):
Good catch.

And I like that you added tests as well.

examples/cis2-multi/src/lib.rs
```diff
@@ -956,6 +956,26 @@ mod tests {
...
+            queries: vec![operator_of_query],
+        };
+        let parameter_bytes = to_bytes(&operator_of_query_vector);
+
+        ctx.set_parameter(&parameter_bytes);
+
+        // Checking the return value of the `contract_operator_of` function
+        let result: ContractResult<OperatorOfQueryResponse> = contract_operator_of(&ctx, &host);
+
+        claim_eq!(result.unwrap().0, [true], "Account should be an operator in the query response");
```
limemloh, line 977: Use `expect_report` from `concordium-std` instead of `unwrap` as this will work better for the tests when using `cargo concordium test`

limemloh (approved):
Looks good, have only one comment

Bargsteen (approved):
LGTM 👍

Issue #123, opened 2022-06-19
Incorrect state management in the test host


abizjak:
**Bug Description**

Test host does not handle rollbacks correctly.

**Steps to Reproduce**

See https://github.com/Anfid/concordium-state-rollback-test

The source is copied below for posterity.

```rust

#![no_std]

use concordium_std::*;

#[derive(Reject)]
enum Error {
    NotFound,
}

#[derive(Debug, Serialize, SchemaType, PartialEq, Eq)]
struct DataEntry {
    key: String,
    value: String,
}

#[derive(Debug, Serial, DeserialWithState)]
#[concordium(state_parameter = "S")]
struct State<S: HasStateApi> {
    storage: StateMap<String, String, S>,
}

#[init(contract = "StateRollbackTest")]
fn init<S: HasStateApi>(
    _: &impl HasInitContext,
    state_builder: &mut StateBuilder<S>,
) -> InitResult<State<S>> {
    Ok(State {
        storage: state_builder.new_map(),
    })
}

// Insert function is omitted for brevity

/// Update function
///
/// Update value for the given key if entry exists. If entry does not exist, return [Error::NotFound]
#[receive(
    mutable,
    contract = "StateRollbackTest",
    name = "update",
    parameter = "Entry"
)]
fn contract_update<S: HasStateApi>(
    ctx: &impl HasReceiveContext,
    host: &mut impl HasHost<State<S>, StateApiType = S>,
) -> ReceiveResult<()> {
    let state = host.state_mut();

    let params = DataEntry::deserial(&mut ctx.parameter_cursor())?;

    match state.storage.insert(params.key, params.value) {
        Some(_) => Ok(()),
        None => Err(Error::NotFound.into()),
    }
}

#[concordium_cfg_test]
mod tests {
    use super::*;
    use concordium_std::test_infrastructure::*;

    #[concordium_test]
    fn test_transactionality() {
        let ctx = TestInitContext::empty();
        let mut state_builder = TestStateBuilder::new();

        // Init the state with contract function
        let state =
            init(&ctx, &mut state_builder).expect_report("Failed during init_StateRollbackTest");

        let mut host = TestHost::new(state, state_builder);

        // Confirm that state is empty
        claim!(host.state().storage.is_empty());

        let params = to_bytes(&DataEntry {
            key: String::from("key"),
            value: String::from("value"),
        });
        let mut ctx = TestReceiveContext::default();
        ctx.set_parameter(&params);

[...]

PR #141, opened 2022-07-18
Implement roll back functionality on `TestHost`

Bargsteen:
## Purpose

Closes https://github.com/Concordium/concordium-rust-smart-contracts/issues/123 by adding the ability to roll back the state (and host) when receive functions reject in a test.

Depends on https://github.com/Concordium/concordium-contracts-common/pull/50

## Changes

- Add `TestHost::with_rollback` method.
- Add `StateClone` trait.
   - Add a blanket implementation for all cloneable types.
   - Add ability to derive it for types that use StateMap/StateSet/StateBox.
- Add tests to two contracts which use `with_rollback`.
- Add `cursor_position` method to `Seek` trait (needed to implement `StateClone` for `StateBox`).
- Add `StateClone` constraint on `HasHost` impl for `TestHost` (breaking change).
- Add `deep_clone` method to the test trie.

Unrelated minor changes:
- Rename the state identifier to lowercase to fix clippy warning (`___state`)

## Checklist

      hard-to-understand areas.


concordium-std/src/traits.rs
```diff
@@ -133,6 +133,9 @@ where
     /// `.seek(SeekFrom::Start(0))` but can be implemented more efficiently.
     fn move_to_start(&mut self);
 
+    /// Get the cursor position counted from the beginning of the stream.
+    fn get_cursor_position(&self) -> u32;
```
limemloh, line 137: Just want to mention that Seek already exposes a way to get the current position. This function is of course more direct.

limemloh (approved):
My comments are only minor things.

concordium-std/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@
 - Add `Seek` requirement for `HasParameter`.
 - Implement `Seek` for `ExternParameter`.
 - Add wrapper type `TestParameterCursor` instead of exposing `Cursor` directly, when using `TestContext`. This is changing the type returned by `parameter_cursor` for `TestContext`, but provides the same interface.
+- Add rollback functionality to the `TestHost`.
+  - Add a function `TestHost::with_rollback` for calling receive functions,
+    which rolls back the host and state if the receive function returns an error.
+  - Add a `StateClone` trait.
+    - The `TestHost` requires the `State` to implement `StateClone` (breaking change).
+    - `StateClone` is implements for all `Clone` types, and can be derived similarly to `DeserialWithState`.
```
abizjak, line 12:
```suggestion
    - `StateClone` is implemented for all `Clone` types, and can be derived similarly to `DeserialWithState`.
```

concordium-std/src/traits.rs
abizjak: I would probably prefer if we add a method `get_cursor_position` to the Seek trait (with a default implementation).

Bargsteen: I couldn't add a default implementation because it would need a `&mut self`. But I added it to the `Seek` trait nonetheless.

PR #144, opened 2022-07-21
More ergonomic policies

### LINKED ISSUE #118: Using policies is not very ergonomic
**Task description**

Currently using policies in contracts requires a bit of boilerplate that we could hide. For example to find whether a given attribute exists we have to do something like

```rust
    for mut policy in policies {
        let mut buf = [0u8; 31];
        while let Some((tag, len)) = policy.next_item(&mut buf) {
            if tag == attributes::COUNTRY_OF_RESIDENCE {
                if buf[0..len.into()] == b"DK"[..] {
                    return Ok(()); // accept the deposit.
                }
            }
        }
    }
```
which is not very ergonomic. There are a number of things that could be done.

**Sub-tasks**
- [ ] The inner while loop is an iterator. We can wrap that into an iterator over (tag, &[u8]) pairs by making a "PolicyAttributesIterator" struct that is something like
   ```
     struct PAI {
       iter: PolicyType,
       buf: [u8; 31]
   }
   ```
- [ ] Once we have that we can make use of all of "Iterator" methods, like collect, find, exists, etc.

Bargsteen:
## Purpose

Closes #118 by implementing an iterator over policy attributes.
It is a first step towards making policies more ergonomic to use.

## Changes

- Add `iter` method to `HasPolicies` which returns the new type `PolicyAttributesIter`.
- The `PolicyAttributesIter` has an iterator implementation over `(AttributeTag, AttributeValue)`.
- Add an example contract that showcases how to use policies.

## Checklist

      hard-to-understand areas.


concordium-std/src/impls.rs
```diff
@@ -1640,6 +1640,24 @@ impl HasPolicy for Policy<AttributesCursor> {
...
+
+    fn iter(&mut self) -> PolicyAttributesIter<Self> {
+        PolicyAttributesIter {
+            iter: self,
+            buf:  [0u8; 31],
+        }
+    }
+}
+
+impl<P: HasPolicy> Iterator for PolicyAttributesIter<'_, P> {
```
limemloh, line 1652: You should be able to implement [`ExactSizeIterator`](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) for `PolicyAttributesIter` and [`IntoIterator`](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) for `HasPolicy` as well

concordium-std/src/impls.rs
```diff
@@ -1640,6 +1640,24 @@ impl HasPolicy for Policy<AttributesCursor> {
         self.items.remaining_items -= 1;
         Some((AttributeTag(tag_value_len[0]), tag_value_len[1]))
     }
+
+    fn iter(&mut self) -> PolicyAttributesIter<Self> {
```
limemloh, line 1644: Calling next on the iterator will mutate the implementor of `HasPolicy`, which is not ideal.
It would be nice, if we could avoid the need for a mutable reference to self here.

Ideally, we could also remove the `next_item` in `HasPolicy`, but is a breaking change of course.

concordium-std/src/types.rs
```diff
@@ -645,6 +646,13 @@ pub struct AttributesCursor {
     pub(crate) remaining_items:  u16,
 }
 
+/// An iterator over the attributes of a policy.
+/// The iterator returns pairs of [`AttributeTag`] and [`AttributeValue`].
+pub struct PolicyAttributesIter<'a, P: HasPolicy> {
+    pub iter: &'a mut P,
+    pub buf:  [u8; 31],
```
limemloh, line 653: I don't think these fields needs to be public

abizjak: They shouldn't be. Especially the `buf`.

concordium-std/src/impls.rs
Bargsteen: I can't implement `IntoIterator` for `P: HasPolicy` due to the orphan rule. But I implemented `ExactSizeIterator` by adding an `remaining_items` function toe `HasPolicy`.

examples/transfer-policy-check/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+# concordium-std = "*"
+concordium-std = {path = "../../concordium-std"}
+
+[features]
+default = ["std"]
+std = ["concordium-std/std"]
```
limemloh, line 16: Either support `no_std`
```suggestion
[dependencies]
concordium-std = {path = "../../concordium-std", default-features = false}

[features]
default = ["std"]
std = ["concordium-std/std"]
```
or remove these features
```suggestion
[dependencies]
concordium-std = {path = "../../concordium-std"}
```

examples/transfer-policy-check/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+ * The contract is initialised with `init`, in which an account address is
+ * provided, which will be set as the state.
+ *
+ * The contract has a single receive entrypoint, `receive`,
+ * which checks whether all sender policies have their country of residence
+ * attribute set to `LOCAL_COUNTRY`. If that is the case, then the `amount`
+ * will be forwarded to the account address held in the state. Otherwise,
+ * the receive function will reject with `ContractError::NotLocalSender`.
+ */
+use concordium_std::*;
```
limemloh, line 15: If you want to support `no_std`
```suggestion
#![cfg_attr(not(feature = "std"), no_std)]
use concordium_std::*;
```

limemloh (approved):
I only have some minor comments

concordium-std/src/types.rs
```diff
@@ -635,14 +635,23 @@ pub type ReadOnlyCallContractResult<A> = Result<Option<A>, CallContractError<A>>
...
+#[derive(Clone, Copy, Default)]
 pub struct AttributesCursor {
     /// Current position of the cursor, starting from 0.
     /// Note that this is only for the variable attributes.
     /// `created_at` and `valid_to` will require.
     pub(crate) current_position: u32,
     /// The number of remaining items in the policy.
     pub(crate) remaining_items:  u16,
+    /// The total number of items. Used for creating new attribute cursors.
+    pub(crate) total_items:      u16,
```
abizjak, line 647: Do we need this? Can we not always get it by using
```
total_items = remaining_items + current_position
```
?

Bargsteen: Yes, we do. `current_position` relates to the number of bytes read _not_ the number of items.

abizjak (approved):

Issue #78, opened 2022-02-13
Better support for return values for failures.


abizjak:
**Task description**

With the introduction of synchronous calls in #71 contracts can return values now, even if they fail due to a logic error.
However the support for this in concordium-std is limited. We only support it via manually implementing `Into<Reject>` for your chosen error type, and to implement this one needs to produce a Vec<u8>. This is less ergonomic than our other support, where we do serialization behind the scenes.

This is related to #71 . It is not blocking for it, but it will likely need some API changes in the library, so it would be good to have it done before release.

Bargsteen:
Closed by https://github.com/Concordium/concordium-rust-smart-contracts/pull/145

PR #145, opened 2022-07-22
Enable deriving Reject for enums with fields

Bargsteen:
## Purpose

Addresses a primary part of #78 by allowing the user to derive `Reject` for enums with fields (if the fields implement `Serial`). 
We will likely also need to update some tools and documentation in order for this task to be finished entirely.

## Changes

- Add the ability to derive Reject for enums with fields (if the enum implements `Serial`).

## Checklist

      hard-to-understand areas.


limemloh (approved):
LGTM

Only a suggestion for an improvement

concordium-std-derive/src/lib.rs
```diff
@@ -1977,6 +1979,76 @@ fn reject_derive_worker(input: TokenStream) -> syn::Result<TokenStream> {
     Ok(gen.into())
 }
 
+/// Generate the cases for matching on the enum. For variants with fields, all
+/// fields are serialized into a buffer that is include in the return value. For
```
abizjak, line 1983:
```suggestion
/// fields are serialized into a buffer that is included in the return value. For
```

concordium-std-derive/src/lib.rs
```diff
@@ -1977,6 +1979,76 @@ fn reject_derive_worker(input: TokenStream) -> syn::Result<TokenStream> {
...
+        let variant_ident = &variant.ident;
+        match variant.fields {
+            syn::Fields::Named(_) => {
+                let field_names: Vec<_> =
+                    variant.fields.iter().map(|field| field.ident.clone().unwrap()).collect();
+                let pattern = quote! { {#(#field_names),*} };
+
+                let mut serial_fields = proc_macro2::TokenStream::new();
+                for n in field_names {
+                    serial_fields.extend(quote!(#n.serial(&mut buf);));
```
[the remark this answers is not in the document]
abizjak: I agree. Serial is a bit generic a name so it is better to be more robust.

concordium-std-derive/src/lib.rs
```diff
@@ -1907,8 +1907,10 @@ const RESERVED_ERROR_CODES: i32 = i32::MIN + 100;
 /// Creating custom enums for error types can provide meaningful error messages
 /// to the user of the smart contract.
 ///
-/// Note that at the moment, we can only derive fieldless enums and the return
-/// value is always [`None`][std::option::Option::None].
+/// For variants with fields, all fields are serialized into the buffer used in
+/// the return value, which is always [`Some`][std::option::Option::Some].
+/// For fieldless variants, the return value is always
+/// [`None`][std::option::Option::None].
```
abizjak, line 1913: I think this comment should be expanded a bit to also include that this error will be returned from the contract.

There is a complication, the same as it is now, that the schema will not support parsing this error.

So we probably need to update the schema for this to be fully done.

Although we can probably just merge this as is, it is not going to be worse than what we have now.

Bargsteen: I've added some more information to the comment. If that is not what you meant, then please let me know 😊

concordium-std-derive/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## Unreleased changes
 
 - Removed `derive(Serial)` and `derive(Deserial)` (moved to `concordium-contracts-common-derive`).
+- Add ability to `derive(Reject)` for enums *with fields*, where all fields
+  implement the `Serial` trait.
```
abizjak, line 7: Something seems wrong with t his changelog entry diff. Can you make sure to merge with `main` before merging into main so we don't get a mangled changelog.

abizjak (approved):

concordium-std-derive/CHANGELOG.md
Bargsteen: Ah, I was too quick when merging in the main. Nice catch ;)

abizjak (approved):

PR #150, opened 2022-08-05
fix balance in auction example

DOBEN:
## Purpose

The auction example was written by setting the self_balance in the TestHost to the smart contract balance instead of the smart contract + amount balance as described [here](https://docs.rs/concordium-std/latest/concordium_std/test_infrastructure/struct.TestHost.html#method.set_self_balance)

As a result, also the balance logic in the smart contract was written assuming that self_balance does not account for the amount which is a wrong assumption as can be seen [here](https://docs.rs/concordium-std/latest/concordium_std/trait.HasHost.html#tymethod.self_balance)

## Changes
- setting the self_balance in the TestHost to the smart contract + amount balance

 - updating balance logic in smart contract

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/auction/src/lib.rs
+++ b/examples/auction/src/lib.rs
@@ -158,9 +158,10 @@ fn view<'a, 'b, S: HasStateApi>(
     Ok(host.state())
 }
 
-/// ViewBalance function that returns the balance of the contract
-#[receive(contract = "auction", name = "viewBalance", return_value = "Amount")]
-fn view_balance<S: HasStateApi>(
+/// ViewHighestBid function that returns the highest bid which is the balance of
+/// the contract
+#[receive(contract = "auction", name = "viewHighestBid", return_value = "Amount")]
+fn view_highest_bid<S: HasStateApi>(
     _ctx: &impl HasReceiveContext,
     host: &impl HasHost<State, StateApiType = S>,
 ) -> ReceiveResult<Amount> {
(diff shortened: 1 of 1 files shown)


examples/auction/src/lib.rs
```diff
@@ -125,7 +125,7 @@ fn auction_bid<S: HasStateApi>(
     };
 
     // Ensure that the new bid exceeds the highest bid so far
-    ensure!(amount > balance, BidError::BidTooLow);
+    ensure!(amount > balance - amount, BidError::BidTooLow);
```
abizjak, line 128: I can see this is rigth, but it does look puzzling :)

abizjak: Maybe introduce a new variable
```
// Balance of the contract before the call
let previous_balance = balance - amount;
```

abizjak (approved):
Looks good. I have one suggestion to improve readability.

examples/auction/src/lib.rs
DOBEN: addressed

examples/auction/src/lib.rs
```diff
@@ -150,6 +158,15 @@ fn view<'a, 'b, S: HasStateApi>(
     Ok(host.state())
 }
 
+/// ViewBalance function that returns the balance of the contract
+#[receive(contract = "auction", name = "viewBalance", return_value = "Amount")]
```
limemloh, line 162: Off-chain you can already get the balance of a smart contract.
But this could be useful for another smart contract, I guess.
Another contract might be more interested in the current highest bid and not the balance, so in this case I would say it should be renamed.

PR #154, opened 2022-08-10
fix bug in Receiver schema

### LINKED ISSUE #153: Schema bug in the `Receiver`
**Bug Description**
Calling the `wrap` function on the wCCD smart contract with `sender != receiver` (The `Receiver` is a smart contract and not an account) fails with the following error `'wrap' in 'CIS2-wCCD-Proxy' at {"index":792,"subindex":0} failed with code -2147483627`

The requested schema is quite weird as follows:

> {
>     "data": "",
>     "to": {
>         "Contract": [
>            {
>                "index": 843,
>                "subindex": 0
>             },
>            {
>                "contract": "auction",
>                "func": "receiveToken"
>            }
>        ]
>    }
> }

Investigation by Emil:
A bug in the schema type for `Receiver` causes this behavior. The Receiver uses `OwnedEntrypointName` but the schema uses `OwnedReceiveName`.

Solution by Emil:
Replacing the `OwnedReceiveName` [here](https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/concordium-cis2/src/lib.rs#L839) with `schema::Type::String(schema::SizeLength::U16)`

The suggested solution was tested on testnet with the `wCCD` token and worked.

**Expected Result**

The `wrap` call should not fail when `Receiver` is a smart contract.

**Actual Result**

The `wrap` call fails when `Receiver` is a smart contract.

DOBEN:
## Purpose

Fix bug in Receiver schema so a smart contract can be passed as a `Receiver` via the schema.

closes #153 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-cis2/src/lib.rs
+++ b/concordium-cis2/src/lib.rs
@@ -836,7 +836,8 @@ impl schema::SchemaType for Receiver {
                 String::from("Contract"),
                 schema::Fields::Unnamed(vec![
                     ContractAddress::get_type(),
-                    OwnedReceiveName::get_type(),
+                    // The below string represents the function entrypoint
+                    schema::Type::String(schema::SizeLength::U16),
                 ]),
             ),
         ])


concordium-cis2/src/lib.rs
```diff
@@ -836,7 +836,7 @@ impl schema::SchemaType for Receiver {
                 String::from("Contract"),
                 schema::Fields::Unnamed(vec![
                     ContractAddress::get_type(),
-                    OwnedReceiveName::get_type(),
+                    schema::Type::String(schema::SizeLength::U16),
```
abizjak, line 839: Perhaps a short comment what the String represents would be helpful in the future. It is also technically not necessary to have U16 for length. U8 should suffice for entrypoint names.

abizjak (approved):

concordium-cis2/src/lib.rs
abizjak: I guess changing U16 -> U8 would break parsing, so never mind.

DOBEN: added a comment

abizjak (approved):

PR #165, opened 2022-09-21
Add templates

DOBEN:
## Purpose

Cargo-concordium can generate smart contracts based on templates. Provide some initial templates.

## Changes

Add `default` and `cis2-nft` templates

## Checklist

      hard-to-understand areas.


templates/cis2-nft/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[package]
+name = "cis2_nft"
```
abizjak, line 4: Should this not be parametrized, i.e., a placeholder, and then substituted by cargo init?

templates/cis2-nft/src/lib.rs
```diff
@@ -0,0 +1,1061 @@
...
+fn build_token_metadata_url(token_id: &ContractTokenId) -> String {
+    let mut token_metadata_url = String::from(TOKEN_METADATA_BASE_URL);
+    token_metadata_url.push_str(&token_id.to_string());
+    token_metadata_url
+}
+
+// Contract functions
+
+/// Initialize contract instance with no token types initially.
+#[init(contract = "CIS2-NFT")]
```
abizjak, line 301: Do we want to keep this name. We did briefly discuss that the name of the contract would be the same as the name of the project, did we not?

templates/default/cargo-generate.toml
```diff
@@ -0,0 +1,6 @@
+[template]
+cargo_generate_version = "0.16.0"
+
+[placeholders]
+description = { type="string", prompt="Description for the crate?" }
+currentAuthors = { type="string", prompt="Authors for this crate?" }
```
abizjak, line 6: Does this mean it will always prompt for authors? Does it default to those in `.gitconfig`?

abizjak (commented):
Looks good, but the CIS2 contract should probably be parametrized by a few more things.

templates/cis2-nft/cargo-generate.toml
```diff
@@ -0,0 +1,7 @@
+[template]
+cargo_generate_version = "0.16.0"
+
+[placeholders]
+description = { type="string", prompt="Description for the crate?" }
+currentAuthors = { type="string", prompt="Authors for this crate?" }
+TOKEN_METADATA_BASE_URL = { type="string", prompt="Token metadata base URL of the token?", default='https://some.example/token/' }
```
Bargsteen, line 7: Why is this in upper-case when the other placeholders are lower-cased?

templates/default/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Your smart contract state.
+#[derive(Serialize, SchemaType)]
+pub struct State {
+    // Your state
+}
+
+/// Your smart contract errors.
+#[derive(Reject, Serial, SchemaType)]
+enum Error {
+    /// YourError
```
Bargsteen, line 17:
```suggestion
    /// Your errors
```

templates/default/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    _ctx: &impl HasInitContext,
+    _state_builder: &mut StateBuilder<S>,
+) -> InitResult<State> {
+    // Your code
+
+    Ok(State {})
+}
+
+/// Payable Receive function.
+#[receive(contract = "{{crate_name}}", name = "receiveFunction", payable, mutable, error = "Error")]
```
Bargsteen, line 33: Perhaps?
```suggestion
#[receive(contract = "{{crate_name}}", name = "receive", payable, mutable, error = "Error")]
```

Bargsteen: Also the view function if you decide to change it 😊

templates/default/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+) -> Result<(), Error> {
+    // Your code
+
+    let error_occurred = false;
+
+    if error_occurred {
+        Err(Error::YourError)
+    } else {
+        Ok(())
+    }
```
Bargsteen, line 47: I think it would be good to include a parameter here. For example something like this? `42` was chosen due to a lack of imagination. Please use something else if you have a better idea.

```suggestion
    let input = ctx.parameter_cursor().get()?; // Returns Error::ParseError on failure

    if input == 42 {
        Ok(())
    } else {
        Err(Error::YourError)
    }
```

**Then we could also show how to set a parameter in a test of the receive function (and construct a `TestStateHost`).**

templates/default/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+        let mut state_builder = TestStateBuilder::new();
+
+        let state_result = your_smart_contract_init(&ctx, &mut state_builder);
+        state_result.expect_report("Contract initialization results in error");
+    }
+
+    #[concordium_test]
+    /// Test 1
+    fn test_1() {
+        // Your test
```
Bargsteen, line 78: Show positive test with setting a parameter and invoking the receive function with a test host and test context.

Bargsteen (changes requested):
Nice work. I think we should expand the default template a little bit to include using a parameter and a test of the receive function.

templates/cis2-nft/cargo-generate.toml
DOBEN: The variable is `const` in the Rust smart contract code and for such variables, it is a convention to use screaming_snake_case (I just copied the variable from the smart contract code here that is how it ended up like this). 

Yes, I agree that in the above code, it would be more consistent if we change the naming to the same style then the other variables. Changed it.

templates/default/src/lib.rs
DOBEN:
```
/// Your smart contract errors.
#[derive(Debug, PartialEq, Eq, Reject, Serial, SchemaType)]
enum Error {
    /// Failed parsing the parameter.
    #[from(ParseError)]
    ParseParamsError,
    /// Your error
    YourError,
}
```

templates/default/src/lib.rs
DOBEN: I added a test that throws an error and a positive test.

templates/default/src/lib.rs
DOBEN: The input parameter is now the boolean variable `throw_error`. If `true` the receive function will throw a custom error, if `false` the receive function executes successfully.

templates/cis2-nft/src/lib.rs
```diff
@@ -0,0 +1,1061 @@
+//! Note: This contract is not meant for production, it is used to illustrate
+//! how to use the standard library and the tooling Concordium provides. There
+//! is no claim that the logic of the contract is reasonable, or safe.
```
abizjak, line 3: Do we want to have this kind of thing in a template we advertise as a cis2-nft?

It seems to me that people will use this if we make it easy, like we are doing. Do we want to revise this disclaimer, or what should we do?

templates/default/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+edition = "2018"
+license = "MPL-2.0"
+authors = [ "{{authors}}" ]
+description = "{{description}}"
+
+[dependencies]
+concordium-std = "4.0.0"
+
+[lib]
+crate-type=["cdylib", "rlib"]
```
abizjak, line 15: Shouldn't we put the build arguments `-s` and `codegen-units=1` here as well?

templates/default/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+#[concordium_cfg_test]
+mod tests {
+    use super::*;
+    use test_infrastructure::*;
+
+    type ContractResult<A> = Result<A, Error>;
+
+    fn expect_error<E, T>(expr: Result<T, E>, err: E, msg: &str)
```
abizjak, line 76: Perhaps we should remove this, or alternatively add a comment to it, since the type signature is rather complex, and the utility is not entirely clear either.

templates/default/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+    where
+        E: Eq + Debug,
+        T: Debug,
+    {
+        let actual = expr.expect_err_report(msg);
+        claim_eq!(actual, err);
+    }
+
+    #[concordium_test]
+    /// Init test
```
abizjak, line 86: Could we do a bit better here
```suggestion
    /// Test that initializing the contract succeeds with some state.
```

templates/default/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+        let ctx = TestInitContext::empty();
+
+        let mut state_builder = TestStateBuilder::new();
+
+        let state_result = init(&ctx, &mut state_builder);
+        state_result.expect_report("Contract initialization results in error");
+    }
+
+    #[concordium_test]
+    /// Throw no error test
```
abizjak, line 97:
```suggestion
    /// Test that invoking the `receive` endpoint with the `false` parameter succeeds in updating the contract.
```

templates/default/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+
+        // Call the contract function.
+        let result: ContractResult<()> = receive(&ctx, &mut host);
+
+        // Check the result.
+        claim!(result.is_ok(), "Results in rejection");
+    }
+
+    #[concordium_test]
+    /// Throw an error test
```
abizjak, line 122: Likewise.

abizjak (commented):
It looks good overall. Before merging, the main thing that I don't understand is how seriously should we claim that the nft contract is "good to go". We have to expect people will just deploy it as is.

templates/cis2-nft/src/lib.rs
DOBEN: Since this disclaimer is part of the template, we can revise/change it at any time. We could start with this disclaimer and once the wCCD token contract is security audited (+ the cis2 library), we have higher confidence that we can remove the disclaimer. Internally, we could security audit the wCCD as well as the cis2-nft token contracts if both of them are used most likely in production.

abizjak: That is a good point.

abizjak (approved):
Looks good for this scope.

Can you also add a task to add these templates to the CI, so that we remember to maintain them? It can be done as a separate PR later, I just don't want to forget it :)

PR #167, opened 2022-09-28
Add linting for smart contract templates

DOBEN:
## Purpose

Add smart contract templates to CI pipeline.

3 things are checked as part of the CI pipeline:
- All templates are generated with cargo-generate and it is checked that 'cargo test' command can be executed without errors on them.
- The cis2-nft template is generated with cargo-generate and it is checked that the code is equivalent to the cis2-nft smart contract in the example folder.
- All templates are generated with cargo-generate and it is checked that the schemas can be built as part of the 'clippy' command.

## Checklist

      hard-to-understand areas.


abizjak:
Have you considered an alternate approach of instantiating a template and checking it yields an example.

So for example for the nft template we would always have that instantiating it, using `cargo-generate` (if possible) with the relevant authors and name would yield exactly the cis2 example. It seems like this should be possible using the `--path` flag to cargo-generate.

abizjak:
This would prevent some of the hackiness with custom Cargo_*.toml files.

abizjak:
Of course we can modify the examples a little if needed.

DOBEN:
I removed the disclaimer "this smart contract is not meant for production ...." at the top of the templates. This is so we can easier compare the generated templates with the smart contracts in the example folder.

.github/workflows/linter.yml
```diff
@@ -55,6 +55,131 @@ jobs:
...
+      # otherwise `cargo` runs would fail
+      # see https://github.com/rust-lang/cargo/issues/9922
+      # Run all tests, including doc tests.
+      - name: Run cargo test
+        run: |
+          mv $PROJECT_NAME ${{ runner.temp }}/
+          cd ${{ runner.temp }}/$PROJECT_NAME
+          cargo test
+
+  cargo-generate-templates-comparison:
```
abizjak, line 97: Could you add a short description of this step (and the previous one).

.github/workflows/linter.yml
```diff
@@ -55,6 +55,131 @@ jobs:
...
+        with:
+          toolchain: stable
+
+      # Run diff to check if there are any differences between the generated smart contracts and the example smart contracts
+      - name: Run diff
+        run: |
+          mv $PROJECT_NAME ${{ runner.temp }}/
+          sed -i "s/root/Concordium <developers@concordium.com>/g" ${{ runner.temp }}/$PROJECT_NAME/Cargo.toml
+          sed -i "s/\"4.0.0\"/{path = \"..\/..\/concordium-std\", default-features = false}/g" ${{ runner.temp }}/$PROJECT_NAME/Cargo.toml
+          sed -i "s/\"1.2.0\"/{path = \"..\/..\/concordium-cis2\", default-features = false}/g" ${{ runner.temp }}/$PROJECT_NAME/Cargo.toml
```
abizjak, line 131: Maybe add a comment here as to why this is needed.

examples/cis2-nft/src/lib.rs
```diff
@@ -801,7 +801,7 @@ mod tests {
                 TokenMetadataEvent {
                     token_id:     TOKEN_0,
                     metadata_url: MetadataUrl {
-                        url:  "https://some.example/token/00000000".to_string(),
+                        url:  TOKEN_METADATA_BASE_URL.to_string() + &"00000000".to_string(),
```
abizjak, line 804: Would it not be better to use `format!("{}/0000000", TOKEN_METADATA_BASE_URL)`?

examples/cis2-nft/src/lib.rs
```diff
@@ -813,7 +813,7 @@ mod tests {
                 TokenMetadataEvent {
                     token_id:     TOKEN_1,
                     metadata_url: MetadataUrl {
-                        url:  "https://some.example/token/2A000000".to_string(),
+                        url:  TOKEN_METADATA_BASE_URL.to_string() + &"2A000000".to_string(),
```
abizjak, line 816: And here.

abizjak (approved):
Looks good. I would appreciate it if you added a bit more documentation to the (relatively complex) github actions.

PR #177, opened 2022-10-13
Add event schemas

### LINKED ISSUE #170: Event schemas
**Task description**

Update concordium-std/concordium-std-derive to support event schema creation/serialization/deserialization.

DOBEN:
## Purpose

close #170 

## Changes

Add event schemas. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-cis2/src/lib.rs
+++ b/concordium-cis2/src/lib.rs
@@ -62,14 +62,6 @@ pub const UPDATE_OPERATOR_EVENT_TAG: u8 = u8::MAX - 3;
 /// Tag for the CIS2 TokenMetadata event.
 pub const TOKEN_METADATA_EVENT_TAG: u8 = u8::MAX - 4;
 
-pub const EVENT_TAG_VECTOR: [u8; 5] = [
-    TRANSFER_EVENT_TAG,
-    MINT_EVENT_TAG,
-    BURN_EVENT_TAG,
-    UPDATE_OPERATOR_EVENT_TAG,
-    TOKEN_METADATA_EVENT_TAG,
-];
-
 /// Sha256 digest
 pub type Sha256 = [u8; 32];
 
@@ -750,17 +742,6 @@ pub struct TransferEvent<T: IsTokenId, A: IsTokenAmount> {
     pub to:       Address,
 }
 
-impl<T: IsTokenId, A: IsTokenAmount> schema::SchemaType for TransferEvent<T, A> {
-    fn get_type() -> schema::Type {
-        schema::Type::Struct(schema::Fields::Named(vec![
-            (String::from("token_id"), T::get_type()),
-            (String::from("amount"), A::get_type()),
-            (String::from("from"), Address::get_type()),
-            (String::from("to"), Address::get_type()),
-        ]))
-    }
-}
-
 /// An untagged event of tokens being minted, could be a new token type or
 /// extending the total supply of existing token.
 /// For a tagged version, use `Cis2Event`.
@@ -774,16 +755,6 @@ pub struct MintEvent<T: IsTokenId, A: IsTokenAmount> {
     pub amount:   A,
     /// The initial owner of these newly minted amount of tokens.
     pub owner:    Address,
-}
-
-impl<T: IsTokenId, A: IsTokenAmount> schema::SchemaType for MintEvent<T, A> {
-    fn get_type() -> schema::Type {
-        schema::Type::Struct(schema::Fields::Named(vec![
-            (String::from("token_id"), T::get_type()),
-            (String::from("amount"), A::get_type()),
-            (String::from("owner"), Address::get_type()),
-        ]))
-    }
 }
 
 /// An untagged event of some amount of a token type being burned.
@@ -800,16 +771,6 @@ pub struct BurnEvent<T: IsTokenId, A: IsTokenAmount> {
     pub owner:    Address,
 }
 
-impl<T: IsTokenId, A: IsTokenAmount> schema::SchemaType for BurnEvent<T, A> {
-    fn get_type() -> schema::Type {
-        schema::Type::Struct(schema::Fields::Named(vec![
-            (String::from("token_id"), T::get_type()),
-            (String::from("amount"), A::get_type()),
-            (String::from("owner"), Address::get_type()),
-        ]))
-    }
-}
-
 /// An untagged event of an update to an operator address for an owner address.
 /// For a tagged version, use `Cis2Event`.
 // Note: For the serialization to be derived according to the CIS2
@@ -824,16 +785,6 @@ pub struct UpdateOperatorEvent {
     pub operator: Address,
 }
 
-impl schema::SchemaType for UpdateOperatorEvent {
-    fn get_type() -> schema::Type {
-        schema::Type::Struct(schema::Fields::Named(vec![
-            (String::from("update"), OperatorUpdate::get_type()),
-            (String::from("owner"), Address::get_type()),
-            (String::from("operator"), Address::get_type()),
-        ]))
-    }
-}
-
 /// An untagged event for setting the metadata for a token.
 /// For a tagged version, use `Cis2Event`.
 // Note: For the serialization to be derived according to the CIS2
@@ -844,15 +795,6 @@ pub struct TokenMetadataEvent<T: IsTokenId> {
     pub token_id:     T,
     /// The location of the metadata.
     pub metadata_url: MetadataUrl,
-}
-
-impl<T: IsTokenId> schema::SchemaType for TokenMetadataEvent<T> {
-    fn get_type() -> schema::Type {
-        schema::Type::Struct(schema::Fields::Named(vec![
-            (String::from("token_id"), T::get_type()),
-            (String::from("metadata_url"), MetadataUrl::get_type()),
-        ]))
-    }
 }
 
 /// Tagged CIS2 event to be serialized for the event log.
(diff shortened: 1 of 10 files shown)


concordium-std-derive/src/lib.rs
```diff
@@ -1053,65 +1084,308 @@ fn contract_function_schema_tokens(
...
+                }),
+            }
+        }
+        (Some(parameter_ty), Some(return_value_ty), Some(error_ty), Some(event_ty)) => {
+            let parameter_ty = parameter_ty.parse::<syn::Type>()?;
+            let return_value_ty = return_value_ty.parse::<syn::Type>()?;
+            let error_ty = error_ty.parse::<syn::Type>()?;
+            let event_ty = event_ty.parse::<syn::Type>()?;
+
+            match event_tags_option {
```
abizjak, line 1281: I think if you refactor the type to be a `struct` with `Option` values this will become much less code.

abizjak (commented):
I think changes will be needed in the `derive(SchemaType)` if you (hopefully) adopt my suggestion of adding a new variant (or modify the existing Enum variant) to the `SchemaType` to be able to supply custom tags.

concordium-std-derive/src/lib.rs
```diff
@@ -998,6 +1006,7 @@ fn receive_worker(attr: TokenStream, item: TokenStream) -> syn::Result<TokenStre
         parameter_option,
         return_value_option,
         error_option,
+        None, // Event schema should be attached to the `init` function and is globally available.
```
abizjak, line 1009: This will likely change, assuming you change the COntractV3 modelling in contracts-common.

abizjak (commented):
This looks good, however could you add the event schema to some examples as well?

At least the wCCD and maybe cis2-nft and cis2-multi would benefit from it.

DOBEN:
```
This looks good, however could you add the event schema to some examples as well?

At least the wCCD and maybe cis2-nft and cis2-multi would benefit from it.
```

I included it in the wCCD and the cis2-multi example. If I add it to cis2-nft, I would also have to change the associated templates but they fail because they have a fixed, old version of `concordium-cis2` library.

abizjak:
> ```
> This looks good, however could you add the event schema to some examples as well?
> 
> At least the wCCD and maybe cis2-nft and cis2-multi would benefit from it.
> ```
> 
> I included it in the wCCD and the cis2-multi example. If I add it to cis2-nft, I would also have to change the associated templates but they fail because they have a fixed, old version of `concordium-cis2` library.

Ah, good point. We need to update the template after we release the library then.

concordium-cis2/src/lib.rs
```diff
@@ -813,6 +871,48 @@ pub enum Cis2Event<T: IsTokenId, A: IsTokenAmount> {
...
+            EVENT_TAG_VECTOR[4],
+            (
+                "TokenMetadata".to_string(),
+                schema::Fields::Unnamed(Vec::from([TokenMetadataEvent::<T>::get_type()])),
+            ),
+        );
+        schema::Type::EnumTag(event_map)
+    }
+}
+
```
abizjak, line 915: We need to add a changelog entry stating there is a schematype for this now.

concordium-cis2/src/lib.rs
```diff
@@ -813,6 +871,48 @@ pub enum Cis2Event<T: IsTokenId, A: IsTokenAmount> {
...
+            (
+                "UpdateOperator".to_string(),
+                schema::Fields::Unnamed(Vec::from([UpdateOperatorEvent::get_type()])),
+            ),
+        );
+        event_map.insert(
+            EVENT_TAG_VECTOR[4],
+            (
+                "TokenMetadata".to_string(),
+                schema::Fields::Unnamed(Vec::from([TokenMetadataEvent::<T>::get_type()])),
```
abizjak, line 909: I'm guessing this implementation will change based on the outcome of the discussion on slack?

abizjak (commented):
Looks good overall.

The only thing that seems to need a bit more work is the SchemaType implementation for CIS2Event.

examples/cis2-wccd/src/lib.rs
```diff
@@ -707,7 +711,7 @@ fn contract_supports<S: HasStateApi>(
 /// - Sender is not the owner of the contract instance.
 /// - It fails to parse the parameter.
 #[receive(
-    contract = "CIS2-wCCD",
+    contract = "cis_wCCD",
```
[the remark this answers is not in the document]
DOBEN: The contract name is used as for the event_schema name and I got an error when using contract names with `-`.

[here](https://github.com/Concordium/concordium-rust-smart-contracts/blob/c7ec90c7cfc85111a622f7955397145ec8ba119b/concordium-std-derive/src/lib.rs#L1082)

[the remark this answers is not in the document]
abizjak: I see. I think that counts as a "good reason". :)

concordium-cis2/src/lib.rs
```diff
@@ -813,6 +871,48 @@ pub enum Cis2Event<T: IsTokenId, A: IsTokenAmount> {
     TokenMetadata(TokenMetadataEvent<T>),
 }
 
+impl<T: IsTokenId, A: IsTokenAmount> schema::SchemaType for Cis2Event<T, A> {
+    fn get_type() -> schema::Type {
+        let mut event_map = BTreeMap::new();
+        event_map.insert(
+            EVENT_TAG_VECTOR[0],
```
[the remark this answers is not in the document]
DOBEN: Removed the `EVENT_TAG_VECTOR` and using the `TRANSFER_EVENT` directly.

concordium-cis2/src/lib.rs
```diff
@@ -771,11 +800,21 @@ pub struct BurnEvent<T: IsTokenId, A: IsTokenAmount> {
...
 
+impl<T: IsTokenId, A: IsTokenAmount> schema::SchemaType for BurnEvent<T, A> {
+    fn get_type() -> schema::Type {
+        schema::Type::Struct(schema::Fields::Named(vec![
+            (String::from("token_id"), T::get_type()),
+            (String::from("amount"), A::get_type()),
+            (String::from("owner"), Address::get_type()),
+        ]))
+    }
+}
```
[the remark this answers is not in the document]
DOBEN: Added `SchemaType` back to the single events and implemented the SchemaType for `Cis2Event<T, A>` globally (without the extra brackets that was caused by the unnamed field)

concordium-std-derive/src/lib.rs
```diff
@@ -1048,6 +1060,41 @@ fn contract_function_optional_args_tokens(
...
+            #[export_name = #schema_name]
+            pub extern "C" fn #schema_ident() -> *mut u8 {
+                #event_embed
+                let schema_bytes = concordium_std::to_bytes(&#event_schema);
+                concordium_std::put_in_memory(&schema_bytes)
+            }
+        })
+    } else {
+        Ok(proc_macro2::TokenStream::new())
+    }
```
abizjak, line 1095: See my comment in the concordium-base PR about maybe making this a bit simpler.

abizjak (commented):
Should we not add the event schema type to all our CIS2 examples as well?

DOBEN:
`Should we not add the event schema type to all our CIS2 examples as well?`
We can add it to the remaining `cis2_nft` once we have published the new concordium_cis2 library and can update this example together with the templates.

concordium-std-derive/Cargo.toml
```diff
@@ -17,7 +17,7 @@ proc-macro2 = "1.0"
```
abizjak, line 17: We also want to bump the version of concordium-std-derive. And add to its changelog that we now support `event` in the schema derivation macros.

concordium-std-derive/src/lib.rs
```diff
@@ -1048,6 +1060,37 @@ fn contract_function_optional_args_tokens(
     (setup_fn_args, fn_args)
 }
 
+#[cfg(feature = "build-schema")]
+fn contract_function_schema_tokens2(
```
abizjak, line 1064: Perhaps name this with something with `event` in the name. the `2` suffix does not say very much.

concordium-std-derive/src/lib.rs
```diff
@@ -1048,6 +1060,37 @@ fn contract_function_optional_args_tokens(
...
+    event_option: Option<syn::LitStr>,
+    contract_name: String,
```
abizjak, line 1066: It would probably be best if you took in a `rust_name`, as we do for `contract_function_schema`.

And name the function using that name. That will avoid that issue where you had to rename some contract names.

concordium-std-derive/src/lib.rs
```diff
@@ -1048,6 +1060,37 @@ fn contract_function_optional_args_tokens(
...
+            let event = <#event_ty as schema::SchemaType>::get_type();
+            let schema_bytes = concordium_std::to_bytes(&event);})
+        }
+        _ => None,
+    };
+
+    // Only produce the schema function if the event was set.
+    if let Some(construct_schema_bytes) = event_embed {
+        let schema_name = format!("concordium_event_schema_{}", contract_name);
+        let schema_ident = format_ident!("concordium_event_schema_{}", contract_name);
```
abizjak, line 1081:
```suggestion
        let schema_ident = format_ident!("concordium_event_schema_{}", rust_name);
```

concordium-std-derive/src/lib.rs
```diff
@@ -627,6 +633,9 @@ fn init_worker(attr: TokenStream, item: TokenStream) -> syn::Result<TokenStream>
     let parameter_option = init_attributes.optional.parameter;
     let return_value_option = None; // Return values are currently not supported on init.
     let error_option = init_attributes.optional.error;
+    let event_option = init_attributes.event;
+    let contract_name = wasm_export_fn_name.clone();
```
abizjak, line 637: With respect to my comment in the concordium-base PR, the `init_` comes from this it seems. 

We could in principle remove it here. But I think it's OK the way it is.

concordium-std/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "concordium-std"
-version = "4.0.0"
+version = "4.1.0"
```
abizjak, line 3: We also need to add to the changelog.

concordium-std/Cargo.toml
```diff
@@ -20,11 +20,11 @@ ed25519-zebra = { version = "2.2", optional = true }
 
 [dependencies.concordium-std-derive]
 path = "../concordium-std-derive"
-version = "4"
+version = "4.1"
```
abizjak, line 23: I think this will need to be 4.2. 4.1 is already released, and does not support the `event` option (There was a small release of this last week adding a tiny feature).

abizjak (approved):
I think this looks good now.

There are a few minor issues to address
- adding changes to the changelog
- versions need to be fixed
- the rust_name should be used for the generated function. This will avoid user confusion if they name their contract using, e.g., `-` in the name.

Issue #181, opened 2022-10-24
concordium-std support for chain queries


abizjak:
This is the concordium-std counterpart to https://github.com/Concordium/concordium-node/issues/562

We need to provide a high-level api for the queries, and support for the test-host to mock responses for these queries.

abizjak:
This has been completed by #185

PR #185, opened 2022-11-01
Account balance, contract balance and exchange rate queries

limemloh:
## Purpose

Closes https://github.com/Concordium/concordium-rust-smart-contracts/issues/181

## Changes

- Extend `HasHost` with functions for querying account balances, contract balances and the current exchange rates.
- Extend `TestHost` with functions for setting up exchange rates, contract and account balances, used for queries.
- Added a minimum raise to the auction contract example, which sets a minimum amount require to overbid the previous bid.
  The minimum raise is specified in Euro cent and uses the chain exchange rate to convert the bid into Euro.
- Set `concordium-std` and `concordium-cis2` to version `5.0.0`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std/src/impls.rs
+++ b/concordium-std/src/impls.rs
@@ -2909,7 +2909,7 @@ impl Serial for AccountBalance {
     fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
         self.total.serial(out)?;
         self.staked.serial(out)?;
-        self.locked.serial(&mut *out)?;
+        self.locked.serial(out)?;
         Ok(())
     }
 }


concordium-cis2/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased changes
 
-- Add `From` implementation from types implementing `From<UpgradeError>` to `Cis2Error`.
+- Add `From` implementation from types implementing `From<UpgradeError>`, `From<QueryAccountBalanceError>` or `From<QueryContractBalanceError>` to `Cis2Error`.
```
abizjak, line 5: Upgrade concordium-std version to ...

concordium-std/src/impls.rs
```diff
@@ -1893,6 +1918,76 @@ fn parse_call_response_code(code: u64) -> CallContractResult<ExternCallResponse>
     }
 }
 
+#[inline(always)]
+fn is_invoke_failure_code(code: u64) -> bool {
+    code & 0xffff_ff00_0000_0000 == 0xffff_ff00_0000_0000
+}
```
abizjak, line 1924: Please document a bit. This is hard to understand without having a lot of prior knowledge.

At least when this is inlined there is some comment, but when it is on its own it is har.d

concordium-std/src/impls.rs
```diff
@@ -1893,6 +1918,76 @@ fn parse_call_response_code(code: u64) -> CallContractResult<ExternCallResponse>
...
+/// This is necessary since Wasm only allows us to pass simple scalars as
+/// parameters. Everything else requires passing data in memory, or via host
+/// functions, both of which are difficult.
+///
+/// The response is encoded as follows.
+/// - success is encoded as 0
+fn parse_query_exchange_rates_response_code(code: u64) -> ExternCallResponse {
+    if is_invoke_failure_code(code) {
+        unsafe { crate::hint::unreachable_unchecked() }
+    }
```
abizjak, line 1985: Add a comment that this query never fails in the node.

concordium-std/src/impls.rs
```diff
@@ -1930,6 +2025,40 @@ fn invoke_contract_construct_parameter(
...
+        )
+    };
+    let mut return_value = parse_query_account_balance_response_code(response)?;
+    Ok(AccountBalance::deserial(&mut return_value).unwrap_abort())
+}
+
+/// Helper factoring out the common behaviour of contract_balance for the
+/// two extern hosts below.
+fn query_contract_balance_worker(address: &ContractAddress) -> QueryContractBalanceResult {
+    let mut data = Vec::with_capacity(16);
```
abizjak, line 2045: We should not be using a vector here.

```
let mut data = [u8;16]
data[..8].copy_from_slice(&data.index.to_le_bytes());
data[8..].copy_from_slice(&data.subindex.to_le_bytes());
...
```

abizjak: We need to avoid using allocations as much as possible.

concordium-std/src/impls.rs
```diff
@@ -2180,6 +2319,16 @@ impl HasHost<ExternStateApi> for ExternLowLevelHost {
...
+    fn account_balance(&self, address: AccountAddress) -> QueryAccountBalanceResult {
+        query_account_balance_worker(&address)
+    }
+
+    fn contract_balance(&self, address: ContractAddress) -> QueryContractBalanceResult {
+        query_contract_balance_worker(&address)
+    }
+
+    fn exchange_rates(&self) -> ExchangeRates { query_exchange_rates_worker() }
+
```
abizjak, line 2331:
```
#[inline(always)]
```

concordium-std/src/impls.rs
```diff
@@ -2128,6 +2257,16 @@ where
...
 
+    fn account_balance(&self, address: AccountAddress) -> QueryAccountBalanceResult {
+        query_account_balance_worker(&address)
+    }
+
+    fn contract_balance(&self, address: ContractAddress) -> QueryContractBalanceResult {
+        query_contract_balance_worker(&address)
+    }
+
+    fn exchange_rates(&self) -> ExchangeRates { query_exchange_rates_worker() }
```
abizjak, line 2268:
```
#[inline(always)]
```

concordium-std/src/test_infrastructure.rs
```diff
@@ -1287,21 +1287,32 @@ pub struct TestHost<State> {
...
+    mocking_fns:             Rc<RefCell<MockFnMap<State>>>,
     /// Transfers the contract has made during its execution.
-    transfers:        RefCell<Vec<(AccountAddress, Amount)>>,
+    transfers:               RefCell<Vec<(AccountAddress, Amount)>>,
     /// The contract balance. This is updated during execution based on contract
     /// invocations, e.g., a successful transfer from the contract decreases it.
-    contract_balance: RefCell<Amount>,
+    contract_balance:        RefCell<Amount>,
+    /// The address of the instance. This is used when querying its own balance.
+    contract_address:        Option<ContractAddress>,
```
abizjak, line 1297: I'm not exactly sure why this is necessary. I don't think there is special logic for this, is there?

concordium-std/src/test_infrastructure.rs
```diff
@@ -1287,21 +1287,32 @@ pub struct TestHost<State> {
...
-    mocking_fns:      Rc<RefCell<MockFnMap<State>>>,
+    mocking_fns:             Rc<RefCell<MockFnMap<State>>>,
     /// Transfers the contract has made during its execution.
-    transfers:        RefCell<Vec<(AccountAddress, Amount)>>,
+    transfers:               RefCell<Vec<(AccountAddress, Amount)>>,
     /// The contract balance. This is updated during execution based on contract
     /// invocations, e.g., a successful transfer from the contract decreases it.
-    contract_balance: RefCell<Amount>,
+    contract_balance:        RefCell<Amount>,
+    /// The address of the instance. This is used when querying its own balance.
```
abizjak, line 1296: Do we need to have this as a special case? Why dont' we just use the map below and if the user cares about querying their own balance they can set it up there.

concordium-std/src/test_infrastructure.rs
```diff
@@ -1460,9 +1478,46 @@ impl<State: Serial + DeserialWithState<TestStateApi> + StateClone<TestStateApi>>
...
         Ok(res)
     }
 
+    fn account_balance(&self, address: AccountAddress) -> QueryAccountBalanceResult {
+        if self.missing_accounts.contains(&address) {
+            Err(QueryAccountBalanceError)
+        } else if let Some(balances) = self.query_account_balances.borrow().get(&address) {
+            Ok(*balances)
+        } else {
+            fail!("No account balance for {:?} has been set up.", address)
```
abizjak, line 1495: Why not use the display implementation? Is it not available here?

concordium-std/src/test_infrastructure.rs
```diff
@@ -1460,9 +1478,46 @@ impl<State: Serial + DeserialWithState<TestStateApi> + StateClone<TestStateApi>>
...
+        }
+    }
+
+    fn contract_balance(&self, address: ContractAddress) -> QueryContractBalanceResult {
+        // If the contract address is set and matches, we return the contract balance
+        // instead.
+        if Some(address) == self.contract_address {
+            Ok(self.contract_balance.borrow().to_owned())
```
abizjak, line 1503: I am wondering whether it is necessay to have this special case. What goes wrong if we handle it generically?

concordium-std/src/traits.rs
```diff
@@ -348,6 +349,16 @@ pub trait HasHost<State>: Sized {
         self.invoke_contract_raw_read_only(to, Parameter(&param), method, amount)
     }
 
+    /// Get the current exchange rates for Euro per Energy and micro CCD per
+    /// Euro.
+    fn exchange_rates(&self) -> ExchangeRates;
+
+    /// Get the current balance of an account.
+    fn account_balance(&self, address: AccountAddress) -> QueryAccountBalanceResult;
```
abizjak, line 357: We need to document the semantics better here.

concordium-std/src/traits.rs
```diff
@@ -348,6 +349,16 @@ pub trait HasHost<State>: Sized {
...
+
+    /// Get the current balance of a contract instance.
```
abizjak, line 359: And here. The case of the balance of the contract's own address is rather ambiguous.

concordium-std/src/types.rs
```diff
@@ -581,6 +584,69 @@ pub struct ExternReturnValue {
...
 }
 
+/// The current public balances of an account.
+#[derive(Debug, Copy, Clone)]
+pub struct AccountBalance {
+    /// The total balance of the account including the staked and locked amount.
+    pub(crate) total:  Amount,
+    /// The current staked amount of the account.
+    pub(crate) staked: Amount,
+    /// The current locked amount of the account.
```
abizjak, line 594:
```suggestion
    /// The current amount locked in releases that resulted from transfers with schedule.
```

concordium-std/src/types.rs
```diff
@@ -581,6 +584,69 @@ pub struct ExternReturnValue {
...
+    pub(crate) locked: Amount,
+}
+
+impl AccountBalance {
+    pub fn new(total: Amount, staked: Amount, locked: Amount) -> Self {
```
abizjak, line 599: Documentaiton.

concordium-std/src/test_infrastructure.rs
limemloh: It is not available here.
We need base58check which is only available when `derive-serde` is enabled in `concordium-contracts-common`

concordium-std/src/traits.rs
limemloh: I have added more to this

concordium-std/src/test_infrastructure.rs
```diff
@@ -1396,6 +1408,12 @@ impl<State: Serial + DeserialWithState<TestStateApi> + StateClone<TestStateApi>>
                 if amount.micro_ccd > 0 {
                     *self.contract_balance.borrow_mut() -= amount;
                 }
+
+                // Update the receiver query balance.
+                if let Some(balance) = self.query_contract_balances.borrow_mut().get_mut(to) {
+                    *balance += amount;
+                }
```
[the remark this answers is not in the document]
limemloh: The current behavior is to use `contract_balance` for the contracts own balance and not the map, meaning we don't have to update the sender in the map.

concordium-std/src/test_infrastructure.rs
limemloh: Resolving this to keep the discussion in https://github.com/Concordium/concordium-rust-smart-contracts/pull/185/files/6ef5176a4dc3f1e8b614132fd114c19d95de3a01#r1012234240

concordium-std/src/test_infrastructure.rs
limemloh: Resolving this to keep the discussion in https://github.com/Concordium/concordium-rust-smart-contracts/pull/185/files/6ef5176a4dc3f1e8b614132fd114c19d95de3a01#r1012234240

concordium-std/src/test_infrastructure.rs
```diff
@@ -1588,6 +1648,36 @@ impl<State: Serial + DeserialWithState<TestStateApi>> TestHost<State> {
...
+
+    pub fn setup_contract_balance(&mut self, address: ContractAddress, balance: Amount) {
+        self.query_contract_balances.borrow_mut().insert(address, balance);
+
+        // Update the contract balance if the address matches.
+        if let Some(self_address) = self.contract_address {
+            if self_address == address {
+                *self.contract_balance.borrow_mut() = balance;
+            }
+        }
```
Bargsteen, line 1674: I wouldn't expect this to change the self balance. Perhaps we should fail instead.

We could add a helper function for querying the balance, which takes in the optional self balance and address along with the contract query map? So we can have a single location for the logic regarding this special case?

And perhaps we should change the name to `setup_query_contract_balance` to distinguish it more from the `setup_self_balance` method.

concordium-std/src/types.rs
```diff
@@ -581,6 +584,93 @@ pub struct ExternReturnValue {
...
+    /// staking.
+    pub(crate) staked: Amount,
+    /// The current amount locked in releases that resulted from transfers with
+    /// schedule. A locked amount can still be used for staking.
+    pub(crate) locked: Amount,
+}
+
+impl AccountBalance {
+    /// Construct a new account balance.
+    pub fn new(total: Amount, staked: Amount, locked: Amount) -> Self {
```
Bargsteen, line 603: We could check that `total >= max(locked, staked)`

concordium-std/src/types.rs
```diff
@@ -581,6 +584,93 @@ pub struct ExternReturnValue {
...
+    pub fn staked(&self) -> Amount { self.staked }
+
+    /// The current amount locked in releases that resulted from transfers with
+    /// schedule. A locked amount can still be used for staking.
+    pub fn locked(&self) -> Amount { self.locked }
+
+    /// The current available balance of the account. This is the amount
+    /// an account currently have available for transfering and is not
+    /// staked or locked in releases by scheduled transfers.
+    pub fn available(&self) -> Amount { self.total - max(self.locked, self.staked) }
```
Bargsteen, line 626: This might underflow due to the missing check in `new`.

concordium-std/src/types.rs
```diff
@@ -581,6 +584,93 @@ pub struct ExternReturnValue {
...
+            euro_per_energy,
+            amount_per_euro,
+        }
+    }
+
+    /// Euro per NRG exchange rate.
+    pub fn euro_per_energy(&self) -> ExchangeRate { self.euro_per_energy }
+
+    /// Micro CCD per Euro exchange rate.
+    pub fn amount_per_euro(&self) -> ExchangeRate { self.amount_per_euro }
```
Bargsteen, line 651: Perhaps make the name more explicit so you don't have to read the comment.
```suggestion
    pub fn microCCD_per_euro(&self) -> ExchangeRate { self.amount_per_euro }
```

Bargsteen: Especially because `amount` in the next two methods actually means `CCD` instead of `microCCD`.

examples/auction/src/lib.rs
```diff
@@ -2,8 +2,12 @@
 //!
 //! Accounts can invoke the bid function to participate in the auction.
 //! An account has to send some CCD when invoking the bid function.
-//! This CCD amount has to exceed the current highest bid to be accepted by the
-//! smart contract.
+//! This CCD amount has to exceed the current highest bid by a minimum raise
+//! to be accepted by the smart contract.
+//!
+//! The minimum raise is set when initializing and defined in Euro cent.
```
Bargsteen, line 8:
```suggestion
//! The minimum raise is set when initializing and is defined in Euro cent.
```

examples/auction/src/lib.rs
```diff
@@ -133,7 +142,16 @@ fn auction_bid<S: HasStateApi>(
...
-    ensure!(amount > previous_balance, BidError::BidTooLow);
+    ensure!(amount > previous_balance, BidError::BidBelowCurrentBid);
+
+    // Calculate the difference between the previous bid and the new bid in CCD.
+    let amount_difference = amount - previous_balance;
+    // Get the current exchange rate used by the chain
+    let exchange_rates = host.exchange_rates();
+    // Convert the CCD difference to EUR
+    let euro_cent_difference = exchange_rates.convert_amount_to_euro_cent(amount_difference);
+    // Ensure that the bid is at least 100 EUR cent more than the previous bid
```
Bargsteen, line 153: Where does the 100 euro cent come from? Isn't it just that the difference is larger than `minimum_raise`?

Bargsteen (commented):
Looks good overall, nice work!

concordium-std/src/types.rs
limemloh: I have renamed it to `micro_ccd_per_euro`

concordium-std/src/test_infrastructure.rs
```diff
@@ -1287,21 +1287,32 @@ pub struct TestHost<State> {
     /// Functions that mock responses to calls.
     // This is Rc+RefCell because it needs to be cloneable. There might be another way to make the
     // MockFn cloneable, but this seemed like the easiest option.
-    mocking_fns:      Rc<RefCell<MockFnMap<State>>>,
+    mocking_fns:             Rc<RefCell<MockFnMap<State>>>,
     /// Transfers the contract has made during its execution.
-    transfers:        RefCell<Vec<(AccountAddress, Amount)>>,
+    transfers:               RefCell<Vec<(AccountAddress, Amount)>>,
```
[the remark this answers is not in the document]
Bargsteen: Oh right, of course :p 🙃

concordium-std/src/impls.rs
```diff
@@ -1893,6 +1918,79 @@ fn parse_call_response_code(code: u64) -> CallContractResult<ExternCallResponse>
...
+fn parse_query_account_balance_response_code(
+    code: u64,
+) -> Result<ExternCallResponse, QueryAccountBalanceError> {
+    if is_invoke_failure_code(code) {
+        if let 0xffff_ff02_0000_0000 = code {
+            return Err(QueryAccountBalanceError);
+        } else {
+            unsafe { crate::hint::unreachable_unchecked() }
+        };
+    }
```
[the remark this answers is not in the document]
limemloh: The main reason is that, in the node, we reuse the return value logic from the other invoke calls, which defines a failed invocation as the first 3 bytes are set.
I am not sure what is best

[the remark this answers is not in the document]
abizjak: It is better to hard fail if there is a bug either in the node or here.

concordium-std/src/test_infrastructure.rs
```diff
@@ -1326,17 +1337,18 @@ impl<State: Serial + DeserialWithState<TestStateApi> + StateClone<TestStateApi>>
         if self.missing_accounts.contains(receiver) {
             return Err(TransferError::MissingAccount);
         }
-        if amount.micro_ccd > 0 {
```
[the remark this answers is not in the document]
abizjak: Let's not change behaviour we have no strong reason for changing. Just because a change is documented does not mean people are not annoyed when there is not much of a reason for it.

concordium-std/src/impls.rs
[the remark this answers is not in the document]
limemloh: Okay I'll keep the check then.

I realized that the first 3 bytes are not set in case of failure for the response code passed to the contract. This is just the case for the response code passed in FFI in the node.
I have update the doc comments stating this and fixed the parsing. (Only the queries actually followed the doc comments, so the rest should behave the same)

concordium-std/src/test_infrastructure.rs
limemloh: I have renamed the setup functions and added a check as you suggested here, as well a check when setting self_address

concordium-std/src/test_infrastructure.rs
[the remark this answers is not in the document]
Bargsteen: Holm correctly realized that I was wrong in my comment. My statement was true back when the contract balance was an `Option<Amount>` in the context, but now it is a `RefCell<Amount>` in the host.

concordium-std/src/types.rs
limemloh: Good catch

concordium-std/src/types.rs
limemloh: I have added a check for this now

Bargsteen (approved):
Looks good, just missing very minor details.

concordium-std/src/test_infrastructure.rs
```diff
@@ -1460,9 +1478,46 @@ impl<State: Serial + DeserialWithState<TestStateApi> + StateClone<TestStateApi>>
         if amount.micro_ccd > 0 {
             *self.contract_balance.borrow_mut() -= amount;
         }
+        // Update the receiver query balance.
+        if let Some(balance) = self.query_contract_balances.borrow_mut().get_mut(to) {
```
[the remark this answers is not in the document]
limemloh: I would prefer to remove the check for amount larger than 0.

concordium-std/src/types.rs
```diff
@@ -581,6 +584,108 @@ pub struct ExternReturnValue {
...
+    /// staking.
+    pub(crate) staked: Amount,
+    /// The current amount locked in releases that resulted from transfers with
+    /// schedule. A locked amount can still be used for staking.
+    pub(crate) locked: Amount,
+}
+
+impl AccountBalance {
+    /// Construct a new account balance, ensuring that the staked and locked
+    /// amount is smaller than or equal to the total balance.
```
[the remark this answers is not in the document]
limemloh: Yes, that could be more clear.
I have added to this

concordium-std/src/test_infrastructure.rs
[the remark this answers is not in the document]
Bargsteen: That's also fine 👍

concordium-std/src/traits.rs
abizjak: It is better now.

But the best would be if we had a simple 
```
let balance_before = host.account_balance(account)?;
host.transfer(...,amount,...)?;
let balance_after = host.account_balance(account)?;
// balance_after = balance_before + amount
```

concordium-std/src/types.rs
```diff
@@ -581,6 +584,109 @@ pub struct ExternReturnValue {
...
+    /// The total balance of the account. Note that part of this balance might
+    /// be staked and/or locked in scheduled transfers.
+    pub(crate) total:  Amount,
+    /// The current staked amount of the account. This amount is used for
+    /// staking.
+    pub(crate) staked: Amount,
+    /// The current amount locked in releases that resulted from transfers with
+    /// schedule. A locked amount can still be used for staking.
+    pub(crate) locked: Amount,
+}
```
abizjak, line 599: What was the main reason for having these private?

I do understand that total >= staked, locked, but having these private leads to more boilerplate below which to me does not seem worthwhile.

Was there some other reason?

concordium-std/src/types.rs
```diff
@@ -581,6 +584,93 @@ pub struct ExternReturnValue {
...
+    pub fn available(&self) -> Amount { self.total - max(self.locked, self.staked) }
+}
+
+/// The current exchange rates.
+#[derive(Debug, Clone, Copy)]
+pub struct ExchangeRates {
+    /// Euro per NRG exchange rate.
+    euro_per_energy:    ExchangeRate,
+    /// Micro CCD per Euro exchange rate.
+    micro_ccd_per_euro: ExchangeRate,
```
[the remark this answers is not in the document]
abizjak: There is no validity imposed on these values.

concordium-std/src/test_infrastructure.rs
abizjak: Ok, we are going to move away from this sort of testing anyhow. 

This does have rather awkward semantics to explain though so I am not convinced that keeping backwards compatibility is enough of a reason. But at this time I don't htink it is worth changing the approach either.

concordium-std/src/test_infrastructure.rs
```diff
@@ -1588,6 +1648,36 @@ impl<State: Serial + DeserialWithState<TestStateApi>> TestHost<State> {
         *self.contract_balance.borrow_mut() = amount;
     }
 
+    pub fn set_self_address(&mut self, address: ContractAddress) {
```
abizjak, line 1648: Document, including the precondition for the relationship with `set_missing_contract`.

concordium-std/src/test_infrastructure.rs
```diff
@@ -1588,6 +1648,36 @@ impl<State: Serial + DeserialWithState<TestStateApi>> TestHost<State> {
...
+        if self.missing_contracts.contains(&address) {
+            fail!("The self_address is marked as a missing contract address.")
+        }
+        self.contract_address = Some(address);
+    }
+
+    pub fn setup_account_balance(
```
abizjak: Document.

concordium-std/src/test_infrastructure.rs
```diff
@@ -1588,6 +1648,36 @@ impl<State: Serial + DeserialWithState<TestStateApi>> TestHost<State> {
...
+        &mut self,
+        address: AccountAddress,
+        account_balance: AccountBalance,
+    ) {
+        self.query_account_balances.borrow_mut().insert(address, account_balance);
```
abizjak, line 1666: Also document that the behaviour when adding the same address twice is to use the second balance.

It would not be unreasonable to expecte the balances were added. SO let's avoid confusion.

concordium-std/src/test_infrastructure.rs
```diff
@@ -1588,6 +1648,36 @@ impl<State: Serial + DeserialWithState<TestStateApi>> TestHost<State> {
...
+    }
+
+    pub fn setup_contract_balance(&mut self, address: ContractAddress, balance: Amount) {
```
abizjak: Document.

Including the preconditions about the self address.

I wonder why this returns a unit instead of a Result? Did you experiment and decide this is more ergonomic?

concordium-std/src/test_infrastructure.rs
```diff
@@ -1588,6 +1648,36 @@ impl<State: Serial + DeserialWithState<TestStateApi>> TestHost<State> {
...
+
+        // Update the contract balance if the address matches.
+        if let Some(self_address) = self.contract_address {
+            if self_address == address {
+                *self.contract_balance.borrow_mut() = balance;
+            }
+        }
+    }
+
+    pub fn set_exchange_rates(&mut self, exchange_rates: ExchangeRates) {
```
abizjak, line 1679: Short documentation.

concordium-std/src/test_infrastructure.rs
```diff
@@ -1621,6 +1711,15 @@ impl<State: Serial + DeserialWithState<TestStateApi>> TestHost<State> {
     pub fn make_account_missing(&mut self, account: AccountAddress) {
         self.missing_accounts.insert(account);
     }
+
+    /// Set a contract to be missing. Any queries for the balance to this
+    /// contract will result in an [QueryContractBalanceError].
+    pub fn make_contract_missing(&mut self, contract: ContractAddress) {
```
abizjak, line 1719: Also document the precondition.

concordium-std/src/test_infrastructure.rs
```diff
@@ -1621,6 +1711,15 @@ impl<State: Serial + DeserialWithState<TestStateApi>> TestHost<State> {
     pub fn make_account_missing(&mut self, account: AccountAddress) {
         self.missing_accounts.insert(account);
     }
+
+    /// Set a contract to be missing. Any queries for the balance to this
+    /// contract will result in an [QueryContractBalanceError].
```
abizjak, line 1718:
```suggestion
    /// contract will result in an [`QueryContractBalanceError`].
```

concordium-std/src/test_infrastructure.rs
abizjak: Part of the issue will be that since this is not a macro, the line number that will be reported will probably be in this library, so it might be a bit trickier to find.

concordium-std/src/types.rs
limemloh: The reason is to future-proof the api by not leaking the struct details.
It also makes the api for different balances more alike (here I am thinking about `available()`)

abizjak: I don't think future-proofing is a good reason. We'll have a major version release anyhow if we change the data returned by the node which is the only place this is relevant.

And if we change the data then the `new` function will in any case have to be changed, so the future proofing isn't really going to be that.

concordium-std/src/test_infrastructure.rs
limemloh: I expect this to be more ergonomic. We did not have to fail here, but it seems like a better user experience and might catch some copy paste mistakes.
Going with Result would mean the user would have to deal with the error and since it is only ment for unit tests they would panic anyway to fail the test.

concordium-std/src/impls.rs
```diff
@@ -1930,6 +2028,40 @@ fn invoke_contract_construct_parameter(
...
+    Ok(Amount::deserial(&mut return_value).unwrap_abort())
+}
+
+/// Helper factoring out the common behaviour of exchange_rates for the
+/// two extern hosts below.
+fn query_exchange_rates_worker() -> ExchangeRates {
+    let response_code = unsafe { prims::invoke(INVOKE_QUERY_EXCHANGE_RATES_TAG, [].as_ptr(), 0) };
+
+    let mut response = parse_query_exchange_rates_response_code(response_code);
+    ExchangeRates::deserial(&mut response).unwrap_abort()
```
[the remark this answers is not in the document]
limemloh: I have updated the deserial implementation of ExchangeRates.
I use some unsafe code so please check the implementation as well

concordium-std/src/impls.rs
```diff
@@ -2867,17 +2867,41 @@ impl Deserial for HashKeccak256 {
...
     }
 }
 
 impl Deserial for ExchangeRates {
     fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
-        let euro_per_energy = source.get()?;
-        let micro_ccd_per_euro = source.get()?;
-        Ok(Self::new(euro_per_energy, micro_ccd_per_euro))
+        let bytes: [u8; 32] = source.read_array()?;
+        let chunks = unsafe { mem::transmute::<[u8; 32], [[u8; 8]; 4]>(bytes) };
```
Bargsteen, line 2879: Nice use-case for transmute ;)

concordium-std/src/test_infrastructure.rs
```diff
@@ -1645,6 +1645,12 @@ impl<State: Serial + DeserialWithState<TestStateApi>> TestHost<State> {
         *self.contract_balance.borrow_mut() = amount;
     }
 
+    /// Set the contract address.
+    /// This is used to redirect queryies for the balance of the contract
```
Bargsteen, line 1649:
```suggestion
    /// This is used to redirect queries for the balance of the contract
```

Bargsteen: `This is used to redirect..` makes it sound like that is the primary purpose of the method, which it isn't. 

Perhaps something like:
```
Set the contract address.

This will be the address returned by `ctx.self_address`.

This is also used to redirect queries...
```

concordium-std/src/test_infrastructure.rs
```diff
@@ -1676,6 +1710,10 @@ impl<State: Serial + DeserialWithState<TestStateApi>> TestHost<State> {
         self.query_contract_balances.borrow_mut().insert(address, balance);
     }
 
+    /// Set the current exchange rates.
+    /// This is used for resolving a query for the current exchange rates.
+    ///
+    /// Note: Setting this overwrites previous set exchange rates.
```
Bargsteen, line 1716:
```suggestion
    /// Note: Setting this overwrites the previously set exchange rates.
```

Bargsteen: The other setter methods panic on duplicate invocations with the same parameter. Why did you decide that this should just overwrite?

concordium-std/src/traits.rs
```diff
@@ -355,23 +355,37 @@ pub trait HasHost<State>: Sized {
...
     /// Get the current public balance of an account. Here public means
     /// unencrypted or unshielded. See [`AccountBalance`] for more.
+    /// This query will fail if the provided address does not exist on chain.
+    ///
+    /// Any amount received by transfers during the transaction
+    /// until the point of querying are reflected in this balance.
+    ///
+    /// ```ignore
+    /// let balance_before = host.account_balance(account)?;
+    /// host.transfer(&account, amount)?;
```
Bargsteen, line 365:
```suggestion
    /// host.invoke_transfer(&account, amount)?;
```

Bargsteen (approved):

concordium-std/src/types.rs
limemloh: Okay, the fields are now public and I removed the corresponding getters

concordium-std/src/test_infrastructure.rs
limemloh: No, none of the current ones do, so I decided to follow that.
I think it would make sense to panic on duplicate invocations with the same parameter.

concordium-std/src/traits.rs
limemloh: I have added this to both contract_balance and account_balance

concordium-std/src/impls.rs
```diff
@@ -2699,6 +2865,66 @@ impl Deserial for HashKeccak256 {
...
+        self.euro_per_energy.serial(out)?;
+        self.micro_ccd_per_euro.serial(out)?;
+        Ok(())
+    }
+}
+
+impl Deserial for ExchangeRates {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        let bytes: [u8; 32] = source.read_array()?;
+        let chunks = unsafe { mem::transmute::<[u8; 32], [[u8; 8]; 4]>(bytes) };
```
abizjak, line 2879: I think this is fine. A bit scary in a sense, but fine.

concordium-std/src/impls.rs
```diff
@@ -2699,6 +2865,66 @@ impl Deserial for HashKeccak256 {
...
+            micro_ccd_per_euro,
+        })
+    }
+}
+
+impl Serial for AccountBalance {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        self.total.serial(out)?;
+        self.staked.serial(out)?;
+        self.locked.serial(&mut *out)?;
```
abizjak, line 2912: Why is the &mut *out reborrow here needed?

abizjak (approved):

concordium-std/src/impls.rs
limemloh: Ohh... not sure why that was inserted

concordium-std/src/test_infrastructure.rs
Bargsteen: Ah, my mistake. Nvm then ;)

concordium-std/src/test_infrastructure.rs
Bargsteen: Ah, I guess I thought I was looking at the one from context.
Since it is a very rare use case, then this is probably fine then.

PR #189, opened 2022-11-10
Add voting example from hackathon

### LINKED ISSUE #184: Add voting example smart contracts from hackathon
Add the voting smart contract example

DOBEN:
## Purpose

close #184 

## Changes

Adding voting example from the hackathon. Thanks to @Bargsteen @bisgardo @tschudid @Kåre for the amazing team.

Front-end part can be found [here](https://github.com/Concordium/concordium-browser-wallet/pull/147) 

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

- [ ] I accept the above linked CLA.


examples/voting/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+        Address::Account(acc) => acc,
+        Address::Contract(_) => return Err(VotingError::ContractVoter),
+    };
+    let param: VoteParameter = ctx.parameter_cursor().get()?;
+
+    let new_vote = param.vote;
+
+    // Check that vote is in range
+    ensure!(
+        (new_vote as usize) < host.state().description.options.len(),
```
limemloh, line 100: We are loading in a potentially large vec here, just for the length.
We should probably just store the length separately in the state.

examples/voting/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    Tally {
+        result:      stats,
+        total_votes: total,
+    }
+}
+
+/// We assume that all ballots contain a valid voteoption index this should be
+/// checked by the vote function Assumption: Each account has at most one vote
+#[receive(contract = "voting", name = "getvotes", return_value = "VotingView")]
```
limemloh, line 129: Even if we accumulate the stats, this will not support a large number of VotingOptions.
Maybe a better design would be to have a "get top 10 query" and then also a getter for votes on a specific vote option.

examples/voting/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+    ContractVoter,
+}
+
+impl From<ParseError> for VotingError {
+    fn from(_: ParseError) -> Self { VotingError::ParsingFailed }
+}
+
+type VotingResult<T> = Result<T, VotingError>;
+
+#[init(contract = "voting", parameter = "InitParameter")]
```
limemloh, line 61: Doc comments everywhere would be great, just like the other examples

limemloh (commented):
I know this is not your contract, but here you have a lot of feedback.
In my opinion, this example needs to be more polished before we would have it as one of the examples.

examples/voting/src/lib.rs
DOBEN: The votes are now accumulated in the `vote` function and saved in the `tally` state variable. There is a getter function `getNumberOfVotes` that retrieves the votes from the tally for a specific `vote option index`.

I added the following comment to the documentation because there is still a limitation in the size of the `Vec<VotingOption>`.

> //! `Vec<VotingOption>` (among other variables) is an input parameter to the
> //! `init` function as well as a return_value (among other variables) of the
> //! `view` function. Since there is a limit to the parameter/return_value size,
> //! the size of the `Vec<VotingOption>` is limited.

examples/voting/src/lib.rs
DOBEN: `num_options` is now a state variable.

examples/voting/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+impl From<LogError> for VotingError {
+    fn from(le: LogError) -> Self {
+        match le {
+            LogError::Full => Self::LogFull,
+            LogError::Malformed => Self::LogMalformed,
+        }
+    }
+}
+
+type VotingResult<T> = Result<T, VotingError>;
```
Bargsteen, line 113: Add doc comment similar to what we do elsewhere. Something like "this is an alias that fixes the error type to `VotingError`"

examples/voting/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+    // Test accounts/addresses.
+    const ACC_0: AccountAddress = AccountAddress([0; 32]);
+    const ACC_1: AccountAddress = AccountAddress([1; 32]);
+    const ADDR_ACC_0: Address = Address::Account(ACC_0);
+    const ADDR_ACC_1: Address = Address::Account(ACC_1);
+
+    // Set up the test host.
+    fn make_test_host(
+        options: Vec<String>,
+        num_options: u32,
```
Bargsteen, line 278: Please remove this argument and just use `options.len()`.

examples/voting/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+    fn test_vote_with_invalid_index() {
+        // Set up the context.
+        let end_time = Timestamp::from_timestamp_millis(100);
+        let current_time = Timestamp::from_timestamp_millis(0);
+        let mut ctx = TestReceiveContext::empty();
+        ctx.set_metadata_slot_time(current_time);
+        ctx.set_sender(ADDR_ACC_0);
+
+        // Set up the parameter.
+        let vote_parameter = to_bytes(&4);
```
Bargsteen, line 327: We could check the border case here instead, as that is easier to mess up (`<` vs `<=`).
```suggestion
        let vote_parameter = to_bytes(&2);
```

examples/voting/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+        // Call the contract function.
+        let result = vote(&ctx, &mut host, &mut logger);
+        // Check the result.
+        claim!(result.is_ok(), "Results in rejection");
+        // Check the ballots (ACCOUNT_0 voted for voting option index 1 and ACCOUNT_1
+        // voted for voting option index 0).
+        let ballots = host.state().ballots.iter().map(|(a, b)| (*a, *b)).collect::<Vec<_>>();
+        claim_eq!(vec![(ACC_0, 1), (ACC_1, 0)], ballots);
+
+        // Vote yet again.
```
Bargsteen, line 400:
```suggestion
        // Vote again, using the same index as before.
```

examples/voting/src/lib.rs
```diff
@@ -0,0 +1,415 @@
+//! A voting smart contract example.
+//!
+//! # Description
+//! A contract that allows for conducting an election with several voting
+//! options. An end_time is set when the election is initialized. Only accounts
```
Bargsteen, line 5: Perhaps
```suggestion
//! options. An end time is set when the election is initialized. Only accounts
```
Or
Perhaps
```suggestion
//! options. An `end_time` is set when the election is initialized. Only accounts
```

And if you apply this change, please also do it for the `end_time`s in the rest of this module doc.

Bargsteen (approved):
Nice 😊

examples/voting/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    // Insert or replace the vote for the account.
+    host.state_mut()
+        .ballots
+        .entry(acc)
+        .and_modify(|vote| *vote = new_vote_index)
+        .or_insert(new_vote_index);
+
+    // Update the tally for the `new_vote_index` with one additional vote.
+    host.state_mut().tally.entry(new_vote_index).and_modify(|vote| *vote += 1).or_insert(1);
```
abizjak, line 204: Is there a specific reason for using this pattern as opposed to maybe a bit easier to read
```rust
*host.state_mut().tally.entry(new_vote_index).or_insert(0) += 1
```

examples/voting/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+use concordium_std::*;
+
+type VotingOption = String;
+type VoteIndex = u32;
+type VoteCount = u32;
+
+/// The parameter type for the contract function `vote`.
+/// Takes a vote_index that the account wants to vote for.
+#[derive(Deserial, SchemaType)]
+struct VoteParameter {
```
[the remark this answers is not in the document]
DOBEN: The front-end had that problem that with the given schema and passing in 0 as voting_index, you couldn't vote for the first option. There seems to be a bug with the schema decoding to handle the the edge case where the input parameter is 0. I suggest we should open a bug report for this and investigate it further.

For now, I use this struct because a struct is correctly decoded given the schema (even if it has a 0 `VoteIndex`)

examples/voting/src/lib.rs
DOBEN: A custom alias type for the `Result` type with the error type set to `VotingError`.

examples/voting/src/lib.rs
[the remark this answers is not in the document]
Bargsteen: Ok, that sounds very odd. Yes, we'll definitely have to look into that.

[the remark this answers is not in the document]
abizjak: Where is this bug with the decoding visible? In cargo-concordium?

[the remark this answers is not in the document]
DOBEN: If you use `VoteIndex` directly instead of using the struct `VoteParameter` in the function `vote`.
#[receive(
    contract = "voting",
    name = "vote",
    mutable,
    enable_logger,
    **parameter = "VoteIndex",**
    error = "VotingError"
)]

Then you get the following output when voting for the first option in the browser wallet:

Not correctly decoded (browser wallet does not show voteIndex 0):
VoteIndex0IsNotDecoded

That is how the other voteIndexes are displayed:
VoteIndex1IsCorrectlyDecoded

This is the line at the front-end:
Using `{vote}` at the front-end at line L119 with `VoteParameter` as input parameter schema works:
Using `vote` at the front-end at line L119 with `VoteIndex` as input parameter schema works not if the user selects a 0 voteIndex:


https://github.com/Concordium/concordium-browser-wallet/blob/8a4b7c85c1b73aae5eb7a979b2ce1f48caece584/examples/voting/src/Wallet.js#L119

examples/voting/src/lib.rs
abizjak: I agree with this @DOBEN

abizjak (approved):
This looks good to me, but as we discussed it would be good to get tot he bottom of why the VoteIndex cannot be used directly as a parameter.

examples/voting/src/lib.rs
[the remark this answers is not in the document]
DOBEN: Issue can be tracked here:
https://github.com/Concordium/concordium-browser-wallet/issues/189

DOBEN:
`This looks good to me, but as we discussed it would be good to get tot he bottom of why the VoteIndex cannot be used directly as a parameter.`

Issue can be tracked here:

https://github.com/Concordium/concordium-browser-wallet/issues/189

abizjak:
> `This looks good to me, but as we discussed it would be good to get tot he bottom of why the VoteIndex cannot be used directly as a parameter.`
> 
> Issue can be tracked here:
> 
> [Concordium/concordium-browser-wallet#189](https://github.com/Concordium/concordium-browser-wallet/issues/189)

Thanks.

PR #192, opened 2022-11-16
191 add admin variable to w ccd example

DOBEN:
## Purpose

Adding smart contract features to the wCCD contract

## Changes

- Add admin variable.
- Add a new function `setURL` to update the URL (only the current admin can invoke it).
- Add a new function `updateAdmin` to update the admin address (only the current admin can invoke it).
- Add logic that admin can pause/unpause the contract (the state-mutative functions `wrap`, `unwrap`, `transfer`, and `updateOperator` can only be executed when the contract is not paused)
- Add new adminEvent and new `WccdEventSchema<T, A>`
- Update the logic that only the admin can upgrade the contract and set Implementors and set URL and pause/unpause the contract.

## Checklist

      hard-to-understand areas.


examples/cis2-wccd/src/lib.rs
```diff
@@ -1,37 +1,51 @@
...
 const TOKEN_ID_WCCD: ContractTokenId = TokenIdUnit();
 
 /// The metadata url for the wCCD token.
-const TOKEN_METADATA_URL: &str = "https://some.example/token/wccd";
+/// The URL can be updated with the `setURL` function by the `admin`.
+const INITIAL_TOKEN_METADATA_URL: &str = "https://some.example/token/wccd";
+
+/// Tag for the NewAdmin event. The CIS-2 library already uses the
+/// event tags from `u8::MAX` to `u8::MAX - 4`.
+pub const NEW_ADMIN_EVENT_TAG: u8 = u8::MAX - 5;
```
abizjak, line 48: Is there a reason for not using `0`. We chose the CIS2 event tags in that way so that other events don't clash with them. And events typically have automatically derived tags, which start with 0...

examples/cis2-wccd/src/lib.rs
```diff
@@ -65,15 +79,25 @@ struct AddressState<S> {
...
+    paused:       bool,
+    /// Map specifying the `AddressState` (balance, and operators) for every
+    /// address.
     token:        StateMap<Address, AddressState<S>, S>,
     /// Map with contract addresses providing implementations of additional
     /// standards.
     implementors: StateMap<StandardIdentifierOwned, Vec<ContractAddress>, S>,
+    /// The metadata URL for the wCCD token following the specification RFC1738.
+    #[concordium(size_length = 2)]
+    url:          String,
```
abizjak, line 96: Should this be a StateBox wrapped? I guess we don't need this field apart from some view functions, and the URL is generally going to be more than 8 bytes.

examples/cis2-wccd/src/lib.rs
```diff
@@ -123,6 +154,140 @@ struct UpgradeParams {
...
+}
+
+/// A dummy `WccdEventSchema` struct to implement an event schema including both
+/// (the events specified in this contract and the events specified in the CIS-2
+/// library).
+#[allow(dead_code)]
+struct WccdEventSchema<T: IsTokenId, A: IsTokenAmount> {
+    dummy_t: T,
+    dummy_a: A,
+}
```
abizjak, line 178: Why are the two dummy fields needed?

I think the way we want to handle these extra events is to define a type

```
enum WccdEvent {
    CIS2Event(CIS2Event),
   NewAdmin(NewAdminEvent)
}
```

and then define the relevant instances for it.

While this approach you have works, it is not really how we would like to do things.

examples/cis2-wccd/src/lib.rs
```diff
@@ -123,6 +154,140 @@ struct UpgradeParams {
...
+
+/// The parameter type for the contract function `setURL`.
+#[derive(Serialize, SchemaType, Clone)]
+struct SetURLParams {
```
abizjak, line 182: I would add `#[repr(transparent)]` to all these wrappers.

examples/cis2-wccd/src/lib.rs
```diff
@@ -123,6 +154,140 @@ struct UpgradeParams {
...
+    /// The URL following the specification RFC1738.
+    #[concordium(size_length = 2)]
+    url: String,
+}
+
+/// The parameter type for the contract function `setPaused`.
+#[derive(Serialize, SchemaType)]
+struct SetPausedParams {
```
abizjak, line 190: And this one.

examples/cis2-wccd/src/lib.rs
```diff
@@ -123,6 +154,140 @@ struct UpgradeParams {
...
+    /// Contract is paused if `paused = true` and unpaused if `paused = false`.
+    paused: bool,
+}
+
+/// A NewAdminEvent introduced by this smart contract.
+#[derive(Serial, SchemaType)]
+struct NewAdminEvent {
```
abizjak, line 197: And this one.

examples/cis2-wccd/src/lib.rs
DOBEN: yes, I moved it into the StateBox. I also removed the `#[concordium(size_length = 2)]` because the length doesn't matter that much anymore.

examples/cis2-wccd/src/lib.rs
```diff
@@ -754,17 +1072,23 @@ fn contract_set_implementor<S: HasStateApi>(
...
     let params: SetImplementorsParams = ctx.parameter_cursor().get()?;
     // Update the implementors in the state
     host.state_mut().set_implementors(params.id, params.implementors);
     Ok(())
 }
 
+/// Upgrade this smart contract instance to a new module and call optionally a
+/// migration function after the upgrade.
+///
+/// It rejects if:
```
Bargsteen, line 1087: Or if the upgrade or migration fails.

examples/cis2-wccd/src/lib.rs
```diff
@@ -1187,13 +1612,257 @@ mod tests {
...
+
+        // Check that invoke failed.
+        claim_eq!(
+            result,
+            Err(ContractError::Unauthorized),
+            "Pause should fail because not the current admin tries to invoke it"
+        );
+    }
+
+    /// Test that one can NOT call state-mutative functions (wrap, unwrap,
```
Bargsteen, line 1799:
```suggestion
    /// Test that one can NOT call non-admin functions (wrap, unwrap,
```
Or whatever you decide to change it to in the previous location 😊

examples/cis2-wccd/src/lib.rs
```diff
@@ -1187,13 +1612,257 @@ mod tests {
...
+    /// transfer, updateOperator functions) when contract is paused.
```
Bargsteen, line 1800:
```suggestion
    /// transfer, updateOperator) when the contract is paused.
```

Bargsteen (approved):
Nice work! 🚀

examples/cis2-wccd/src/lib.rs
```diff
@@ -475,9 +662,114 @@ fn contract_unwrap<S: HasStateApi>(
...
+    // Log a new admin event.
+    logger.log(&WccdEvent::NewAdmin(NewAdminEvent {
+        new_admin,
+    }))?;
+
+    Ok(())
+}
+
+/// Pause/Unpause this smart contract instance by the admin. All state-mutative
+/// functions (wrap, unwrap, transfer, updateOperator) cannot be executed when
```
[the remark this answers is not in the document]
DOBEN: The query functions e.g. `blanceOf` are not pausable as well which is why I kept both adjectives:

All **non-admin state-mutative** functions (wrap, unwrap, transfer, updateOperator) cannot be executed when the contract is paused.

examples/cis2-wccd/src/lib.rs
```diff
@@ -291,6 +454,9 @@ impl<S: HasStateApi> State<S> {
         Ok(())
     }
 
+    /// Burn an amount of wCCD tokens.
+    /// Results in an error if the token id does not exist in the state or if
```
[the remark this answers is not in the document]
DOBEN: It seems to check the `tokenId` first and will therefore always throw the error "tokenIdDoesNotExist" if there is a mismatch no matter the amount.

abizjak (approved):
Looks OK to me now, except the fact that we never make use of the metadata checksum.

Is that not useful in this case?

examples/cis2-wccd/src/lib.rs
```diff
@@ -1,37 +1,50 @@
...
 //! wCCD back into CCD. The function takes the amount of tokens to unwrap, the
 //! address owning these wCCD and a receiver for the CCD. If the sender is the
-//! owner or an operator of the owner, the wCCD are burned and the amount of
-//! CCD is sent to the receiver.
-
+//! owner or an operator of the owner, the wCCD tokens are burned and the amount
+//! of CCD is sent to the receiver.
+//!
+//! There is a corresponding tutorial for this smart contract available here:
+//! https://developer.concordium.software/en/mainnet/smart-contracts/tutorials/wCCD/index.html
```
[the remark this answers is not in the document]
DOBEN: yes, I will update the documentation.

https://github.com/Concordium/concordium.github.io/issues/674

examples/cis2-wccd/src/lib.rs
```diff
@@ -123,6 +155,133 @@ struct UpgradeParams {
...
+    /// The metadata URL following the specification RFC1738.
+    url:    String,
+}
+
+/// The parameter type for the contract function `setURL`.
+#[derive(Serialize, SchemaType, Clone)]
+#[repr(transparent)]
+struct SetURLParams {
+    /// The URL following the specification RFC1738.
+    url: String,
```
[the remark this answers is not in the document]
DOBEN: When there are NFTs tracked in the smart contract, the `ipfs-hash` for each `tokenID` could be tracked with this `hash` variable in the metaData. In the case of the wCCD, I am not sure exactly which `hash` we could define/track (maybe the hash of the metaData document that we upload to a server at the given URL?).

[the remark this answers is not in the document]
abizjak: This hash that is logged as part of the TokenMetadata event is always meant to be the hash of the JSON file that is at the given URL.

So I think it always can make sense to have.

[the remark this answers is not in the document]
DOBEN: The hash can now be set as part of the setter function.

examples/cis2-wccd/Cargo.toml
```diff
@@ -6,7 +6,8 @@ edition = "2018"
 license = "MPL-2.0"
 
 [features]
-default = ["std"]
+default = ["std", "crypto-primitives"]
```
abizjak, line 9: Do we need crypto-primitives in the default? I thought we did not use that in wccd?

abizjak (approved):

limemloh (approved):
LGTM!
The simplicity compared to having a proxy is very satisfying.

examples/cis2-wccd/Cargo.toml
DOBEN: We use it in the test cases now but not in the smart contract itself that is why I included it here.

examples/cis2-wccd/src/lib.rs
```diff
@@ -1204,6 +1222,73 @@ mod tests {
...
+        let crypto_primitives = TestCryptoPrimitives::new();
+
+        // Create a new_url and a new_hash
+        let new_url = "https://some.example/token/wccd/updated".to_string();
+        let new_hash = Some(crypto_primitives.hash_sha2_256("document".as_bytes()).0);
+
+        // Set up the parameter.
+        let parameter = SetMetadataUrlParams {
+            url:  new_url.clone(),
+            hash: new_hash,
```
Bargsteen, line 1248: Nitpick, feel free to ignore:
```suggestion
        let new_hash = crypto_primitives.hash_sha2_256("document".as_bytes()).0;

        // Set up the parameter.
        let parameter = SetMetadataUrlParams {
            url:  new_url.clone(),
            hash: Some(new_hash),
```

Bargsteen (approved):
Looks good. But I think you should address Holm's two comments at least. Then we are good to go ;)

abizjak (approved):

PR #193, opened 2022-11-18
Fix tally when account changes its vote

DOBEN:
## Purpose

Tally needs to reduce the vote count when the account changes its vote.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/voting/src/lib.rs
+++ b/examples/voting/src/lib.rs
@@ -213,7 +213,7 @@ fn vote<S: HasStateApi>(
     if let Some(old_vote_index) = host.state().ballots.get(&acc) {
         let old_vote_index = *old_vote_index;
         // Update the tally for the `old_vote_index` by reducing one vote.
-        *host.state_mut().tally.entry(old_vote_index).or_insert(0) -= 1;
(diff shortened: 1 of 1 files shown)


examples/voting/src/lib.rs
```diff
@@ -210,11 +210,17 @@ fn vote<S: HasStateApi>(
     // Check that vote is in range
     ensure!(new_vote_index < host.state().num_options, VotingError::InvalidVoteIndex);
 
+    if let Some(old_vote_index) = host.state().ballots.get(&acc) {
+        let old_vote_index = *old_vote_index;
+        // Update the tally for the `old_vote_index` by reducing one vote.
+        *host.state_mut().tally.entry(old_vote_index).or_insert(0) -= 1;
```
abizjak, line 216:
```suggestion
        *host.state_mut().tally.entry(old_vote_index).or_insert(1) -= 1;
```
isn't this the more correct option? I guess the old_vote_index should always exist, so the `or_insert` won't matter?

abizjak (approved):
Good catch.

PR #195, opened 2022-11-21
Add event schema to templates

DOBEN:
## Purpose

After the new version of the cis2 library is published on `crate.io`, we can update the templates with the new event schema.

## Changes

Add event schemas to cis2-nft template example

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -187,8 +187,8 @@ jobs:
         run: |
           mv $PROJECT_NAME ${{ runner.temp }}/
           sed -i "s/root/Concordium <developers@concordium.com>/g" ${{ runner.temp }}/$PROJECT_NAME/Cargo.toml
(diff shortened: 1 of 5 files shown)


abizjak (approved):

abizjak:
Just to be clear, in order for them to use these templates they need 2.4.0 of cargo-concordium. Maybe not for building, but definitely for testing, right?

DOBEN:
```Just to be clear, in order for them to use these templates they need 2.4.0 of cargo-concordium. Maybe not for building, but definitely for testing, right?```

I added to the `ReadMe` that `cargo-concordium` version 2.4.0 is needed for the templates. I think they might be able to build it with a lower `cargo-concordium` version but they would not be able to make use of the eventSchema in that case.

Bargsteen (approved):

PR #200, opened 2022-12-12
Fix compilation of smart contracts on native OSX and Windows platforms.

### LINKED ISSUE #201: `upgrade` does not have a dummy implementation, causing linking issues.
**Bug Description**

When building smart contracts with concordium-std to native code on some platforms (OSX, Windows) there can be linking failure since the contracts expect a "upgrade" host function. Even if this is not used, those platforms expect to find the function.

For other host functions we handle this by adding dummy implementations of functions that crash when called. This works around the issue and tests can be run when compiled to native code.

For `upgrade` this was not done.

abizjak:
## Purpose

Fix a linking issue.

Fixes #201 

## Changes

Add a dummy function that implements the right interface to avoid linking failure.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std/CHANGELOG.md
+++ b/concordium-std/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased changes
 
 - Add a new primitive `get_random` for generating random numbers in Wasm code testing; `get_random` can be used in tests only, not available for smart contracts on the chain.
+- Fix a linking issue when compiling contracts to native code on Windows and OSX.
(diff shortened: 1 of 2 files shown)


Bargsteen:
Hah, I hit that issue a few minutes ago. Great ;)

DOBEN (approved):

Bargsteen (approved):
LGTM 👍

PR #207, opened 2023-01-02
Add eSealing smart contract

### LINKED ISSUE #197: Write an e-Sealing smart contract
**Task description**
Write a smart contract for the e-Sealing demo

DOBEN:
## Purpose

Add an eSealing smart contract

closes #197 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/eSealing/Cargo.toml
+++ b/examples/eSealing/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "e_sealing"
+version = "0.1.0"
+authors = ["Concordium <developers@concordium.com>"]
+edition = "2018"
+license = "MPL-2.0"
+
+[features]
+default = ["std", "crypto-primitives"]
+crypto-primitives = ["concordium-std/crypto-primitives"]
+std = ["concordium-std/std"]
+
+[dependencies]
+base58check= "*"
+concordium-std = {path = "../../concordium-std", default-features = false}
+
+[lib]
+crate-type=["cdylib", "rlib"]
+
+[profile.release]
+opt-level = "s"
+codegen-units = 1
--- a/examples/eSealing/Cargo.toml
+++ b/examples/eSealing/Cargo.toml
@@ -6,12 +6,10 @@ edition = "2018"
 license = "MPL-2.0"
 
 [features]
-default = ["std", "crypto-primitives"]
-crypto-primitives = ["concordium-std/crypto-primitives"]
+default = ["std"]
 std = ["concordium-std/std"]
 
 [dependencies]
-base58check= "*"
 concordium-std = {path = "../../concordium-std", default-features = false}
 
 [lib]
--- a/examples/eSealing/src/lib.rs
+++ b/examples/eSealing/src/lib.rs
@@ -1,0 +1,332 @@
+//! A contract implementing an eSealing service.
+use base58check::ToBase58Check;
+use concordium_std::*;
+use std::ops::Deref;
+
+/// The different errors the contract can produce.
+#[derive(Serialize, Debug, PartialEq, Eq, Reject, SchemaType)]
+enum ContractError {
+    /// Failed parsing the parameter.
+    #[from(ParseError)]
+    ParseParams,
+    /// Failed logging: Log is full.
+    LogFull,
+    /// Failed logging: Log is malformed.
+    LogMalformed,
+    /// Only account.
+    OnlyAccount,
+}
+
+/// Mapping the logging errors to ContractError.
+impl From<LogError> for ContractError {
+    fn from(le: LogError) -> Self {
+        match le {
+            LogError::Full => Self::LogFull,
+            LogError::Malformed => Self::LogMalformed,
+        }
+    }
+}
+
+/// The state tracked for each address.
+#[derive(Serial, DeserialWithState, Deletable, StateClone)]
+#[concordium(state_parameter = "S")]
+struct AddressState<S> {
+    /// The number of tokens owned by this address.
+    timestamps: StateSet<Timestamp, S>,
+    /// The address which are currently enabled as operators for this token and
+    /// this address.
+    witnesses:  StateSet<AccountAddress, S>,
+}
+
+#[derive(Serial, DeserialWithState, StateClone)]
+#[concordium(state_parameter = "S")]
+struct State<S> {
+    files: StateMap<String, AddressState<S>, S>,
+}
+
+impl<S: HasStateApi> State<S> {
+    fn new(state_builder: &mut StateBuilder<S>) -> Self {
+        State {
+            files: state_builder.new_map(),
+        }
+    }
+
+    fn find_file(&self, file_hash: &str) -> bool {
+        let record = self.files.get(&file_hash.to_string());
+
+        match record {
+            Some(_entry) => {
+                // file exists
+                true
+            }
+            None => false,
+        }
+    }
+
+    fn get_witnesses(&self, file_hash: &str) -> ReceiveResult<Vec<AccountAddress>> {
+        let record = self.files.get(&file_hash.to_string());
+
+        match record {
+            Some(entry) => {
+                // file exists
+                Result::Ok(entry.witnesses.iter().map(|x| *x.deref()).collect())
+            }
+            None => Result::Ok([].to_vec()),
+        }
+    }
+
+    fn add_witness(&mut self, file_hash: &str, witness: AccountAddress) {
+        self.files.get_mut(&file_hash.to_string()).unwrap().witnesses.insert(witness);
+    }
+
+    fn add_file(&mut self, file_hash: &str, state_builder: &mut StateBuilder<S>) {
+        self.files.insert(file_hash.to_string(), AddressState {
+            timestamps: state_builder.new_set(),
+            witnesses:  state_builder.new_set(),
+        });
+    }
+}
+
+#[derive(Debug, Serial, SchemaType)]
+enum Event {
+    WitnessAdded(WitnessAddedEvent),
+    FileAdded(FileAddedEvent),
+}
+
+/// The event is logged when a new witness is added.
+#[derive(Debug, Serialize, SchemaType)]
+pub struct WitnessAddedEvent {
+    ///
+    file_hash: String,
+    ///
+    witness:   String,
+}
+
+/// The event is logged when a new file is added.
+#[derive(Debug, Serialize, SchemaType)]
+pub struct FileAddedEvent {
+    ///
+    file_hash: String,
+}
+
+#[init(contract = "eSealing", event = "Event")]
+fn contract_init<S: HasStateApi>(
+    _ctx: &impl HasInitContext,
+    state_builder: &mut StateBuilder<S>,
+) -> InitResult<State<S>> {
+    Ok(State::new(state_builder))
+}
+
+#[receive(
+    contract = "eSealing",
+    name = "registerFile",
+    parameter = "String",
+    error = "ContractError",
+    mutable,
+    enable_logger
+)]
+fn register_file<S: HasStateApi>(
+    ctx: &impl HasReceiveContext,
+    host: &mut impl HasHost<State<S>, StateApiType = S>,
+    logger: &mut impl HasLogger,
+) -> Result<(), ContractError> {
+    // Ensure that only accounts can register a file
+    let sender_address = match ctx.sender() {
+        Address::Contract(_) => bail!(ContractError::OnlyAccount),
+        Address::Account(account_address) => account_address,
+    };
+
+    let file_hash: String = ctx.parameter_cursor().get()?;
+
+    let witness_string = sender_address.0.to_base58check(1);
+
+    let (state, builder) = host.state_and_builder();
+    let record = state.find_file(&file_hash);
+
+    match record {
+        true => {
+            // file exists
+            state.add_witness(&file_hash, sender_address);
+            logger.log(&Event::WitnessAdded(WitnessAddedEvent {
+                file_hash,
+                witness: witness_string,
+            }))?;
+        }
+        false => {
+            // file does not exist. Add file, Add Witness
+            state.add_file(&file_hash, builder);
+            logger.log(&Event::FileAdded(FileAddedEvent {
+                file_hash: file_hash.clone(),
+            }))?;
+
+            state.add_witness(&file_hash, sender_address);
+            logger.log(&Event::WitnessAdded(WitnessAddedEvent {
+                file_hash,
+                witness: witness_string,
+            }))?;
+        }
+    }
+
+    Result::Ok(())
+}
+
+#[receive(contract = "eSealing", name = "getfile", parameter = "String")]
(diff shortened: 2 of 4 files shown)


examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+//! by the same account would not give us any additional information.
+//!
+//! But: A second account can register a file hash that was already previously
+//! registered. This second account will be added as another witness together
+//! with the timestamp when the second account called the `register_file`
+//! function. We know that the second account is most likely now also in
+//! possession of the file if it has used this eSealing service via the
+//! front-end. But this is not guaranteed since the second account could
+//! register a random file hash it found in the events by interacting with the
+//! smart contract directly (without using the front-end).
```
abizjak, line 19: Seeing that you already point out the issue with the second registration. What is the use-case for it?

We cannot use it for a guarantee. So I'm not sure why we have the added complexity.

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+}
+
+/// The state tracked for each file.
+#[derive(Serial, DeserialWithState, Deletable, StateClone)]
+#[concordium(state_parameter = "S")]
+struct FileState<S> {
+    /// The timestamps when this file hash was recorded.
+    timestamps:    Vec<Timestamp>,
+    /// The witnesses (sender_accounts) that recorded this file hash.
+    witnesses:     Vec<AccountAddress>,
```
abizjak, line 56: I am not sure why we need this vector. Is it to order by increasing timestamp?

Is the length of timestamps always the same as witnesses?

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    /// Check if a file exists.
+    fn file_exists(&self, file_hash: &str) -> bool {
+        let record = self.files.get(&file_hash.to_string());
+
+        match record {
+            Some(_entry) => true,
+            None => false,
+        }
```
abizjak, line 86:
```suggestion
        record.is_some()
```

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+    /// without iterating through the above array which
+    /// saves execution time in the `register_file` function.
+    witnesses_set: StateSet<AccountAddress, S>,
+}
+
+/// The contract state.
+#[derive(Serial, DeserialWithState, StateClone)]
+#[concordium(state_parameter = "S")]
+struct State<S> {
+    files: StateMap<String, FileState<S>, S>,
```
abizjak, line 68: I am not sure why keys are strings and not `[u8; 32]`.

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+    /// Creates a new state with no files recorded.
+    fn new(state_builder: &mut StateBuilder<S>) -> Self {
+        State {
+            files: state_builder.new_map(),
+        }
+    }
+
+    /// Check if a file exists.
+    fn file_exists(&self, file_hash: &str) -> bool {
+        let record = self.files.get(&file_hash.to_string());
```
abizjak, line 81: We should not have to convert to string and then take a reference. We can take `&String` as a parameter if we have to keep strings.

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+
+        match record {
+            Some(_entry) => true,
+            None => false,
+        }
+    }
+
+    /// Check if a witness exists.
+    fn witness_exists(&self, file_hash: &str, witness: AccountAddress) -> bool {
+        let records = self.files.get(&file_hash.to_string());
```
abizjak, line 91: Same here.

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+            }
+            None => Result::Ok(([].to_vec(), [].to_vec())),
+        }
+    }
+
+    /// Add a new records to a specific file hash.
+    fn add_record(&mut self, file_hash: &str, witness: AccountAddress, time: Timestamp) {
+        self.files.get_mut(&file_hash.to_string()).unwrap().timestamps.push(time);
+        self.files.get_mut(&file_hash.to_string()).unwrap().witnesses.push(witness);
+        self.files.get_mut(&file_hash.to_string()).unwrap().witnesses_set.insert(witness);
```
abizjak, line 119: Doing it like this means you are constructing 3 strings and looking up the same value in the map 3 times. It would be better to do this once.
```rust
let val = self.files.get_mut(file_hash);
if let Some(val) = val {
    val.timestamps.push(time);
    ...
}
```

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+enum Event {
+    WitnessAdded(WitnessAddedEvent),
+    FileAdded(FileAddedEvent),
+}
+
+/// The WitnessAddedEvent is logged when a new witness is added.
+#[derive(Debug, Serialize, SchemaType)]
+pub struct WitnessAddedEvent {
+    /// Hash of the file to be registered by the witness (sender_account).
+    file_hash: String,
```
abizjak, line 143: I would strongly prefer using `[u8; 32]` instead of Strings. The base16 encoding is done in clients.

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+    host: &impl HasHost<State<S>, StateApiType = S>,
+) -> ReceiveResult<(Vec<AccountAddress>, Vec<Timestamp>)> {
+    let file_hash: String = ctx.parameter_cursor().get()?;
+    let state = host.state();
+    let exists = state.file_exists(&file_hash);
+
+    match exists {
+        true => state.get_records(&file_hash),
+        false => Result::Ok(([].to_vec(), [].to_vec())),
+    }
```
abizjak, line 254:
```suggestion
    state.get_records(&file_hash)
```
It looks to me like `get_records` already returns an empty list if the hash does not exist.

abizjak (commented):
I'm not a fan of the added complexity when recording a list of timestamps.

As you yourself point out it is not really a guarantee of anything to record the same hash with an additional timestamp. Is there a use-case specifically for this?

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+
+    Result::Ok(())
+}
+
+/// Get all records about a specific file hash.
+///
+/// It rejects if:
+/// - It fails to parse the parameter.
+#[receive(
```
limemloh, line 236: This does not scale that well.
I am not sure of the requirements for this contract, but since anyone can register as a witness, the list of all witnesses does not seem that interesting.
Would a query for whether a given account is a witness of a given file suffice? It could return the timestamp as well when there is a witness.

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+//!
+//! Only accounts can register a file hash. During the registration, the
+//! timestamp is recorded together with the witness (sender_account) that
+//! registered the file hash.
+//!
+//! Every account can only register each file hash
+//! once. This is because we want to record the time when the witness
+//! (sender_account) first had access to that file. Re-registering a file hash
+//! by the same account would not give us any additional information.
+//!
```
limemloh, line 11: Maybe add a bit on the use case.
To my understanding, we want an account to be able to claim to have seen some document (hash) at some point in time, such that others cannot claim the document was created/modified at a later time.

DOBEN: We plan to have these two basic functionalities supported for the beginning:
- Upload a file from the computer => register its file hash in the smart contract via the Concordium Browser wallet
Similar to:
http://filesealing.newtrust.co.s3-website-us-east-1.amazonaws.com/
- Upload a file from the computer => retrieve the time-stamp and witness (sender_account) that registered the file hash
Similar to:
https://timestamp.northstake.dk/

Bargsteen (commented):
I had a look at this, but I didn't find more issues than currently found. I'll check back for the second review ;)

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+//!   witness (sender_account) was in possession of that file at that time.
+//!
+//! Only accounts can register a file hash. During the registration, the
+//! timestamp is recorded together with the witness (sender_account) that
+//! registered the file hash. Each file hash can only be registered once.
+//! This is because we want to record the first time when the witness
+//! (sender_account) had access to that file. Re-registering a file hash (by a
+//! different witness) would not prove that the second witness is also in
+//! possession of that file or that it just observed the file hash during the
+//! initial registration hence a file hash can only be registered once.
```
abizjak, line 18: The part "or that it just observed" I don't understand. Is there something off in the sentence?

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+        &self,
+        file_hash: FileHash,
+    ) -> ReceiveResult<(Timestamp, AccountAddress)> {
+        let record = self.files.get(&file_hash);
+
+        match record {
+            // File exists.
+            Some(entry) => Result::Ok((entry.timestamp, entry.witness)),
+            // File does not exist.
+            None => Result::Ok((Timestamp::from_timestamp_millis(0), AccountAddress([0u8; 32]))),
```
abizjak, line 92: I'm not sure why we are not returning `None` in this case. That is our usual approach at least. Is there a reason why we can't return an option here instead? These "default values" are not the Rust way :)

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+    contract = "eSealing",
+    name = "getFile",
+    parameter = "FileHash",
+    error = "ContractError",
+    return_value = "(Timestamp, AccountAddress)"
+)]
+fn get_file<S: HasStateApi>(
+    ctx: &impl HasReceiveContext,
+    host: &impl HasHost<State<S>, StateApiType = S>,
+) -> ReceiveResult<(Timestamp, AccountAddress)> {
```
abizjak, line 194: I recommend changing this to `Option<(Timestamp, AccountAddress)>` instead.

DOBEN: I changed it to `Option<FileState>`

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+        // Set up the context.
+        let ctx = TestInitContext::empty();
+        let mut builder = TestStateBuilder::new();
+
+        // Initialize the contract.
+        let init_result = contract_init(&ctx, &mut builder);
+
+        // Check the state.
+        let state = init_result.expect_report("Contract Initialization failed");
+        claim_eq!(state.files.iter().count(), 0, "No files present after initialization");
```
Bargsteen, line 209:
```suggestion
        claim!(state.files.is_empty(), "No files present after initialization");
```

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+#[derive(Serial, Deserial, Clone, Copy, SchemaType)]
+struct FileState {
+    /// The timestamp when this file hash was registered.
+    timestamp: Timestamp,
+    /// The witness (sender_account) that registered this file hash.
+    witness:   AccountAddress,
+}
+
+/// FileHash type is a fixed-length byte array.
+type FileHash = [u8; 32];
```
abizjak, line 57: This does allow you to specify the hash as a hex string in JSON, right?

abizjak (approved):

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+        let result = register_file(&ctx, &mut host, &mut logger);
+        claim!(result.is_ok(), "results in rejection");
+
+        // Check the event.
+        let event = Event::Registration(RegistrationEvent {
+            file_hash: FILE_HASH,
+            witness:   ACCOUNT_0,
+            timestamp: Timestamp::from_timestamp_millis(TIME),
+        });
+        claim!(logger.logs.contains(&to_bytes(&event)), "should contain event");
```
Bargsteen, line 240:
```suggestion
        claim!(logger.logs.contains(&to_bytes(&event)), "should contain event");
        claim!(host.state().file_exists(&FILE_HASH), "state should contain file");
```

Bargsteen (approved):
Nice work! I just had few minor findings 😊

examples/eSealing/src/lib.rs
limemloh: I don't think it does. It needs the `ByteArray(32)` schema type for this and I think this will have the more generic `Array(32, U8)`.

examples/eSealing/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "e_sealing"
+version = "0.1.0"
+authors = ["Concordium <developers@concordium.com>"]
+edition = "2018"
+license = "MPL-2.0"
+
+[features]
+default = ["std"]
+std = ["concordium-std/std"]
```
limemloh, line 10: these feature flags only makes sense if the contract supports no_std, which is seems not to do.
So maybe remove them and allow the default features of `concordium-std`

examples/eSealing/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    /// Get recorded FileState (timestamp and witness) from a specific file
+    /// hash.
+    fn get_file_state(&self, file_hash: FileHash) -> ReceiveResult<Option<FileState>> {
+        Ok(self.files.get(&file_hash).map(|v| *v))
+    }
+
+    /// Add a new file hash.
+    fn add_file(&mut self, file_hash: FileHash, timestamp: Timestamp, witness: AccountAddress) {
+        self.files.insert(file_hash, FileState {
```
limemloh, line 89: Minor improvement:
[`StateMap.insert`](https://docs.rs/concordium-std/latest/concordium_std/struct.StateMap.html#method.insert) returns whether the map already contained the key. If you make `add_file` return an error in the case of file already existing, you don't need to first look it up in the `registerFile` receive function.
Might be less readable, so I am okay without the improvement.

limemloh (approved):

examples/eSealing/src/lib.rs
DOBEN: The front-end/concordium-client require as input parameter a `ByteArray(32)` in the format:
`[1,2, ..,32]`

a hex string does not work

examples/eSealing/Cargo.toml
DOBEN: I added `#![cfg_attr(not(feature = "std"), no_std)]` into the `lib.rs` file.

examples/eSealing/src/lib.rs
abizjak: Well could we make it work then please? That will be a better experience. I know this is merged now, but can we please can you fix this?

PR #221, opened 2023-01-27
Make wee_alloc a feature

### LINKED ISSUE #220: Find alternative to `wee_alloc`
**Task description**

`wee_alloc` is unmaintained now but we depend on it in our code. This causes security warnings in our code and does not look good. Explore the effects of removing it or replacing it with similarly maintained allocators. Suggested was to replace it with `dlmalloc`.

`wee_alloc` was introduced to optimize/reduce the size of our smart contract modules at a cost of increasing the execution cost.

Removing `wee_alloc` will likely result in an increase in the smart contract module size but could result in a decrease in the execution cost.

Task: Use some of the smart contract examples and compare the module size with/without `wee_alloc` (and replace it with `dlmalloc`). Also deploy them on chain and compare execution cost before and after.

DOBEN:
## Purpose

closes #220 

Tests were conducted to compare the performance (module size, and execution cost) on several of our smart contract examples to compare the `wee_alloc` and the `dlmalloc` allocators. 
The `wee_alloc` had a better performance than the `dlmalloc` in terms of module size reduction. 
Because the execution cost also takes into account the module size on the Concordium chain, the `wee_alloc` also performed better in the execution cost comparison. 

The decision was made to keep the `wee_alloc` but make it a feature to give people the option to use their own custom allocator.

## Changes
The "wee-alloc" feature should be enabled by default in your smart contract project.
```
[features]
default = ["std", "crypto-primitives", "wee-alloc"]
wee-alloc = ["concordium-std/wee-alloc"]
```

then you can use wee-alloc in the no_std version as follows:
```
cargo +nightly concordium build -- --no-default-features --features wee-alloc
```

Note: The linter will fail on the `cis2-nft` example at the moment. We need to publish a new `concordium-std` library version. Then we can update the `cis2-nft` example and its associated templates to resolve the linter error.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std/Cargo.toml
+++ b/concordium-std/Cargo.toml
@@ -11,7 +11,9 @@ readme = "./README.md"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [dependencies]
-dlmalloc = { version = "0.2.4", features = ["global"]}
+# Reduce code size, at the cost of performance in allocation heavy-code.
+wee_alloc= { version = "0.4.5", optional = true}
+dlmalloc = { version = "0.2.4", features = ["global"], optional = true}
 sha2 = { version = "0.10", optional = true }
 sha3 = { version = "0.10", optional = true }
 secp256k1 = { version = "0.22", optional = true }
@@ -30,7 +32,8 @@ default-features = false
 
 [features]
 default = ["std"]
-
+wee-alloc = ["wee_alloc"]
+dl-malloc = ["dlmalloc"]
 std = ["concordium-contracts-common/std"]
 wasm-test = ["concordium-std-derive/wasm-test"]
 build-schema = ["concordium-std-derive/build-schema"]
--- a/concordium-std/Cargo.toml
+++ b/concordium-std/Cargo.toml
@@ -11,8 +11,7 @@ readme = "./README.md"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [dependencies]
-# Reduce code size, at the cost of performance in allocation heavy-code.
-wee_alloc="0.4.5"
+dlmalloc = { version = "0.2.4", features = ["global"]}
 sha2 = { version = "0.10", optional = true }
 sha3 = { version = "0.10", optional = true }
 secp256k1 = { version = "0.22", optional = true }
--- a/concordium-std/Cargo.toml
+++ b/concordium-std/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "concordium-std"
-version = "5.2.0"
+version = "6.0.0"
 authors = ["Concordium <developers@concordium.com>"]
 edition = "2018"
 license = "MPL-2.0"
@@ -26,7 +26,7 @@ version = "5.1"
 
 [dependencies.concordium-contracts-common]
 path = "../concordium-contracts-common/concordium-contracts-common"
-version = "5.1"
+version = "5.2"
 default-features = false
 
 [features]
--- a/concordium-std/Cargo.toml
+++ b/concordium-std/Cargo.toml
@@ -31,7 +31,6 @@ default-features = false
 
 [features]
 default = ["std"]
-use-wee-alloc = ["wee_alloc"]
 std = ["concordium-contracts-common/std"]
 wasm-test = ["concordium-std-derive/wasm-test"]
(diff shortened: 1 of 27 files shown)


concordium-std/src/lib.rs
```diff
@@ -273,9 +277,19 @@ pub use impls::*;
 pub use traits::*;
 pub use types::*;
 
+#[cfg(feature = "dl-malloc")]
+extern crate dlmalloc;
+// Use `globalDlmalloc` as the global allocator.
+#[allow(dead_code)]
+#[cfg(feature = "dl-malloc")]
+#[cfg_attr(not(feature = "std"), global_allocator)]
+static ALLOC: dlmalloc::GlobalDlmalloc = dlmalloc::GlobalDlmalloc;
```
abizjak, line 286: The idea (as I thought it) was to just have the `wee-alloc` feature. 

And we document how users can use their own allocator if they want. But we would not be supporting multiple allocators in concordium-std.

concordium-std/src/lib.rs
```diff
@@ -273,9 +277,19 @@ pub use impls::*;
...
+
+#[cfg(feature = "wee-alloc")]
 extern crate wee_alloc;
 // Use `wee_alloc` as the global allocator to reduce code size.
 #[global_allocator]
+#[cfg(feature = "wee-alloc")]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
```
abizjak, line 293: I think this needs to be 
```suggestion
#[cfg_attr(feature = "wee-alloc", global_allocator)]
static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
```

examples/cis2-wccd/Cargo.toml
```diff
@@ -6,9 +6,11 @@ edition = "2018"
 license = "MPL-2.0"
 
 [features]
-default = ["std", "crypto-primitives"]
+default = ["std", "crypto-primitives", "wee-alloc"]
 crypto-primitives = ["concordium-std/crypto-primitives"]
 std = ["concordium-std/std", "concordium-cis2/std"]
+wee-alloc = ["concordium-std/wee-alloc"]
+dl-malloc = ["concordium-std/dl-malloc"]
```
abizjak, line 13: I'd recommend not having this feature. Users can do this themselves.

concordium-std/Cargo.toml
```diff
@@ -31,7 +32,8 @@ default-features = false
 
 [features]
 default = ["std"]
-
+wee-alloc = ["wee_alloc"]
+dl-malloc = ["dlmalloc"]
```
abizjak, line 36: I'd recommend not having this feature. Users can do this themselves.

DOBEN:
Comments/changes look good to me. Thanks, @abizjak

Bargsteen (approved):
LGTM 👍

concordium-std/src/lib.rs
```diff
@@ -41,8 +30,9 @@
 //! This library has the following features:
 //! [`std`](#std-build-with-the-rust-standard-library),
 //! [`build-schema`](#build-schema-build-for-generating-a-module-schema),
-//! [`wasm-test`](#wasm-test-build-for-testing-in-wasm), and
-//! [`crypto-primitives`][crypto-feature].
+//! [`wasm-test`](#wasm-test-build-for-testing-in-wasm),
+//! [`crypto-primitives`][crypto-feature], and
+//! [`use-wee-alloc`](#use-a-custom-allocator)
```
limemloh, line 35: I would prefer if we followed the naming convention set by the Rust library team and call it `wee-alloc`.
They explicitly mention not to use `use-` as a prefix for an optional dependency https://rust-lang.github.io/api-guidelines/naming.html#feature-names-are-free-of-placeholder-words-c-feature

abizjak: Well "Do not include words in the name of a [Cargo feature](http://doc.crates.io/manifest.html#the-features-section) that convey zero meaning". 

I don't think `use` conveys zero meaning :P

abizjak: Admittedly `set` would be even more apt.

abizjak: The feature is not for using the optional dependency, it is for setting a global allocator.

PR #223, opened 2023-02-06
Add cis3-nft-sponsored-txs

### LINKED ISSUE #222: Create a Cis3-NFT contract that works with gasless transactions.
**Task description**

The contract will need a registry for mapping `publicKey` to the `accountAddress` before a gasless transaction can be processed in the smart contract.

DOBEN:
## Purpose

closes #222

Next step: Move parts of this example to a CIS3 library.

## Changes

New functions:
- `registerPublicKey` (only the owner or account itself can register a public key)
- `publicKeyOf` (query the public keys of a list of given accounts)
- `permit` (sponsored transaction counterpart to the regular `updateOperator/transfer` function)

## Checklist

      hard-to-understand areas.


examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2083 @@
...
+    }
+    Ok(())
+}
+
+type TransferParameter = TransferParams<ContractTokenId, ContractTokenAmount>;
+
+/// Internal `transfer/permitTransfer` helper function. Invokes the `transfer`
+/// function of the state. Logs a `Transfer` event and invokes a receive hook
+/// function.
+fn _transfer<S: HasStateApi>(
```
Bargsteen, line 747: We don't really use the `_` prefix for internal functions. It is internal since it doesn't have a `receive` attribute, or a `pub` keyword in front 😊

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2083 @@
...
+/// - A different nonce is expected.
+/// - The signature was intended for a different contract.
+/// - The signature was intended for a different `entry_point`.
+/// - No public key was registered for the given account.
+/// - The signature can not be validated.
+/// - Any of the transfers fail to be executed, which could be if:
+///     - The `token_id` does not exist.
+///     - The token is not owned by the `from`.
+/// - Fails to log event.
+/// - Any of the receive hook function calls rejects.
```
Bargsteen, line 856:
```suggestion
/// - Any of the receive hook function calls reject.
```

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2083 @@
...
+            None => bail!(CustomContractError::NoPublicKey.into()),
+            Some(public_key) => {
+                // Check signature.
+                let is_valid = crypto_primitives.verify_ed25519_signature(
+                    public_key,
+                    param.signature,
+                    &message_hash,
+                );
+
+                match is_valid {
```
Bargsteen, line 929: `if is_valid` is probably more idiomatic 😊

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2083 @@
...
+    host: &mut impl HasHost<State<S>, StateApiType = S>,
+    logger: &mut impl HasLogger,
+    crypto_primitives: &impl HasCryptoPrimitives,
+) -> ContractResult<()> {
+    // Parse the parameter.
+    let PermitUpdateOperatorParams(params) = ctx.parameter_cursor().get()?;
+
+    for param in params {
+        // Get the next consecutive nonce.
+        let nonce = host
```
Bargsteen, line 1050: If `params.len > 1` then it is cheaper to keep the `nonce` in a variable outside the loop and only write it to the state once the for loop has finished.

Bargsteen (approved):
Nice work! I just had a few minor findings.

Regarding the batched registration of public keys: I think we should add that. It will just be a few extra bytes (dependent on the `size_length` we choose), but will reduce the number of transactions if someone wants to batch these.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2083 @@
...
+//! that pay fiat money off-chain to the third-party service to request the
+//! submission of sponsored transactions on their behalf.
+//!
+//! The Concordium team is working on exposing the public key within the smart
+//! contract code and this feature is planned to be included in the next
+//! protocol update in 6-12 months. After that protocol update, the
+//! `public_key_registry` will not be necessary anymore and except for that the
+//! third-party service can refuse to submit a transaction on behalf of the
+//! user, no trust assumption needs to be given to the third-party service
+//! anymore.
```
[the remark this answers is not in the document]
DOBEN: I removed that sentence about the trust assumption for the third-party service.

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: Because of the *any/anyone* (singular), I feel it should be `rejects`.

**Any** of the receive hook function calls rejects

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: The `param.owner` can be different in every loop, so also the nonce can be different in every loop and that is why the nonce has to be inside of the loop and updated for every iteration in the for loop.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2083 @@
...
+#[derive(Serialize, SchemaType)]
+struct ViewState {
+    state:      Vec<(Address, ViewAddressState)>,
+    all_tokens: Vec<ContractTokenId>,
+}
+
+/// View function that returns the entire contents of the state. Meant for
+/// testing.
+#[receive(contract = "cis3_nft", name = "view", return_value = "ViewState")]
+fn contract_view<S: HasStateApi>(
```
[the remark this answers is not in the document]
DOBEN: I added it and also a test for this view function.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2083 @@
...
+//! (often paying some fees in fiat money). The third-party service provider
+//! submits the transaction on behalf of the user and pays the transaction fee
+//! to execute the transaction on-chain.
+//!
+//! Concordium smart contracts currently have no way to query the corresponding
+//! public key/keys of an account within the smart contract code. For
+//! the time being this smart contract uses a `public_key_registry` that
+//! allows only the owner of the contract instance (or the account iteself) to
+//! register a public key to a given account. Once an account has a public key
+//! registered it can not be updated anymore. The users/clients are advised to
```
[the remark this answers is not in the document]
DOBEN: Updated the sentence to:

```
Once an account has a public key registered, the mapping between the public key and the account can not be updated anymore. 
```

examples/cis3-nft-sponsored-txs/src/lib.rs
Bargsteen: Ah, right. That was a blunder on my part ;)

examples/cis3-nft-sponsored-txs/src/lib.rs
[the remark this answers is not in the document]
Bargsteen: Great 😊

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -609,18 +613,35 @@ fn contract_view<S: HasStateApi>(
...
         all_tokens,
+        all_nonces,
+        all_public_keys,
     })
 }
 
-/// Register a public key to a given account. The corresponding private
+/// The parameter type for the contract function `registerPublicKey`.
+#[derive(Debug, Serialize, SchemaType)]
+pub struct RegisterPublicKeyParams(#[concordium(size_length = 2)] pub Vec<RegisterPublicKeyParam>);
```
Bargsteen, line 639: If you use a set (fx `BTreeSet`), then you can avoid duplicates automatically.
We also support sets in the schema, and our tools (at least concordium-client) will check whether it contains any duplicates before making a transaction.

Bargsteen (approved):
LGTM if you use a set ;)

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2228 @@
...
+        );
+
+        transfer(token_id, amount, from, to, data, host, logger)?;
+    }
+    Ok(())
+}
+
+/// The parameter type for the contract function `permitTransfer`.
+#[derive(Debug, Serialize, SchemaType)]
+pub struct PermitTransferParams(#[concordium(size_length = 2)] pub Vec<PermitTransferParam>);
```
limemloh, line 860: I would limit the entry point to only take one signed message instead of a list.

Also, the signed message should contain a list of transfers, like in CIS-2.
Currently, you cannot sign multiple transfer in one message, and signing multiple messages allows the account sponsoring the transaction to split it up, which might not be ideal.

limemloh (changes requested):

abizjak (changes requested):
I haven't gone through all the details, because there seemed to have been a misunderstanding.

The idea was to have a standard, CIS3, specifying how to separate the sender from the person intending to do the action. That would be specifying essentially a format like the PermitTransferParams which would have

- signature
- signer
- nonce
- timestamp
- parameters of UpdatePayload like a normal transaction has (minus the amount I suppose since that should be 0 always, otherwise there probably isn't much point)

Then this example would be an extension of a standard token that would add this extra endpoint and dispatch to the normal endpoint of the contract.

Does that make sense?

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: I kept it for now since it is the same pattern as the `balanceOf` and `operatorOf` functions to keep it consistent.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2228 @@
...
+    pub account: AccountAddress,
+}
+
+/// Get the public keys of accounts.
+///
+/// It rejects if:
+/// - It fails to parse the parameter.
+#[receive(
+    contract = "cis3_nft",
+    name = "publicKeyOf",
```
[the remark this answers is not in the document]
DOBEN: yes, I extended the getter with the current nonce.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2228 @@
...
+/// The message struct that is signed in the `permitTransfer` function.
+/// Takes all input parameters to the regular `transfer` function,
+/// the `contract_address`/`entry_point` that the signature is intended for, and
+/// a `nonce` to prevent replay attacks.
+#[derive(SchemaType, Serialize)]
+struct PermitTransferMessage {
+    /// The `from` parameter from the regular `transfer` function. Because only
+    /// an `accountAddress` can generate a signature, the `from` has
+    /// to be an `accountAddress`.
+    from:             AccountAddress,
```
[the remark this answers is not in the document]
DOBEN: Support transfers from contracts would mean that the smart contract needs to `sign` a transaction but we don't have a standard yet for that. For now, it seems that the `permit` function can be used by only `AccountAddresses` since they have public keys and can sign.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2076 @@
...
+    /// The entry_point that the signature is intended for.
+    entry_point:      OwnedEntrypointName,
+    /// A nonce to prevent replay attacks.
+    nonce:            u64,
+    /// A timestamp to make signatures expire.
+    timestamp:        Timestamp,
+    /// The payload that should be forwarded to either the `transfer` or the
+    /// `updateOperator` function.
+    payload:          PermitPayload,
+}
```
abizjak, line 279: So I think there is a bit of an annoying issue here.

The serialization of `PermitPayload` will have a one byte tag, and then the transfer params or update operator params.

This means that in the CIS3 standard we would have to specify that the "payload" should have this extra byte tag or something like it instead of just the payload of the entrypoint. That's not necessarily bad, but needs to be clarified.

What exactly do we want here? Do we want to allow the permit to add extra data in addition to the parameters of the entrypoints?

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2076 @@
...
+
+/// The parameter type for the contract function `permit`.
+/// Takes a signature, the signer, and the message that was signed.
+#[derive(Serialize, SchemaType)]
+pub struct PermitParam {
+    /// Signature.
+    signature: SignatureEd25519,
+    /// Signer that signed the above signature.
+    signer:    AccountAddress,
```
abizjak, line 288: I think we should support the same type of signature, which is a multisig like we do for transactions.

That will make it simpler to later use accounts. That signature is essnetially just a map
```
BTreeMap<u8, BTreeMap<u8, SignatureEd25519>>
```

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2076 @@
...
+
+    // Update the nonce.
+    host.state_mut().public_key_registry.entry(param.signer).and_modify(|(_, a)| *a += 1u64);
+
+    // Get the public key and the current nonce.
+    let entry = host
+        .state_mut()
+        .public_key_registry
+        .get(&param.signer)
+        .ok_or(CustomContractError::NoPublicKey);
```
abizjak, line 817:
```suggestion
        .ok_or(CustomContractError::NoPublicKey)?;
```

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2076 @@
...
+    let public_key = entry.as_ref().unwrap().0;
+    let nonce = entry.as_ref().unwrap().1;
```
abizjak, line 819:
```suggestion
    let public_key = entry.as_ref().0;
    let nonce = entry.as_ref().1;
```

DOBEN: Changed it to 

```
    let public_key = entry.0;
    let nonce = entry.1;
```

examples/cis3-nft-sponsored-txs/src/lib.rs
limemloh: Could we make the payload generic and leave it to the implementation?

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: I implemented the following: The CIS3 standard supports multi-sig accounts but because the `public_key_registry` in this contract maps one public key to one account, only one signature has to be provided for this contract. This signature has to be at the key 0 in both maps below:
```
  signature: BTreeMap<u8, BTreeMap<u8, SignatureEd25519>>
 ```

File: examples/cis3-nft-sponsored-txs/src/lib.rs (hunk 12 above)
Comment (Bargsteen) [reply examples/cis3-nft-sponsored-txs/src/lib.rs:279]: If we allow any type of payload, then we can completely remove the link to CIS-2, right?
Then CIS-3 can become a universal standard for sponsored transactions, which I think is useful.

File: examples/cis3-nft-sponsored-txs/src/lib.rs [hunk 16]
```diff
@@ -0,0 +1,2171 @@
...
+    host.state_mut().public_key_registry.entry(param.signer).and_modify(|(_, a)| *a += 1u64);
+
+    // Get the public key and the current nonce.
+    let entry = host
+        .state_mut()
+        .public_key_registry
+        .get(&param.signer)
+        .ok_or(CustomContractError::NoPublicKey)?;
+    let public_key = entry.0;
+    let nonce = entry.1;
```
abizjak, line 902:
```suggestion
    // Update the nonce.
    let mut entry = host.state_mut().public_key_registry.entry(param.signer).occupied_or(CustomContractError::NoPublicKey)?;
    entry.1 += 1; // bump nonce
    // Get the public key and the current nonce.
    let public_key = entry.0;
    let nonce = entry.1;
    drop(entry);
```

abizjak: To avoid a double lookup and make the code a bit simpler.

abizjak (approved):
You need to merge with main.

But otherwise this looks reasonable.

Maybe there will be some changes needed when developing the frontend and writing the standard, but for now this looks OK to me.

limemloh (approved):
Looks good!
Only some minor comments/considerations

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2169 @@
...
+    enable_logger
+)]
+fn contract_permit<S: HasStateApi>(
+    ctx: &impl HasReceiveContext,
+    host: &mut impl HasHost<State<S>, StateApiType = S>,
+    logger: &mut impl HasLogger,
+    crypto_primitives: &impl HasCryptoPrimitives,
+) -> ContractResult<()> {
+    // Parse the parameter.
+    let param: PermitParam = ctx.parameter_cursor().get()?;
```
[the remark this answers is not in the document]
DOBEN: There is a `wrongEntryPoint` error in the smart contract already which is thrown when the specified `entry_point` is accidentally swapped around (e.g. somebody wants to call the `transfer` function but provides the `operatorInputParamter`).

https://github.com/Concordium/concordium-rust-smart-contracts/blob/d00aa22d7f6e17111c5af6f6ee211bc947a8537e/examples/cis3-nft-sponsored-txs/src/lib.rs#L326

The embedded schema in the smart contract will specify which input parameters are supported for the `permit` function, so I hope users will make use of the schema and not come up with other inputParameters/functionEntryPoints that they think could be supported. In general, it will be difficult to figure out if this input Parameter deserialization fails due to a small mistake in the inputParameter provided or if the user completely misunderstood it and thought that additional entryPoints/inputParameters are supported. I suggest keeping it for now but I am open to suggestions if there are improvements we can do done the line.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -0,0 +1,2169 @@
...
+                param.message.entry_point,
+                OwnedEntrypointName::new_unchecked("contract_transfer".into()),
+                CustomContractError::WrongEntryPoint.into()
+            );
+
+            let TransferParams(transfers): TransferParameter = transfer_parameter;
+
+            for transfer_struct in transfers {
+                ensure!(
+                    transfer_struct.from.matches_account(&param.signer),
```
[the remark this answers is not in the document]
DOBEN: I will bring this up for discussion tomorrow and implement in a new MR if we want to do this change.

examples/cis3-nft-sponsored-txs/src/lib.rs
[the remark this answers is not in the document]
limemloh: While it is helpful, we should not (IMO) rely on the schema to provide information like this.
We could have an endpoint for endpoints supporting sponsored transactions, but a good error code is still useful.

examples/cis3-nft-sponsored-txs/src/lib.rs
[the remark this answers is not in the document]
abizjak: To finalize this, I think the agreement at the meeting was to do this for consistency, not that it would likely have any specific utility.

PR #226, opened 2023-02-27
Fix testing with crypto primitives.

### LINKED ISSUE #225: Running tests with secp256k1 leads to invalid modules.
Excessive memory usage in secpk1 .

Running `cargo concordium test` on the following example
```rust
//! # A Concordium V1 smart contract
use concordium_std::*;

#[concordium_cfg_test]
mod tests {
    use super::*;
    use test_infrastructure::*;

    #[concordium_test]
    fn test_ecdsa() {
        let crypto_primitives = TestCryptoPrimitives::new();

        let ecdsa_msg: [u8; 32] = [0; 32];
        let ecdsa_signature: [u8; 64] = [0; 64];
        let key: [u8; 33] = [0; 33];
        let ecdsa_ok = crypto_primitives.verify_ecdsa_secp256k1_signature(
            PublicKeyEcdsaSecp256k1(key),
            SignatureEcdsaSecp256k1(ecdsa_signature),
            ecdsa_msg,
        );
        claim!(ecdsa_ok, "ECDSA verify failed");
    }
}
```

Yields 
```
Caused by:
    Initial memory allocation of 34 pages exceeds maximum of 32.
```

The cause of this is that the crate secp256k1 has a number of static definitions to speed up various operations. These are allocated in initial Wasm memory which needs to be under 2MB.

abizjak:
## Purpose

Fixes #225 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std/CHANGELOG.md
+++ b/concordium-std/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Fix testing of crypto primitives using secp256k1 signature scheme.
+
 ## concordium-std 6.0.0 (2023-02-08)
 
 - [`wee_alloc`](https://docs.rs/wee_alloc/latest/wee_alloc/index.html) is no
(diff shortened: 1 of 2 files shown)


Bargsteen (approved):
Ah, so that was the reason. Nice that you fixed it! 👍

PR #237, opened 2023-03-16
Change signature verification

### LINKED ISSUE #236: Adjust the sponsored transaction signature verification in the smart contract to work with the Concordium browser wallet
**Task description**

The Concordium browser wallet prepends the message with the account and 8 (zero bytes) before hashing and signing it. This needs to be reflected in the smart contracts.

DOBEN:
## Purpose

closes #236 

## Changes

- Prepend the message with the account and 8 (zero bytes) before hashing it. This is how the Concordium browser wallet signs messages.
- Add `serilizationHelper` function which can be called by the front-end to serialize the `PermitMessage`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/cis3-nft-sponsored-txs/Cargo.toml
+++ b/examples/cis3-nft-sponsored-txs/Cargo.toml
@@ -7,14 +7,14 @@ license = "MPL-2.0"
 
 [features]
 default = ["std", "wee_alloc", "crypto-primitives"]
-std = ["concordium-std/std", "concordium-cis2/std"]
+std = ["concordium-std/std", "concordium-cis2/std", "hex/std"]
 wee_alloc = ["concordium-std/wee_alloc"]
 crypto-primitives = ["concordium-std/crypto-primitives"]
 
 [dependencies]
 concordium-std = {path = "../../concordium-std", default-features = false}
 concordium-cis2 = {path = "../../concordium-cis2", default-features = false}
-hex = "0.4.3"
+hex = { version = "0.4", default-features = false }
 
 [lib]
 crate-type=["cdylib", "rlib"]
--- a/examples/cis3-nft-sponsored-txs/Cargo.toml
+++ b/examples/cis3-nft-sponsored-txs/Cargo.toml
@@ -14,6 +14,7 @@ crypto-primitives = ["concordium-std/crypto-primitives"]
 [dependencies]
 concordium-std = {path = "../../concordium-std", default-features = false}
 concordium-cis2 = {path = "../../concordium-cis2", default-features = false}
+hex = "0.4.3"
 
 [lib]
 crate-type=["cdylib", "rlib"]
--- a/examples/cis3-nft-sponsored-txs/src/lib.rs
+++ b/examples/cis3-nft-sponsored-txs/src/lib.rs
@@ -770,16 +770,12 @@ fn contract_transfer<S: HasStateApi>(
     Ok(())
 }
 
-/// Helper function that can be invoked to serializ the `PermitMessage`.
-#[receive(
-    contract = "cis3_nft",
-    name = "serializationHelper",
-    parameter = "PermitMessage",
-    mutable
-)]
+/// Helper function that can be invoked at the front-end to serialize the
+/// `PermitMessage` before signing it in the wallet.
+#[receive(contract = "cis3_nft", name = "serializationHelper", parameter = "PermitMessage")]
 fn contract_serialization_helper<S: HasStateApi>(
     _ctx: &impl HasReceiveContext,
-    _host: &mut impl HasHost<State<S>, StateApiType = S>,
+    _host: &impl HasHost<State<S>, StateApiType = S>,
 ) -> ContractResult<()> {
     Ok(())
 }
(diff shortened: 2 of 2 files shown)


examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -916,8 +869,19 @@ fn contract_permit<S: HasStateApi>(
...
+    let message_hash = crypto_primitives
+        .hash_sha2_256(
+            &[
+                to_bytes(&param.signer),
+                zeros,
+                hex::encode(to_bytes(&param.message)).as_bytes().to_vec(),
+            ]
+            .concat(),
+        )
+        .0;
```
abizjak, line 884: It's rather annoying to have to do this hex encoding. I see it is necessary the way wallets are now.

This can also be done more efficiently I think without having to first deserialize the parameters and then serialize them back again here. It's a bit more complicated to write, but I think the gains will be worth it. I can try to write that when I'm in the office later today.

examples/cis3-nft-sponsored-txs/Cargo.toml
```diff
@@ -14,6 +14,7 @@ crypto-primitives = ["concordium-std/crypto-primitives"]
 [dependencies]
 concordium-std = {path = "../../concordium-std", default-features = false}
 concordium-cis2 = {path = "../../concordium-cis2", default-features = false}
+hex = "0.4.3"
```
abizjak, line 17: What is the size of this contract? I'm not sure we have other contracts with the hex dependency.

abizjak (approved):

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -770,65 +770,18 @@ fn contract_transfer<S: HasStateApi>(
...
-    name = "calculateMessageHash",
+    name = "serializationHelper",
     parameter = "PermitMessage",
-    return_value = "[u8;32]",
-    crypto_primitives,
     mutable
 )]
-fn contract_calculate_message_hash<S: HasStateApi>(
-    ctx: &impl HasReceiveContext,
+fn contract_serialization_helper<S: HasStateApi>(
```
limemloh, line 780: I don't understand the point of this receive function.
If it is needed, consider adding more to the description

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -770,65 +770,18 @@ fn contract_transfer<S: HasStateApi>(
...
 
-/// Helper function to calculate the message hash used in the permit function.
-///
-/// It rejects if:
-/// - It fails to parse the parameter.
-/// - The message was intended for a different contract.
-/// - The message was intended for a different `entry_point`.
-/// - The message is expired.
-
+/// Helper function that can be invoked to serializ the `PermitMessage`.
```
limemloh, line 773:
```suggestion
/// Helper function that can be invoked to serialize the `PermitMessage`.
```

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: This function is only meant as a method to serialize the `PermitMessage` at the front-end before signing it in the wallet. It is technically a `view` function that should return nothing except for having an input parameter (the `PermitMessage`).

https://github.com/Concordium/concordium-browser-wallet/blob/df9e3cb8672d0cc7dabf6f64836b699ece53d15b/examples/sponsoredTXs/src/sponsoredTx.tsx#L176

DOBEN: I removed the `mutative` and updated the explanation:
```
Helper function that can be invoked at the front-end to serialize the `PermitMessage` before signing it in the wallet.
```

examples/cis3-nft-sponsored-txs/Cargo.toml
DOBEN: cargo concordium build -e
smart contract module 172.766 kB

cargo concordium build
smart contract module 166.729 kB

examples/cis3-nft-sponsored-txs/src/lib.rs
abizjak: I pushed a commit to this branch that does some of these optimizations.

I am not sure it is worth it as such, so feel free to revert it.

However one advantage is that we can now build the contract with `no-std`. With that the contract size is around 102kB.

examples/cis3-nft-sponsored-txs/src/lib.rs
abizjak: I will just hijack this conversation (I also find this functions rather puzzling and I don't understand why it is needed).

There is a function in the JS helpers which takes a schema and JSON and serializes it into a buffer.

DOBEN: We need the `PermitMessage` in the schema and therefore it needs to be an input parameter of some function in the smart contract (currently there is no function using the plain `PermitMessage` as input). We could create the schema (so we have it for the front-end) and then remove this function from the smart contract again. I just think that it might be easier to understand for outside people if they see where we got this `schema` (that serializes the `PermitMessage`) comes from and hence leaving this dummy `serializingHelper` in the smart contract code. But I can also remove it.

abizjak: Right, I missed that. I read both as having PermitParams, sorry about that.

limemloh (approved):

PR #246, opened 2023-04-13
Fix derive(SchemaType) macro when used without serialization. Add helper

### LINKED ISSUE #238: derive(SchemaType) cannot be used without derive(Serialize) when attributes are used
The following does not work
```
#[derive(SchemaType)]
struct Foo {
     #[concordium(size_length = 1)]
     foo: BTreeMap<u8, u64>
}
```

Because the SchemaType macro does not declare `concordium` as an attribute. The following works
```
#[derive(Serialize,SchemaType)]
struct Foo {
     #[concordium(size_length = 1)]
     foo: BTreeMap<u8, u64>
}
```

While the cases of where we want to derive SchemaType, but not serialization, are likely limited, this is still a bug, and those cases do exist since a more efficient, serialization can sometimes be achieved than what the derived one is (e.g., we can reuse buffers when deserializing perhaps).

abizjak:
Fixes #238 .

Additionally add a helper `write_root` function that can be used in migrations. It is dual to the already existing `read_root`.

## Checklist

      hard-to-understand areas.


concordium-std/src/traits.rs
```diff
@@ -232,6 +232,12 @@ pub trait HasStateApi: Clone {
     fn read_root<A: DeserialWithState<Self>>(&self) -> ParseResult<A> {
         A::deserial_with_state(self, &mut self.lookup_entry(&[]).ok_or(ParseError {})?)
     }
+
+    /// Serialize and write the state at the root of the state trie. Dual to
+    /// [`read_root`](Self::read_root).
+    fn write_root<A: Serial>(&mut self, new_root: &A) {
```
DOBEN, line 238:
```suggestion
    fn write_root<A: Serial>(&mut self, new_state: &A) 
```

abizjak: Good suggestion.

abizjak: Fixed now.

DOBEN (approved):

PR #247, opened 2023-04-13
Add NonceEvent

DOBEN:
## Purpose

- Add NonceEvent
- Bump nonce at the end. This allows the front-end to use the current nonce it queries from the smart contract (without having to increase it by 1) to create the permitMessage.
- Improve comments about the signing and message generation based on feedback and questions got from people.

### DIFF
--- a/examples/cis3-nft-sponsored-txs/src/lib.rs
+++ b/examples/cis3-nft-sponsored-txs/src/lib.rs
@@ -84,7 +84,7 @@ pub const REGISTRATION_EVENT_TAG: u8 = 0u8;
 pub const NONCE_EVENT_TAG: u8 = u8::MAX - 5;
 
 /// Tagged events to be serialized for the event log.
-#[derive(Debug, Serial)]
+#[derive(Debug)]
 enum Event {
     /// Registration of a public key for a given account. The
     /// corresponding private key will have to sign the message that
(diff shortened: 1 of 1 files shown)


examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -78,8 +78,10 @@ const SUPPORTS_STANDARDS: [StandardIdentifier<'static>; 3] =
 const SUPPORTS_PERMIT_ENTRYPOINTS: [EntrypointName; 2] =
     [EntrypointName::new_unchecked("updateOperator"), EntrypointName::new_unchecked("transfer")];
 
-/// Tag for the CIS3 Registration event.
+/// Tag for the Registration event.
 pub const REGISTRATION_EVENT_TAG: u8 = 0u8;
+/// Tag for the CIS3 Nonce event.
+pub const NONCE_EVENT_TAG: u8 = u8::MAX - 5;
```
abizjak, line 84: Why did you choose this value for it? I don't think this'll work since the serialization is derived, so the tag of the "Nonce" variant below will be 1, not 250.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -78,8 +78,10 @@ const SUPPORTS_STANDARDS: [StandardIdentifier<'static>; 3] =
...
 
 /// Tagged events to be serialized for the event log.
 #[derive(Debug, Serial)]
```
abizjak, line 87: This serialization here.

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: Right, I fixed the `Serial`. The CIS-3 standard defines `250` as tag for this event.

abizjak (approved):

PR #248, opened 2023-04-17
Adjust payload to use not an enum anymore

DOBEN:
## Purpose

Based on this discussion, we intend to remove the tag at the payload.
https://github.com/Concordium/concordium-update-proposals/pull/41#discussion_r1168010841


abizjak (approved):
Looks good.

We could in principle make this a tiny bit more efficient by not deserializing into a vector first.

But it is fine for a demo like this.

PR #249, opened 2023-04-17
Reexport quickcheck

annenkov:
## Purpose

Make `quickcheck` available when writing property based tests. This is useful for defining `Arbitrary` instances for custom data types in smart contracts.

### DIFF
--- a/concordium-std/src/lib.rs
+++ b/concordium-std/src/lib.rs
@@ -298,4 +298,7 @@ pub use types::*;
 #[cfg_attr(feature = "wee_alloc", global_allocator)]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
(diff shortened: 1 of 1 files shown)


Bargsteen (approved):
LGTM. But it should also go into the changelog.

abizjak:
I am not convinced this is a good idea. We are not re-exporting third-party libraries in this crate and I am not sure we should.

What is the problem of just adding that dependency wherever it is needed?

annenkov:
After some discussions it was decided not to re-export `quickcheck`, but use it in a smart contract as a dependency. That is, by adding in the smart contract's project `Cargo.toml`

```toml
[dependencies]
quickcheck = { version = "1.0.3"}
```

abizjak:
> What are the downsides of re-exporting?

To document here, the downside is that it makes the public API larger and maintenance more complex. It's a tradeoff and at this time it seems it is better to not re-export.

PR #255, opened 2023-04-25
Add viewMessageHash function

DOBEN:
## Purpose

Improvements and fixes.

## Changes

- Add `viewMessageHash` helper function.
- Add Display + FromStr Trait to PublicKeyEd25519/PublicKeyEcdsaSecp256k1/SignatureEd25519/SignatureEcdsaSecp256k1/HashSha2256 types.

- Fix `cargo clippy warning`: calling `set_len()` immediately after reserving a buffer creates uninitialized values
https://rust-lang.github.io/rust-clippy/master/index.html#uninit_vec.
- Fix warning of duplicated licence field in `proxy` example.
- Fix `two-step-transfer` example dependency: `concordium-quickcheck` needs to be a `dependency` (not `dev_dependency`) otherwise `cargo concordium test` will not run.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std/src/types.rs
+++ b/concordium-std/src/types.rs
@@ -1036,13 +1036,10 @@ impl FromStr for PublicKeyEd25519 {
             return Err(ParseError {});
         }
 
-        let public_key: [u8; 32] = (0..s.len())
-            .step_by(2)
-            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).unwrap())
-            .collect::<Vec<u8>>()
-            .as_slice()
-            .try_into()
-            .unwrap();
+        let mut public_key: [u8; 32] = [0u8; 32];
+        for (i, place) in public_key.iter_mut().enumerate() {
+            *place = u8::from_str_radix(&s[2 * i..2 * i + 2], 16).map_err(|_| ParseError {})?;
+        }
 
         Ok(PublicKeyEd25519(public_key))
     }
@@ -1070,13 +1067,10 @@ impl FromStr for PublicKeyEcdsaSecp256k1 {
             return Err(ParseError {});
         }
 
-        let public_key: [u8; 33] = (0..s.len())
-            .step_by(2)
-            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).unwrap())
-            .collect::<Vec<u8>>()
-            .as_slice()
-            .try_into()
-            .unwrap();
+        let mut public_key: [u8; 33] = [0u8; 33];
+        for (i, place) in public_key.iter_mut().enumerate() {
+            *place = u8::from_str_radix(&s[2 * i..2 * i + 2], 16).map_err(|_| ParseError {})?;
+        }
 
         Ok(PublicKeyEcdsaSecp256k1(public_key))
     }
@@ -1104,13 +1098,10 @@ impl FromStr for SignatureEd25519 {
             return Err(ParseError {});
         }
 
-        let signature: [u8; 64] = (0..s.len())
-            .step_by(2)
-            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).unwrap())
-            .collect::<Vec<u8>>()
-            .as_slice()
-            .try_into()
-            .unwrap();
+        let mut signature: [u8; 64] = [0u8; 64];
+        for (i, place) in signature.iter_mut().enumerate() {
+            *place = u8::from_str_radix(&s[2 * i..2 * i + 2], 16).map_err(|_| ParseError {})?;
+        }
 
         Ok(SignatureEd25519(signature))
     }
@@ -1139,13 +1130,10 @@ impl FromStr for SignatureEcdsaSecp256k1 {
             return Err(ParseError {});
         }
 
-        let signature: [u8; 64] = (0..s.len())
-            .step_by(2)
-            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).unwrap())
-            .collect::<Vec<u8>>()
-            .as_slice()
-            .try_into()
-            .unwrap();
+        let mut signature: [u8; 64] = [0u8; 64];
+        for (i, place) in signature.iter_mut().enumerate() {
+            *place = u8::from_str_radix(&s[2 * i..2 * i + 2], 16).map_err(|_| ParseError {})?;
+        }
 
         Ok(SignatureEcdsaSecp256k1(signature))
     }
@@ -1173,13 +1161,10 @@ impl FromStr for HashSha2256 {
             return Err(ParseError {});
         }
 
-        let hash: [u8; 32] = (0..s.len())
-            .step_by(2)
-            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).unwrap())
-            .collect::<Vec<u8>>()
-            .as_slice()
-            .try_into()
-            .unwrap();
+        let mut hash: [u8; 32] = [0u8; 32];
+        for (i, place) in hash.iter_mut().enumerate() {
+            *place = u8::from_str_radix(&s[2 * i..2 * i + 2], 16).map_err(|_| ParseError {})?;
+        }
 
         Ok(HashSha2256(hash))
     }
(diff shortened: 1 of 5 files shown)


examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -1304,6 +1355,29 @@ fn contract_set_implementor<S: HasStateApi>(
...
+fn decode_hex(s: &str) -> Result<Vec<u8>, CustomContractError> {
+    if s.len() % 2 != 0 {
+        Err(CustomContractError::ParseError)
+    } else {
+        (0..s.len())
+            .step_by(2)
+            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).map_err(|e| e.into()))
+            .collect()
+    }
+}
```
abizjak, line 1369: Right, so the idea was to add a `FromStr` implementation to all these types https://docs.rs/concordium-std/latest/concordium_std/struct.HashSha2256.html https://docs.rs/concordium-std/latest/concordium_std/struct.PublicKeyEd25519.html https://docs.rs/concordium-std/latest/concordium_std/struct.PublicKeyEcdsaSecp256k1.html and the signature types.

In concordium-std directly.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -1304,6 +1355,29 @@ fn contract_set_implementor<S: HasStateApi>(
...
+
+// Helper function to encode bytes into a hex string.
+#[allow(dead_code)]
+fn encode_hex(bytes: &[u8]) -> String {
+    let mut s = String::with_capacity(bytes.len() * 2);
+    for &b in bytes {
+        s = format!("{}{:02x}", s, b);
+    }
+    s
+}
```
abizjak, line 1379: And this should be the display implementation, e.g., 
```
impl std::fmt::Display for HashSha2256 {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        for b in self.0 {
            write!(f, "{:02x}", b)?;
        }
        Ok(())
    }
}
```

in concordium-std directly (or contracts-common if the types are defined there).

Then you can just do `to_string`

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -829,6 +837,50 @@ fn contract_serialization_helper<S: HasStateApi>(
...
+    // The message signed in the Concordium browser wallet is prepended with the
+    // `account` address and 8 zero bytes.
+    let mut msg_prepend = vec![0; 32 + 8];
+    // Prepend the `account` address of the signer.
+    msg_prepend[0..32].copy_from_slice(param.signer.as_ref());
+    // Prepend 8 zero bytes.
+    msg_prepend[32..40].copy_from_slice(&[0u8; 8]);
+    // Calculate the message hash.
+    let message_hash =
+        crypto_primitives.hash_sha2_256(&[&msg_prepend[0..40], &message_bytes].concat()).0;
```
abizjak, line 879: Shouldn't we reuse this function in the verification function as well, where we compute the same hash?

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -829,6 +829,49 @@ fn contract_serialization_helper<S: HasStateApi>(
...
+fn contract_view_message_hash<S: HasStateApi>(
+    ctx: &impl HasReceiveContext,
+    _host: &mut impl HasHost<State<S>, StateApiType = S>,
+    crypto_primitives: &impl HasCryptoPrimitives,
+) -> ContractResult<[u8; 32]> {
+    // Parse the parameter.
+    let mut cursor = ctx.parameter_cursor();
+    // The input parameter is `PermitParam` but we only read the initial part of it
+    // with `PermitParamPartial`. I.e. we only read the `signature`, the
+    // `signer` but WITHOUT the `message` here.
```
Bargsteen, line 850:
```suggestion
    // with `PermitParamPartial`. I.e. we read the `signature` and the
    // `signer`,but not `message` here.
```

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -829,6 +829,49 @@ fn contract_serialization_helper<S: HasStateApi>(
...
+    let param: PermitParamPartial = cursor.get()?;
+
+    // The input parameter is `PermitParam` but we only read the initial part of it
```
Bargsteen, line 853:
```suggestion
    // The input parameter is `PermitParam` but we've only read the initial part of it
```

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -829,6 +829,49 @@ fn contract_serialization_helper<S: HasStateApi>(
...
+    // with `PermitParamPartial` so far. We read in the `message` now.
+    // `(cursor.size() - cursor.cursor_position()` is the length of the message in
+    // bytes.
+    let mut message_bytes = vec![0; (cursor.size() - cursor.cursor_position()) as usize];
+
+    cursor.read_exact(&mut message_bytes)?;
+
+    // The message signed in the Concordium browser wallet is prepended with the
+    // `account` address and 8 zero bytes.
```
Bargsteen, line 862: What are the eight zero bytes? Could you explain it in the comment? 😊

Bargsteen (approved):
LGTM 👍

concordium-std/src/types.rs
```diff
@@ -1018,27 +1019,172 @@ pub struct ExternCryptoPrimitives;
...
+            return Err(ParseError {});
+        }
+
+        let public_key: [u8; 32] = (0..s.len())
+            .step_by(2)
+            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).unwrap())
+            .collect::<Vec<u8>>()
+            .as_slice()
+            .try_into()
+            .unwrap();
```
abizjak, line 1045:
```suggestion
        let mut public_key: [u8; 32] = [0u8; 32];
        for (i, place) in public_key.iter_mut().enumerate() {
            *place = u8:;from_str_radix(&s[2*i, 2*i+2], 16).map_err(|_| ParseError{})?;
        }
```

We should not use `unwrap` since parsing is supposed to be able to fail, and if it does so it should do it gracefully.

concordium-std/src/types.rs
```diff
@@ -1018,27 +1019,172 @@ pub struct ExternCryptoPrimitives;
...
+            return Err(ParseError {});
+        }
+
+        let public_key: [u8; 33] = (0..s.len())
+            .step_by(2)
+            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).unwrap())
+            .collect::<Vec<u8>>()
+            .as_slice()
+            .try_into()
+            .unwrap();
```
abizjak, line 1079: Similarly here.

concordium-std/src/types.rs
```diff
@@ -1018,27 +1019,172 @@ pub struct ExternCryptoPrimitives;
...
+            return Err(ParseError {});
+        }
+
+        let signature: [u8; 64] = (0..s.len())
+            .step_by(2)
+            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).unwrap())
+            .collect::<Vec<u8>>()
+            .as_slice()
+            .try_into()
+            .unwrap();
```
abizjak, line 1113: And here.

abizjak (commented):
Looks good if you address those unwraps.

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: The 8 bytes zero are just a replication of what the browser wallet is doing at the moment. The 8 zero bytes could be used in the future in case we want to add versioning or we want to distinguish different signing mechanisms but for the time being there is no further meaning of the 8 zero bytes. As long as we don't have a good explanation for what the 8 zero bytes could represent in the future, I would leave the sentence as it is.

abizjak: The 8 zero bytes are quite important to be all 0.

This makes sure the user is not accidentally signing a transaction!

DOBEN: Could you elaborate on what is encoded instead of the 8 zero bytes in a normal transaction (which will never be 0) then I can add a sentence about how these 8 bytes are used to distinguish between signing a message and signing a transaction, please.

PR #257, opened 2023-04-25
Added schema for the view function's return value in piggybank contract

rasmus-kirk:
## Purpose & Changes

I specified a return type PiggyBank smart contract's view function so that a schema can be generated. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/piggy-bank/part1/src/lib.rs
+++ b/examples/piggy-bank/part1/src/lib.rs
@@ -17,13 +17,17 @@
 use concordium_std::*;
 
 /// The state of the piggy bank
-#[derive(Debug, Serialize, PartialEq, Eq, Clone, Copy)]
+#[derive(Debug, SchemaType, Serialize, PartialEq, Eq, Clone, Copy)]
 enum PiggyBankState {
     /// Alive and well, allows for CCD to be inserted.
     Intact,
     /// The piggy bank has been emptied, preventing further CCD to be inserted.
     Smashed,
 }
+
+/// The return type of the view function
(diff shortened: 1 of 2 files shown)


examples/piggy-bank/part1/src/lib.rs
```diff
@@ -74,12 +78,12 @@ fn piggy_smash<S: HasStateApi>(
...
+#[receive(contract = "PiggyBank", name = "view", return_value = "ViewState")]
 fn piggy_view<S: HasStateApi>(
     _ctx: &impl HasReceiveContext,
     host: &impl HasHost<PiggyBankState, StateApiType = S>,
-) -> ReceiveResult<(PiggyBankState, Amount)> {
+) -> ReceiveResult<ViewState> {
     let current_state = *host.state();
     let current_balance = host.self_balance();
-    Ok((current_state, current_balance))
+    Ok(ViewState(current_state, current_balance))
```
rasmus-kirk, line 88: Not as elegant as the regular tuple, but I don't think the macro supports it

abizjak: Why do you say that?
```
#[receive(contract = "PiggyBank", name = "view", return_value = "(PiggyBankState, Amount)")]
```
works perfectly fine as I see it.

rasmus-kirk: I was mistaking, I tested it and most likely got another error and thought that the tuple was the culprit.

abizjak (approved):

PR #259, opened 2023-04-28
Fix CI jobs

DOBEN:
## Purpose
It seems that two `run commands` with parallel matrixes caused issues and some CI pipeline jobs weren't correctly marked as being completed. Resolve issue by simplifying CI pipeline.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -488,9 +488,6 @@ jobs:
           - examples/signature-verifier/Cargo.toml
           - examples/transfer-policy-check/Cargo.toml
 
-        withquickcheck:
-          - examples/two-step-transfer/Cargo.toml
-
     steps:
       - name: Checkout sources
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/linter.yml
```diff
@@ -512,12 +510,6 @@ jobs:
           command: clippy
           args: --manifest-path ${{ matrix.crates }} --target=${{ matrix.target }} --features concordium-std/build-schema --features concordium-std/wasm-test -- -D warnings
 
-      - name: Run cargo clippy with schema and tests requiring quickcheck
-        uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: --manifest-path ${{ matrix.withquickcheck }} --target=${{ matrix.target }} --features concordium-std/build-schema --features concordium-std/concordium-quickcheck -- -D warnings
```
abizjak, line 519: So the quickcheck feature should be enabled for this crate. I'm a bit confused now why this works where I needed to introduce it previously.

abizjak: Right, it works because you moved the quickcheck from a dev-dependency to a normal dependency in https://github.com/Concordium/concordium-rust-smart-contracts/pull/255

abizjak (approved):

PR #262, opened 2023-05-01
Add smart contract upgrade example

DOBEN:
## Purpose

closes #260 

Add an example how to execute a `low-level` upgrade with a migration function that adjusts the state from the old `Contract_Version1` to the new state from `Contract_Version2`.

## Changes

- Add smart contract `Contract_Version1`
- Add smart contract `Contract_Version2`

TODO: Unit tests with `low-level` upgrade functions need to be constructed differently.

## Checklist

- [x] My code follows the style of this project.
- [x] The code compiles without warnings.
- [x] I have performed a self-review of the changes.
- [x] I have documented my code, in particular the intent of the
      hard-to-understand areas.
- [x] (If necessary) I have updated the CHANGELOG.

PR #267, opened 2023-05-03
Add initial credential registry storage contract

DOBEN:
## Purpose

Add initial credential registry storage contract


DOBEN:
One part that is still unclear to me is:
Can the credential be only decoded off-chain when having access to the private key (associated with the public key) that stored the credential in this contract? Meaning the credential stored in the wallet is the only entity that can decode the stored `credential_secret` in this smart contract?

abizjak:
> 

Yes, that is the intention.

examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// A timestamp to make signatures expire.
+    timestamp:         Timestamp,
+    /// The entry_point that the signature is intended for.
+    entry_point:       OwnedEntrypointName,
+    /// Metadata associated with the credential.
+    metadata:          Metadata,
+    /// The serialized credential_secret.
+    #[concordium(size_length = 2)]
+    credential_secret: Vec<u8>,
+}
```
abizjak, line 171: I am not sure this is needed here. The idea is that the keys are separate anyhow and as I though we just sign the message prefixed with
```
WEB3ID:STORE
```

Do you think it is necessary to add this context to the message?

examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+    // that is already implemented in the Concordium browser wallet. The message
+    // signed in the Concordium browser wallet is prepended with the `account`
+    // address and 8 zero bytes. Accounts in the Concordium browser wallet can
+    // either sign a regular transaction (in that case the prepend is `account`
+    // address and the nonce of the account which is by design >= 1) or sign a
+    // message (in that case the prepend is `account` address and 8 zero bytes).
+    // Hence, the 8 zero bytes ensure that the user does not accidentally sign a
+    // transaction. The account nonce is of type u64 (8 bytes).
+    let mut msg_prepend = vec![0; 32 + 8];
+    // Prepend the `account` address of the signer.
```
abizjak, line 241: This comment does not seem correct.

examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+    msg_prepend[0..32].copy_from_slice(&param.public_key.0);
+    // Prepend 8 zero bytes.
+    msg_prepend[32..40].copy_from_slice(&[0u8; 8]);
```
abizjak, line 244: This is not necessary, but we should instead prepend `WEB3ID:STORE`. That's the chosen prefix.

examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+    // Check that the signature was intended for this contract.
+    ensure_eq!(message.contract_address, ctx.self_address(), CustomContractError::WrongContract);
+
+    // Check signature is not expired.
+    ensure!(message.timestamp > ctx.metadata().slot_time(), CustomContractError::Expired);
+
+    // Check signature was intended for this `entry_point`.
+    ensure_eq!(
+        message.entry_point.as_entrypoint_name(),
+        EntrypointName::new_unchecked("store"),
```
abizjak, line 260: I think you should be able to use just `"store"` here. I think I added the ability to check equality of entrypoint name and string.

examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+    // signed in the Concordium browser wallet is prepended with the `account`
+    // address and 8 zero bytes. Accounts in the Concordium browser wallet can
+    // either sign a regular transaction (in that case the prepend is `account`
+    // address and the nonce of the account which is by design >= 1) or sign a
+    // message (in that case the prepend is `account` address and 8 zero bytes).
+    // Hence, the 8 zero bytes ensure that the user does not accidentally sign a
+    // transaction. The account nonce is of type u64 (8 bytes).
+    let mut msg_prepend = vec![0; 32 + 8];
+    // Prepend the `account` address of the signer.
+    msg_prepend[0..32].copy_from_slice(&param.public_key.0);
```
abizjak, line 242: I don't think we need to prepend the public key to the message.

examples/credential-registry-storage-contract/src/lib.rs
DOBEN: This prefixed seems `arbitrary`. Meaning if third parties clone this smart contract and use it for their own use case, they also come up with `arbitrary` prefixes, and one day there are two smart contracts with the same prefix and the same input parameter length and you could use one signature on both contracts. I would suggest adding the `contract_address` to the signed message to avoid that problem. `Timestamp` is discussable (it is a nice feature to have expired signatures). `entry_point` is needless since we only have one entrypoint in this contract (we can remove it or keep it if we want to show how to write a contract which has several signature verification entrypoints).

examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+    crypto_primitives: &impl HasCryptoPrimitives,
+) -> ContractResult<()> {
+    // Parse the parameter.
+    let param: StoreParam = ctx.parameter_cursor().get()?;
+
+    // Check that the signature was intended for this contract.
+    ensure_eq!(param.contract_address, ctx.self_address(), CustomContractError::WrongContract);
+
+    // Check signature is not expired.
+    ensure!(param.timestamp > ctx.metadata().slot_time(), CustomContractError::Expired);
```
abizjak, line 274:
```suggestion
    ensure!(param.timestamp >= ctx.metadata().slot_time(), CustomContractError::Expired);
```

examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+
+    // Perepare message bytes as it is signed by the wallet.
+    // Note that the message is prepended by a domain separation string.
+    let mut message: Vec<u8> = SIGNARUTE_DOMAIN.as_bytes().to_vec();
+    param.message_bytes(&mut message)?;
+
+    // Calculate the message hash.
+    let message_hash = crypto_primitives.hash_sha2_256(&message).0;
```
abizjak, line 282: We don't really need to hash it since we are designing a new protocol anyhow. The signature scheme hashes the message already anyhow.

abizjak: That should make it cheaper as well.

abizjak (approved):
Looks good. We might need some small tweaks later when the wallet starts implementing things. But I think it's good to merge now.

examples/smart-contract-upgrade/contract-version1/src/lib.rs
```diff
@@ -47,7 +47,7 @@ struct UpgradeParams {
 
 /// Smart contract errors.
 #[derive(PartialEq, Eq, Reject, Serial, SchemaType)]
-enum CostumContractError {
+enum CustomContractError {
```
annenkov, line 50: Is the `smart-contract-upgade` contract related to the credential storage contract?

examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+    timestamp:            Timestamp,
+}
+
+/// Helper function that can be invoked at the front-end to serialize
+/// the message to be signed by the wallet.
+#[receive(
+    contract = "credential-registry-storage",
+    name = "serializationHelper",
+    parameter = "SerializationHelperParam"
+)]
```
annenkov, line 235: Could you please explain a bit more about how it works on the front-end side? (If you think it makes sense to explain here)
I see that the input parameter is serialized when the entrypoint is called, but it doesn't return anything.

annenkov (approved):
Looks good, just a couple of minor comments.

examples/smart-contract-upgrade/contract-version1/src/lib.rs
DOBEN: It is just a leftover comment from my last MR:
https://github.com/Concordium/concordium-rust-smart-contracts/pull/262#discussion_r1182565655

examples/credential-registry-storage-contract/src/lib.rs
DOBEN: In the 'old' signing mechanism, we needed to serialize the `SerializationHelperParam` (which includes all the input parameters from `StoreParam` except for the `signature` and the `public_key`). Meaning `SerializationHelperParam` includes everything the wallet is supposed to sign. In the below code line, you can see we only need the schema of the `serializationHelper` function. The `serializationHelper` function is not executed at any point in time, therefore the logic of the function is irrelevant which is why the function has no logic and just returns `Ok(())`.

https://github.com/Concordium/concordium-dapp-examples/blob/main/sponsoredTransactions/frontend/src/SponsoredTransactions.tsx#L180

The 'new' signing mechanism for the verifiable credentials is still not yet implemented and subject to changes in the wallets. I added this statement above the function but I think we will revisit this function once we have a working frontend/browser wallet example:
```
Helper function that can be invoked at the front end to serialize the `SerializationHelperParam` to be signed by the wallet. `SerializationHelperParam` includes all the input parameters from `StoreParam` except for the `signature` and the `public_key`. We need only the input parameter schema of this function at the front end. The `serializationHelper` function is not executed at any point in time, therefore the logic of the function is irrelevant.
```

PR #280, opened 2023-05-24
Block time alias

EmilLa1:
## Purpose

Introduce an alias `block_time` for the primitive `get_slot_time`.

This is relevant from P6 and onwards as slots are no longer a thing, instead blocks have an associated timestamp of when the block is being baked.

## Changes

Introduced `block_time` which just calls `get_slot_time`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std/CHANGELOG.md
+++ b/concordium-std/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased changes
 
 - Set minimum Rust version to 1.65.
+- Introduce `block_time` for `HasChainMetadata` which should be used for contracts written on P6 and onwards.
 
 ## concordium-std 6.2.0 (2023-05-08)
 
--- a/concordium-std/src/impls.rs
+++ b/concordium-std/src/impls.rs
@@ -1640,6 +1640,11 @@ impl HasChainMetadata for ExternChainMeta {
     fn slot_time(&self) -> SlotTime {
         Timestamp::from_timestamp_millis(unsafe { prims::get_slot_time() })
     }
+
+    #[inline(always)]
+    fn block_time(&self) -> Timestamp {
+        Timestamp::from_timestamp_millis(unsafe { prims::get_slot_time() })
+    }
 }
 
 impl AttributesCursor {
--- a/concordium-std/src/impls.rs
+++ b/concordium-std/src/impls.rs
@@ -1642,9 +1642,7 @@ impl HasChainMetadata for ExternChainMeta {
     }
 
     #[inline(always)]
-    fn block_time(&self) -> Timestamp {
-        Timestamp::from_timestamp_millis(unsafe { prims::get_slot_time() })
-    }
+    fn block_time(&self) -> Timestamp { self.slot_time() }
 }
 
 impl AttributesCursor {
--- a/concordium-std/src/prims.rs
+++ b/concordium-std/src/prims.rs
@@ -223,6 +223,8 @@ extern "C" {
 
     // Getters for the chain meta data
     /// Slot time (in milliseconds) from chain meta data
+    /// Note that this is a Timestamp (unix timestamp with millisecond
+    /// precision).
     pub fn get_slot_time() -> u64;
 
     // Cryptographic primitives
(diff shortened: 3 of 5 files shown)


concordium-std/src/prims.rs
```diff
@@ -223,6 +223,8 @@ extern "C" {
 
     // Getters for the chain meta data
     /// Slot time (in milliseconds) from chain meta data
+    /// Note that this is a Timestamp (unix timestamp with millisecond
+    /// precision).
```
abizjak, line 227: The comment just above already states it is milliseconds.

concordium-std/src/test_infrastructure.rs
```diff
@@ -448,6 +448,8 @@ fn unwrap_ctx_field<A>(opt: Option<A>, name: &str) -> A {
 // Getters for testing-context
 impl HasChainMetadata for TestChainMeta {
     fn slot_time(&self) -> SlotTime { unwrap_ctx_field(self.slot_time, "metadata.slot_time") }
+
+    fn block_time(&self) -> Timestamp { self.slot_time() }
```
abizjak, line 452: This is not necessary since there is a default implementation.

concordium-std/src/traits.rs
```diff
@@ -39,6 +39,8 @@ pub trait HasCallResponse: Read {
 pub trait HasChainMetadata {
     /// Get time in milliseconds at the beginning of this block.
     fn slot_time(&self) -> SlotTime;
+    /// Get time in milliseconds at the beginning of this block.
+    fn block_time(&self) -> Timestamp;
```
abizjak, line 43: Add default implementation here seeing that this is an alias.

concordium-std/src/impls.rs
```diff
@@ -1640,6 +1640,9 @@ impl HasChainMetadata for ExternChainMeta {
     fn slot_time(&self) -> SlotTime {
         Timestamp::from_timestamp_millis(unsafe { prims::get_slot_time() })
     }
+
+    #[inline(always)]
+    fn block_time(&self) -> Timestamp { self.slot_time() }
```
abizjak, line 1645: Add default implementation instead. Then there is no breaking change.

abizjak (changes requested):

concordium-std/src/prims.rs
EmilLa1: Yea.. I removed one of the 'milliseconds' :)

concordium-std/src/traits.rs
```diff
@@ -39,6 +39,8 @@ pub trait HasCallResponse: Read {
 pub trait HasChainMetadata {
     /// Get time in milliseconds at the beginning of this block.
     fn slot_time(&self) -> SlotTime;
+    /// Get time in milliseconds at the beginning of this block.
+    fn block_time(&self) -> Timestamp { self.slot_time() }
```
abizjak, line 43:
```suggestion
    #[inline(always)]
    fn block_time(&self) -> Timestamp { self.slot_time() }
```

concordium-std/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased changes
 
 - Set minimum Rust version to 1.65.
+- Introduce `block_time` for `HasChainMetadata` which should be used for contracts written on P6 and onwards.
```
abizjak, line 6:
```suggestion
- Introduce a new method `block_time` as an alias to `slot_time` in the `HasChainMetadata` trait.
```

There is no reason for telling them to use it from p6 onward. There is nothing wrong with using slot_time since it is semantically equivalent.

concordium-std/src/prims.rs
```diff
@@ -222,7 +222,9 @@ extern "C" {
     pub fn get_receive_entrypoint(start: *mut u8);
 
     // Getters for the chain meta data
-    /// Slot time (in milliseconds) from chain meta data
+    /// Slot time from chain meta data
```
abizjak, line 225:
```suggestion
    /// Slot time from chain meta data.
```

abizjak (approved):

Issue #283, opened 2023-05-25
Derive serialization and schema type for types with generics


limemloh:
Deriving any of `SchemaType`, `Serial` and `Deserial` fails for types when any of the fields are generic.
We could derive the implementation, bounding the generics to implement the trait.

- [x] `Serial` and `Deserial`
- [x] `SchemaType`

PR #286, opened 2023-05-25
Add test cases for deriving Serial and Deserial

limemloh:
## Purpose

Closes https://github.com/Concordium/concordium-rust-smart-contracts/issues/283.
Should be merged after https://github.com/Concordium/concordium-contracts-common/pull/88

## Changes

- Bumps contracts-common
- Add test cases for deriving `Serial` and `Deserial`.
- Support deriving `SchemaType` for types with type parameters.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std-derive/src/lib.rs
+++ b/concordium-std-derive/src/lib.rs
@@ -1567,8 +1567,6 @@ fn schema_type_derive_worker(input: TokenStream) -> syn::Result<TokenStream> {
 
     let data_name = &ast.ident;
 
-    let (impl_generics, ty_generics, where_clauses) = ast.generics.split_for_impl();
-
     let body = match ast.data {
         syn::Data::Struct(ref data) => {
             let fields_tokens = schema_type_fields(&data.fields)?;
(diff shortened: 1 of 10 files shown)


abizjak (commented):
Looks good, but some cases where you have bounds on generics would be useful as well.

For example what happens when you do
```
struct Foo<A: Serial> {
   inner: A
}
```
?

abizjak:
Also, we kind of have to do the same for SchemaType (adding generic bounds) otherwise we will have an inconsistent experience, where serialize can be derived, but SchemaType cannot.

limemloh:
> Looks good, but some cases where you have bounds on generics would be useful as well.
> 
> For example what happens when you do
> 
> ```
> struct Foo<A: Serial> {
>    inner: A
> }
> ```
> 
> ?

So I did test this, and the compiler is fine with duplicate bounds. Not even a warning.

limemloh:
I will create another PR building on top of https://github.com/Concordium/concordium-rust-smart-contracts/pull/291

PR #287, opened 2023-05-26
Simplify the data layout in parameters.

abizjak:
## Purpose

Simplify the parameters by grouping them in such a way that the signature and key are first, and then the data to be signed.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/credential-registry-storage-contract/src/lib.rs
+++ b/examples/credential-registry-storage-contract/src/lib.rs
@@ -214,8 +214,8 @@ pub struct DataToSign {
 }
 
 /// Helper function that can be invoked at the front end to serialize
-/// the `SerializationHelperParam` to be signed by the wallet. The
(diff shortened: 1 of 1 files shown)


examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -236,7 +223,7 @@ pub struct SerializationHelperParam {
 #[receive(
```
DOBEN, line 223: Please update the comment which talks about `SerilaizationHelperParam`.

DOBEN (approved):

DOBEN (approved):

examples/credential-registry-storage-contract/src/lib.rs
abizjak: Fixed now. Good catch.

PR #289, opened 2023-05-31
Register and store credentials

annenkov:
## Purpose

Store encrypted credential data by calling the storage contract.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/credential-registry/src/lib.rs
+++ b/examples/credential-registry/src/lib.rs
@@ -686,7 +686,7 @@ fn contract_credential_status<S: HasStateApi>(
 #[receive(
     contract = "credential_registry",
     name = "registerCredential",
-    parameter = "CredentialInfo",
+    parameter = "RegisterCredentialParam",
     error = "ContractError",
     enable_logger,
     mutable
--- a/examples/credential-registry/src/lib.rs
+++ b/examples/credential-registry/src/lib.rs
@@ -546,14 +546,23 @@ impl Deserial for CredentialEvent {
 
 #[derive(Serialize, SchemaType)]
 pub struct InitParams {
+    /// The issuer's metadata.
     issuer_metadata: MetadataUrl,
+    /// An address of the credential storage contract.
     storage_address: ContractAddress,
+    /// Credential schemas available right after initialization.
+    #[concordium(size_length = 1)]
+    schemas:         Vec<(CredentialType, SchemaRef)>,
 }
 
 /// Init function that creates a fresh registry state given the issuer's
 /// metadata
 ///
-/// Logs `IssuerMetadataEvent`
+/// Logs `CredentialEvent::IssuerMetadata, CredentialEvent::Schema
+///
+/// It rejects if:
+///   - Fails to log the events.
+///   - Fails to add any of the inital schemas.
 #[init(
     contract = "credential_registry",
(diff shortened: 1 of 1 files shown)


examples/credential-registry/src/lib.rs
```diff
@@ -521,13 +544,19 @@ impl Deserial for CredentialEvent {
     }
 }
 
+#[derive(Serialize, SchemaType)]
+pub struct InitParams {
+    issuer_metadata: MetadataUrl,
+    storage_address: ContractAddress,
+}
```
abizjak, line 551: While we are at it we should also add 
```
#[concordium(size_length = 1)]
schemas: Vec<(CredentialType, SchemaRef)>,
```

to this since you have to add some schemas before issuing any credentials, and this makes setup easier.

abizjak: And of course then call `add_schemas`.

abizjak:
The storage stuff looks good, but I would like to add this additional parameter to init to simplify setup.

annenkov:
I've added schemas to the initialization parameters.

examples/credential-registry/src/lib.rs
```diff
@@ -17,16 +17,21 @@
 //! - update the credential metadata;
```
Bargsteen, line 17: ^ I see that this mentions the restore functionality. Didn't we agree not to have that?

examples/credential-registry/src/lib.rs
```diff
@@ -521,13 +544,28 @@ impl Deserial for CredentialEvent {
...
+    /// Credential schemas available right after initialization.
+    #[concordium(size_length = 1)]
+    schemas:         Vec<(CredentialType, SchemaRef)>,
+}
+
 /// Init function that creates a fresh registry state given the issuer's
-/// metadata
+/// metadata, storage contract address and initial credential schemas.
+///
+/// Logs `CredentialEvent::IssuerMetadata, CredentialEvent::Schema
```
Bargsteen, line 561:
```suggestion
/// Logs `CredentialEvent::IssuerMetadata` and `CredentialEvent::Schema`.
```

examples/credential-registry/src/lib.rs
```diff
@@ -569,6 +620,17 @@ pub struct CredentialInfo {
     metadata_url:     MetadataUrl,
 }
 
+/// Parameters for registering a credential and storing encrypted private data.
+#[derive(Serialize, SchemaType, Clone, Debug)]
+pub struct RegisterCredentialParam {
+    /// Public credential data.
+    credential_info: CredentialInfo,
+    /// Auxiliary data. In this contract is is used to pass encrypted private
```
Bargsteen, line 628:
```suggestion
    /// Auxiliary data. In this contract it is used to pass encrypted private
```

Bargsteen (approved):
LGTM, nice work!

abizjak (approved):

examples/credential-registry/src/lib.rs
annenkov: We agreed not to have "arbitrary" updates and keep the possibility to extend the credential lifecycle. So, it is currently possible for the issuer to restore a credential (flip back the revocation flag).

PR #296, opened 2023-06-09
Credential storage: rename metadata to version

annenkov:
## Purpose

The purpose of the `metadata` field is to store a version used by off-chain tools to decode the credential data. The name `version` is used in the Rust SDK. This PR renames `metadata` to `version` and updates the corresponding documentation for consistency.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/credential-registry-storage-contract/src/lib.rs
+++ b/examples/credential-registry-storage-contract/src/lib.rs
@@ -1,25 +1,23 @@
 //! The contract is used for storing verifiable credentials for the Web3Id
 //! infrastructure. The contract stores encrypted credentials (generated with
-//! the symmetric AES-256 encryption standard) with some metadata associated
-//! with them.
+//! the symmetric AES-256 encryption standard) and a version.
 //!
 //! The state contains a simple key-value store where each public key
 //! can store its associated encrypted credential so that the credential can be
 //! recovered by off-chain wallets. The `keys` for the store are the Ed25519
-//! public keys and the `values` are the `metadata + encrypted_credentials`. The
-//! metadata associated with the credential contains information needed for
-//! decoding the encrypted credential. Only if an off-chain tool has access to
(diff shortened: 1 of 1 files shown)


examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -51,16 +49,16 @@ use concordium_std::*;
 
 const SIGNATURE_DOMAIN: &str = "WEB3ID:STORE";
 
-/// Metadata is a string of two bytes that contain information needed for
+/// Version is a string of two bytes that contain information needed for
 /// decoding the credentials.
 #[derive(Serial, SchemaType, Clone, Copy, Deserial, PartialEq, Debug)]
-struct Metadata(u16);
+struct Version(u16);
```
abizjak, line 55: Just make this into a type alias instead
```
type Version = u16
```

abizjak (approved):

examples/credential-registry-storage-contract/src/lib.rs
```diff
@@ -51,16 +49,16 @@ use concordium_std::*;
 
 const SIGNATURE_DOMAIN: &str = "WEB3ID:STORE";
 
-/// Metadata is a string of two bytes that contain information needed for
+/// Version is a string of two bytes that contain information needed for
```
DOBEN, line 52: The version will be a number, isn't it? The current comment is a leftover from the `mySome.id` repo where the metadata was a string of two bytes of the `platform` e.g. `LI` for `LinkedIn`.

We could update the comment above a bit, e.g. `Version number to decode the credentials`.

DOBEN (approved):

examples/credential-registry-storage-contract/src/lib.rs
annenkov: I updated the comment and also fixed a test, making it more clear that it's a number and not an encoding of two ASCII chars.

PR #297, opened 2023-06-09
Use transparent attribute for deriving SchemaType

### LINKED ISSUE #285: Provide a transparent attribute similar to serde's for SchemaType
A common design pattern in Rust is to wrap other types in a struct, to work around the rules related to trait coherence (like the orphan rule).

Deriving the `SchemaType` for such a type exposes the struct wrapping the type, which, in most cases, is not intended to be exposed to the user, since it makes the input more verbose, and the developer have to implement it manually to remove this.

Provide an attribute `#[concordium(transparent)]` for the derive macro, instructing it to hide the outer struct in the implementation of SchemaType.
```rust
#[derive(Debug, Serialize, SchemaType)]
#[concordium(transparent)]
pub struct OperatorOfQueryParams {
    /// List of operatorOf queries.
    #[concordium(size_length = 2)]
    pub queries: Vec<OperatorOfQuery>,
}
```

This should fail at compile time for anything other than structs containing a single field.

limemloh:
## Purpose

Closes #285.

## Changes

- Use `#[concordium(transparent)]` for a number of types in `concordium-cis2`.
- Add testcase for deriving SchemaType with transparent.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std/tests/derive-schema-type/success-transparent.rs
+++ b/concordium-std/tests/derive-schema-type/success-transparent.rs
@@ -1,16 +1,26 @@
 /// Test ensuring derivation works for newtype wrappers with transparent
 use concordium_std::*;
 
+/// Simple struct
 #[derive(SchemaType)]
 #[concordium(transparent)]
 struct MyStruct {
     field: u32,
 }
 
+/// Nested struct
 #[derive(SchemaType)]
 #[concordium(transparent)]
 struct MyOtherStruct {
     field: MyStruct,
+}
+
+/// With field attributes
+#[derive(SchemaType)]
+#[concordium(transparent)]
+struct MyCollection {
+    #[concordium(size_length = 1)]
+    field: Vec<u32>,
 }
 
 fn main() {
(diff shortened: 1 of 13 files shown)


concordium-std/tests/derive-schema-type/success-transparent.rs
```diff
@@ -0,0 +1,26 @@
+/// Test ensuring derivation works for newtype wrappers with transparent
+use concordium_std::*;
+
+#[derive(SchemaType)]
+#[concordium(transparent)]
+struct MyStruct {
+    field: u32,
+}
```
abizjak, line 8: Please add a test with a container and `size_length`.

abizjak (approved):

concordium-std/tests/derive-schema-type/success-transparent.rs
limemloh: Done

Bargsteen (approved):
LGTM

Issue #299, opened 2023-06-12
Tag attribute for specifying tags when deriving serialization


limemloh:
The schema supports setting the tag for enums explicitly, but there is no way to provide this information when deriving the traits `SchemaType`, `Serial` and `Deserial`, meaning the developer has to implement these manually.

We should support an attribute `tag` for enum variants specifying the tag to use. 

Something like `#[concordium(tag = <number>)]` where `<number>` is replaced with the value to use for the tag.
```rust
#[derive(SchemaType, Serialize)]
enum MyType {
  #[concordium(tag = 42)]
  First(...),
  Second(...)
}
```

It should use the tag 42 just the variant `First` and any variants without explicit tags should use the usual incrementing tag, which would be 1 for `Second`, meaning the tag only result in changes for the specified variant.

The macro should also ensure all variants are uniquely tagged.

Additionally, the developer can add a `#[concordium(repr(x))` (where `x` is some unsigned integer type like `u8`) at the top of the type, which will set the size used for each of the tags. When not set, the behavior is to use the smallest size possible for representing all the variants.
Currently, using `repr(x)` cannot be expressed in the schema for most of the cases and a new version of `SchemaType` is needed.

PR #301, opened 2023-06-16
Test cases for tag attribute in derive serialization

limemloh:
## Purpose

Closes https://github.com/Concordium/concordium-rust-smart-contracts/issues/299.
Should be merged after https://github.com/Concordium/concordium-contracts-common/pull/95

## Changes

- Added a number of test cases for serialization when using `tag` attribute.
- Derive serialization for `Cis2Event`

## Checklist

      hard-to-understand areas.


concordium-cis2/src/lib.rs
```diff
@@ -854,51 +861,6 @@ impl<T: IsTokenId, A: IsTokenAmount> schema::SchemaType for Cis2Event<T, A> {
     }
 }
 
-impl<T: IsTokenId, A: IsTokenAmount> Serial for Cis2Event<T, A> {
-    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
-        match self {
-            Cis2Event::Transfer(event) => {
-                out.write_u8(TRANSFER_EVENT_TAG)?;
```
abizjak, line 861: Could you not clean up the constants as well?

concordium-std/tests/derive-deserial-with-state/fail-tag-collide-with-inferred.rs
```diff
@@ -0,0 +1,14 @@
+//! Ensure `derive(DeserialWithState)` fails when 'tag' attribute are collides
```
abizjak, line 1: "are collides"

concordium-std/tests/derive-serial/fail-tag-without-repr.stderr
```diff
@@ -0,0 +1,7 @@
+error: 'repr(..)' attribute must be set on the type when using 'tag' attribute
+ --> tests/derive-serial/fail-tag-without-repr.rs:5:10
+  |
+5 | #[derive(Serial)]
+  |          ^^^^^^
+  |
+  = note: this error originates in the derive macro `Serial` (in Nightly builds, run with -Z macro-backtrace for more info)
```
abizjak, line 7: These exact error messages seems less than ideal. Any simple change in the compiler, or adding comments to the test, will invalidate them since line numbers will change.

Is that the only way to test this?

abizjak (approved):
I think you can also remove the manually written serialization in the credential registry contract. And the sponsored transaction example.

abizjak (commented):
Also, please update the changelog.

The functionality is re-exported, so we need to make this new functionality visible to users of concordium-std.

concordium-cis2/src/lib.rs
limemloh: You mean remove them? Unfortunately, we still have some manual implementations left where the constants are useful.

abizjak: Ok then. :)

concordium-std/tests/derive-serial/fail-tag-without-repr.stderr
limemloh: I agree that this is not ideal, but this is the workflow that `trybuild` advocate. Changes in the compiler seems like something we would want to notice, like if something else fails before this point, but yes changing the tests will mean updating these files. However, these files are generated, so it is not a big deal to update them.

abizjak: Ok.

abizjak (approved):

PR #308, opened 2023-06-29
Single credential type per registry

annenkov:
## Purpose

Modify the credential registry example so it stores public data of verifiable credentials of one type.
This type and the corresponding schema are specified during contract initialization.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/credential-registry/src/lib.rs
+++ b/examples/credential-registry/src/lib.rs
@@ -602,6 +602,7 @@ pub struct InitParams {
 /// `CredentialEvent::Schema` for each schema in the input.
 ///
 /// It rejects if:
+///   - Fails to parse the input parameter.
 ///   - Fails to log the events.
 ///   - Fails to register any of the initial revocation keys.
 #[init(
--- a/examples/credential-registry/src/lib.rs
+++ b/examples/credential-registry/src/lib.rs
@@ -1275,18 +1275,30 @@ fn contract_revocation_keys<S: HasStateApi>(
     Ok(host.state().view_revocation_keys())
 }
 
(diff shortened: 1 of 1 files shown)


abizjak (approved):

examples/credential-registry/src/lib.rs
```diff
@@ -96,13 +91,12 @@ pub struct CredentialEntry<S: HasStateApi> {
     revocation_nonce: u64,
     /// Revocation flag
     revoked:          bool,
-    /// Schema reference and credential metadata.
-    /// This data is only needed when credential info is requested. In other
-    /// operations, `StateBox` defers loading credential data.
-    credential_data:  StateBox<CredentialData, S>,
+    /// Metadata URL of the credential (not to be confused with the metadata URL
+    /// of the **issuer**).
+    metadata_url:     MetadataUrl,
```
DOBEN, line 96: We can put the `MetaDataUrl` in a StateBox as well, so it is lazy loaded.

examples/credential-registry/src/lib.rs
```diff
@@ -635,7 +597,6 @@ pub struct InitParams {
 /// It rejects if:
 ///   - Fails to log the events.
```
DOBEN, line 598: - Fails to parse the input parameter.

This is an additional reason why the function invoke could reject.

DOBEN (approved):
LGTM

examples/credential-registry/src/lib.rs
annenkov: Ok, I'll wrap it in `StateBox`

annenkov: Done.

PR #318, opened 2023-07-14
Replace `public_key_registry` from `cis3_sponsored_transaction` with corresponding smart contract features

### LINKED ISSUE #258: Update sponsored transaction dApp once public keys are exposed within the smart contract code
**Task description**

- [x] Update smart contract (e.g. remove `public_key_registry` and related comments):
https://github.com/Concordium/concordium-rust-smart-contracts/tree/main/examples/cis3-nft-sponsored-txs
- [x] Update dApp frontend example to not have a tab anymore to register a `publicKey` via the frontend:
 https://github.com/Concordium/concordium-dapp-examples/tree/main/sponsoredTransactions
- [x] Sponsored transaction tutorial needs to be updated to remove comments about `public_key_registry`:
https://github.com/Concordium/concordium.github.io/pull/786
- [x] Update CIS3 standard to reflect the serialization of the new input parameter to the `permit` function.

DOBEN:
## Purpose

addresses #258 

After the new protocol update P6, smart contracts can now get the `public_keys` of accounts within the smart contract code. There is also a new `check_account_signature` function available. 

The `public_key_registry` is not necessary anymore in the sponsored transaction smart contract and should be replaced.

Note: This PR is a draft and has some local dependencies hardcoded to simplify the use of a "local/unreleased" smart contract testing integration library. This needs to be updated/changed before merging.
https://github.com/Concordium/concordium-smart-contract-tools/tree/main/concordium-smart-contract-testing
 
## Changes

- Remove `public_key_registry`.
- Add `check_account_signature` to verify the signature in the smart contract.
- Replace old `publicKeyOf` function (returned a tuple pair (public key, nonce)) with two separate functions `publicKeyOf` (returns public keys for a vector of accounts) and `nonceOf` (returns nonces for a vector of accounts).
- Add integration tests.
- Add integration testings for a signature transfer permit function call.
- Add integration testings for a signature update operator permit function call.
- `Signature/PublicKey` types were removed from `concordium_std` and are now available via `concordium_contracts_common`.

### DIFF
--- a/examples/cis3-nft-sponsored-txs/Cargo.toml
+++ b/examples/cis3-nft-sponsored-txs/Cargo.toml
@@ -15,6 +15,10 @@ crypto-primitives = ["concordium-std/crypto-primitives"]
 concordium-std = {path = "../../concordium-std", default-features = false}
 concordium-cis2 = {path = "../../concordium-cis2", default-features = false}
 
+[dev-dependencies]
+concordium-smart-contract-testing = {path = "../../../../LocalTestingLibrary/concordium-smart-contract-tools/concordium-smart-contract-testing"}
+ed25519-dalek = "1.0.1"
+
 [lib]
 crate-type=["cdylib", "rlib"]
 
--- a/examples/cis3-nft-sponsored-txs/tests/test.rs
+++ b/examples/cis3-nft-sponsored-txs/tests/test.rs
@@ -1,0 +1,625 @@
+//! This module contains integration tests for the cis3-sponsored-transaction
+//! contract. To run them, use `cargo test`.
+use cis3_nft_sponsored_txs::{
+    ContractTokenAmount, ContractTokenId, MintParams, NonceEvent, NonceOfQueryResponse,
+    PermitMessage, PermitParam, PublicKeyOfQueryResponse, VecOfAccountAddresses, NONCE_EVENT_TAG,
+};
+use concordium_cis2::{TokenIdU32, *};
+use concordium_smart_contract_testing::{AccountAccessStructure, *};
+use concordium_std::{
+    AccountPublicKeys, AccountSignatures, CredentialSignatures, PublicKey, SignatureEd25519,
+    Timestamp,
+};
+use std::collections::BTreeMap;
+
+const ACC_ADDR_OWNER: AccountAddress = AccountAddress([0u8; 32]);
+const ADDR_OWNER: Address = Address::Account(ACC_ADDR_OWNER);
+const ACC_ADDR_OTHER: AccountAddress = AccountAddress([1u8; 32]);
+const ADDR_OTHER: Address = Address::Account(ACC_ADDR_OTHER);
+const NON_EXISTING_ACCOUNT: AccountAddress = AccountAddress([2u8; 32]);
+const ACC_INITIAL_BALANCE: Amount = Amount::from_ccd(1000);
+
+const TOKEN_1: TokenIdU32 = TokenIdU32(1);
+
+// Private key: 8ECA45107A878FB879B84401084B55AD4919FC0F7D14E8915D8A5989B1AE1C01
+const PUBLIC_KEY: [u8; 32] = [
+    120, 154, 141, 6, 248, 239, 77, 224, 80, 62, 139, 136, 211, 204, 105, 208, 26, 11, 2, 208, 195,
+    253, 29, 192, 126, 199, 208, 39, 69, 4, 246, 32,
+];
+
+const SIGNATURE_TRANSFER: SignatureEd25519 = SignatureEd25519([
+    46, 96, 133, 143, 232, 24, 149, 54, 217, 245, 162, 135, 64, 125, 32, 61, 209, 147, 240, 151,
+    19, 244, 137, 244, 91, 59, 120, 202, 39, 201, 82, 39, 64, 210, 250, 183, 187, 27, 147, 50, 31,
+    88, 78, 79, 78, 135, 192, 72, 38, 234, 90, 226, 89, 75, 124, 86, 1, 190, 196, 195, 248, 19,
+    181, 11,
+]);
+
+const SIGNATURE_UPDATE_OPERATOR: SignatureEd25519 = SignatureEd25519([
+    50, 83, 36, 119, 75, 223, 109, 182, 152, 68, 46, 46, 249, 11, 169, 109, 51, 211, 47, 56, 60,
+    155, 62, 59, 70, 8, 107, 72, 147, 141, 226, 153, 43, 234, 82, 164, 168, 43, 202, 244, 213, 95,
+    101, 101, 155, 226, 130, 77, 47, 243, 161, 219, 100, 208, 106, 109, 102, 189, 171, 159, 22,
+    179, 148, 0,
+]);
+
+/// A helper method for setting up:
+///  - The chain,
+///  - The two accounts `ACC_ADDR_OWNER` and `ACC_ADDR_OTHER`, both with
+///    `ACC_INITIAL_BALANCE` CCD. `ACC_ADDR_OTHER` will be set up with valid
+///    public keys so it can generate signatures.
+///  - The deployed and initialized contract.
+fn setup_chain_and_contract() -> (Chain, ContractInitSuccess) {
+    let mut chain = Chain::new();
+
+    let balance = AccountBalance {
+        total:  ACC_INITIAL_BALANCE,
+        staked: Amount::zero(),
+        locked: Amount::zero(),
+    };
+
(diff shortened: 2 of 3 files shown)


examples/cis3-nft-sponsored-txs/Cargo.toml
```diff
@@ -14,6 +14,10 @@ crypto-primitives = ["concordium-std/crypto-primitives"]
 [dependencies]
 concordium-std = {path = "../../concordium-std", default-features = false}
 concordium-cis2 = {path = "../../concordium-cis2", default-features = false}
+ed25519-dalek = "1.0.1"
```
abizjak, line 17: I don't think we should need this. If we need it for integration testing then it should go under dev-dependencies I think.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -382,6 +360,16 @@ type ContractError = Cis2Error<CustomContractError>;
 
 type ContractResult<A> = Result<A, ContractError>;
 
+/// Mapping CustomContractError to ContractError
+impl From<CheckAccountSignatureError> for CustomContractError {
```
abizjak, line 364: The reason that you have to unwrap below is that your contract error type is `ContractError`, which is 
```
Cis2Error<CustomContractError>;
```
So you want a 
```
From<CheckAccountSignatureError> for Cis2Error<CustomContractError>
```

as well.

I think we should add this to concordium-std, like we have these from instances 
```
impl<X> From<NewReceiveNameError> for Cis2Error<X>
where
    X: From<NewReceiveNameError>,
{
    #[inline]
    /// Converts the error by wrapping it in [Self::Custom].
    fn from(err: NewReceiveNameError) -> Self { Cis2Error::Custom(X::from(err)) }
}
```
already.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -1160,35 +1085,35 @@ pub struct PublicKeyQuery {
...
-    Ok(result)
-}
+// #[receive(
+//     contract = "cis3_nft",
+//     name = "publicKeyOf",
+//     parameter = "ContractPublicKeyQueryParams",
+//     return_value = "ContractPublicKeyQueryResponse",
+//     error = "ContractError"
+// )]
+// fn contract_public_key_of<S: HasStateApi>(
```
abizjak, line 1095: What is the intention here?

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: ok, It should be added to `concordium-cis2` library instead of `concordium-std`. I added it.

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: I replaced this function with the `publicKeyOf` function and the `nonceOf` function.

examples/cis3-nft-sponsored-txs/tests/test.rs
```diff
@@ -0,0 +1,643 @@
...
+        CredentialIndex {
+            index: 0u8,
+        },
+        credential_public_keys,
+    );
+
+    let keys = AccountAccessStructure {
+        keys:      key_map,
+        threshold: AccountThreshold::ONE,
+    };
```
abizjak, line 86:
```suggestion
let rng = &mut rand::thread_rng();
let keypairs = AccountKeys::singleton(rng);
let keys: AccountAccessStructure = (&keypairs).into();
```

examples/cis3-nft-sponsored-txs/tests/test.rs
```diff
@@ -0,0 +1,643 @@
...
+    let payload = UpdateOperatorParams(vec![update_operator]);
+
+    let mut inner_signature_map = BTreeMap::new();
+    inner_signature_map.insert(0u8, concordium_std::Signature::Ed25519(SIGNATURE_UPDATE_OPERATOR));
+
+    let mut signature_map = BTreeMap::new();
+    signature_map.insert(0u8, CredentialSignatures {
+        sigs: inner_signature_map,
+    });
```
abizjak, line 149: If we unify AccountSignature and TransactionSignature this will become
```
let signature = keypairs.sign(&to_bytes(...))
```

examples/cis3-nft-sponsored-txs/tests/test.rs
```diff
@@ -0,0 +1,643 @@
...
+        threshold: SignatureThreshold::ONE,
+    };
+
+    let mut key_map: BTreeMap<u8, concordium_std::CredentialPublicKeys> = BTreeMap::new();
+    key_map.insert(0u8, credential_public_keys);
+
+    let account_public_keys: AccountPublicKeys = AccountPublicKeys {
+        keys:      key_map,
+        threshold: AccountThreshold::ONE,
+    };
```
abizjak, line 282:
```suggestion
let account_public_keys = (&keypairs).into();
```

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -28,35 +28,32 @@
...
 //! As follows from the CIS3 specification, the contract has a `permit`
-//! function. It is the sponsored counterparts to the `transfer/updateOperator`
+//! function. It is the sponsored counterpart to the `transfer/updateOperator`
 //! function and can be executed by anyone on behalf of an account given a
-//! signed message by the private key corresponding to the public key that
-//! is registered for that account.
+//! signed message.
+//!
+//! Concordium supports natively multi-sig accounts. Each account address on
+//! Concordium is controlled by one (or several) credentials (real-world
```
limemloh, line 42: Reading this sentence skipping `(or several)` is off, so I suggest either:
```suggestion
//! Concordium is controlled by one or several credentials (real-world
```
or
```suggestion
//! Concordium is controlled by one (or several) credential(s) (real-world
```

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -28,35 +28,32 @@
...
+//! identities) and each credential has one (or several) public-private key
```
limemloh, line 43: Similar here, see my other comments

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -28,35 +28,32 @@
...
+//! pairs. The key/signature structures of an account are represented as
```
limemloh, line 44:
```suggestion
//! pairs. Both the key and signature structures of an account are represented as
```

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -588,81 +561,18 @@ fn contract_view<S: HasStateApi>(
 
     let all_tokens = state.all_tokens.iter().map(|x| *x).collect();
 
-    let mut all_public_keys_and_nonces = Vec::new();
-    for (account_address, public_key) in state.public_key_registry.iter() {
-        all_public_keys_and_nonces.push((*account_address, *public_key));
+    let mut all_nonces = Vec::new();
+    for (account_address, nonce) in state.nonces_registry.iter() {
+        all_nonces.push((*account_address, *nonce));
     }
```
limemloh, line 567: I think you could do this:
```suggestion
    let all_nonces: Vec<_> = state.nonces_registry.iter().copied().collect();
```

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -905,35 +814,30 @@ fn contract_permit<S: HasStateApi>(
...
-        param.signature.len() == 1
-            && param.signature.get(&0).ok_or(CustomContractError::SignatureMapMisconfigured)?.len()
+        param.signature.sigs.len() == 1
+            && param
+                .signature
+                .sigs
+                .get(&0)
+                .ok_or(CustomContractError::SignatureMapMisconfigured)?
+                .sigs
+                .len()
```
limemloh, line 825: Consider introducing a variable for this and use the variable name to help the reader understand what this is

limemloh (approved):
The first version only supported accounts with one key, and the reason I remember for this was to simplify the endpoints for registering keys. So now that we are removing those endpoints, I think we should support multi-sig accounts here

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: This check is for making sure that the signature is at index 0 in both signature maps. Since we now support multi-sig accounts, this check was removed altogether.

DOBEN:
close as completed in 
https://github.com/Concordium/concordium-rust-smart-contracts/pull/363
https://github.com/Concordium/concordium-rust-smart-contracts/pull/359

PR #321, opened 2023-07-21
Contract state is no longer required to implement `StateClone` for unit testing

target-san:
## Purpose

Remove `StateClone` trait, which was needed only to clone toplevel contract state during unit testing

## Changes

* `TestHost` implementation no longer imposes `StateClone` constraint onto state type
* State is cloned by deserializing it from storage
* `StateClone` is removed ([see comment](https://github.com/Concordium/concordium-rust-smart-contracts/pull/321#issuecomment-1653574143))

## Checklist

      hard-to-understand areas.

## CLA acceptance

- [x] I accept the above linked CLA.


target-san:
One thing which I didn't touch is whether `TestHost::with_rollback` should be modified to perform actual operation on cloned state and leave original state as-is, just in case

abizjak:
Hi. thanks for the effort.

We were planning to deprecate the entire test infrastructure as it is now in favor of the [testing library](https://crates.io/crates/concordium-smart-contract-testing) and with that we would remove in particular the StateClone, but also the odd and hard to explain "state_parameter = S" annotation on the macros, and a few generics on entrypoints.

I would prefer to do that in one big release, rather than deprecating/removing things one at a time.

Is this blocking you?

If it is, I don't think deprecating is the right approach. People should not be writing these by hand anyhow, so we should just remove the StateClone trait entirely.

target-san:
Well, this test infra is kind of useful in some scenarios 'cause it allows debugging contract directly.

Nope, it's not blocking me or anything. Just had some spare time and inspiration to drop it. Thought that soft deprecation is better, as it won't break other users' code. Shall I remove `StateClone` completely?

abizjak:
> Well, this test infra is kind of useful in some scenarios 'cause it allows debugging contract directly.

Sure, but the issue is that it adds a lot of complexity, both in terms of the way the contracts look (with extra generics), and with the amount of code we need to maintain (for example the entire implementation of test infrastructure, including its own state implementation). So our conclusion is that iti s not worth it.

> Nope, it's not blocking me or anything. Just had some spare time and inspiration to drop it. Thought that soft deprecation is better, as it won't break other users' code. Shall I remove `StateClone` completely?

Yes, that'd be my preference.

target-san:
I dropped `StateClone` completely, both in this repo and in `concordium-contracts-common`: please check [PR 104](https://github.com/Concordium/concordium-contracts-common/pull/104)
Unfortunately this one's incomplete, as I'll have to fix smart contract templates checking on CI

target-san:
Ok, so `StateClone` is now completely removed, and all checks are fixed.
Plz note that [PR 104](https://github.com/Concordium/concordium-contracts-common/pull/104) in order to set proper submodule reference

.github/workflows/linter.yml
```diff
@@ -107,7 +107,10 @@ jobs:
       # Run all tests, including doc tests.
       - name: Run cargo test
         run: |
+          TEMPLATE_DIR=`pwd`
           mv $PROJECT_NAME ${{ runner.temp }}/
+          sed -i "s/{version = \"7.0\", default-features = false}/{path = \"${TEMPLATE_DIR//\//\\\/}\/concordium-std\", default-features = false}/g" ${{ runner.temp }}/$PROJECT_NAME/Cargo.toml
+          sed -i "s/{version = \"4.0\", default-features = false}/{path = \"${TEMPLATE_DIR//\//\\\/}\/concordium-cis2\", default-features = false}/g" ${{ runner.temp }}/$PROJECT_NAME/Cargo.toml
```
abizjak, line 113: I don't think this is what we want.

Since `cargo concordium init` will take the template from the repository, we must always ensure that the template that exists works with the published versions of the libraries.

I realize that this complicates things, and it's something we should improve #331

abizjak (approved):
Looks good.

Sorry for the tardiness here as well. This will be released on Monday.

I have merged it into #330

PR #324, opened 2023-08-03
Add deploy script projects to templates

DOBEN:
## Purpose

Addresses  https://github.com/Concordium/concordium-smart-contract-tools/issues/117

This PR addresses the first part: `Automated flow`

When creating a new smart contract project with the `cargo-generate` command using the `default` template, we provide now a `deploy-scripts` folder that contains some boilerplate code to deploy/initialize/update the contracts on chain. The scripts can be run with e.g. this command.

```
cargo run -- --node http://node.testnet.concordium.com:20000 --account ./3PXwJYYPf6fyVb4GJquxSZU8puxrHfzc4XogdMVot8MUQK53tW.export --modules ./default.wasm.v1 --modules ./default2.wasm.v1
```

## Changes

- Adding deploy script projects to the `default` templates


templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,427 @@
...
+    /// expiry time.
+    pub async fn update_contract(
+        &self,
+        update_payload: UpdateContractPayload,
+        energy: Option<GivenEnergy>,
+        expiry: Option<TransactionTime>,
+    ) -> Result<HashBytes<TransactionMarker>, DeployError> {
+        println!();
+
+        println!("Updating contract....");
```
Bargsteen, line 210:
```suggestion
        println!("\nUpdating contract....");
```

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,427 @@
...
+    /// be given. If `None` is provided, 50000 energy is used as a default
+    /// max_energy value.
+    pub async fn estimate_energy(
+        &self,
+        payload: UpdateContractPayload,
+        max_energy: Option<Energy>,
+    ) -> Result<Energy, DeployError> {
+        println!();
+
+        println!("Estimating energy....");
```
Bargsteen, line 274:
```suggestion
        println!("\nEstimating energy....");
```

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,427 @@
...
+                    )))
+                }
+                AccountTransactionEffects::ContractInitialized {
+                    data,
+                } => Ok(ContractInitialized {
+                    contract: data.address,
+                }),
+                _ => Err(DeployError::InvalidBlockItem("Invalid transaction effects".into())),
+            },
+            _ => Err(DeployError::InvalidBlockItem("Expected Account transaction".into())),
```
Bargsteen, line 397: These errors do not help the user that much. Could we elaborate a bit on what might have gone wrong?

Bargsteen: Same below.

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+async fn main() -> Result<(), DeployError> {
+    let app: App = App::parse();
+
+    let concordium_client = v2::Client::new(app.url).await?;
+
+    let deployer = Deployer::new(concordium_client, &app.key_file)?;
+
+    let mut modules_deployed: Vec<ModuleDeployed> = Vec::new();
+
+    for contract in app.modules {
```
Bargsteen, line 117: We could check for duplicates up front.

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    let update_payload = transactions::UpdateContractPayload {
+        amount:       Amount::from_ccd(0),
+        address:      contract,
+        receive_name: OwnedReceiveName::new_unchecked("default.receive".to_string()),
+        message:      bytes.try_into()?,
+    }; // Example
+
+    let mut energy = deployer.estimate_energy(update_payload.clone(), None).await?; // Example
+
+    // We add 100 energy to be save.
```
Bargsteen, line 152:
```suggestion
    // We add 100 energy to be safe.
```

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    energy.energy += 100; // Example
```
Bargsteen, line 153: Adding some energy seems reasonable, so I think we should recommend it in the docs of `estimate_energy` 😊

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    let wasm_module = std::fs::read(file).context("Could not read the WASM file")?;
+    let mut cursor = Cursor::new(wasm_module);
+    let wasm_module: WasmModule = concordium_rust_sdk::common::from_bytes(&mut cursor)?;
+    Ok(wasm_module)
+}
+
+/// Command line flags.
+#[derive(clap::Parser, Debug)]
+#[clap(author, version, about)]
+struct App {
```
Bargsteen, line 81: Perhaps we should add subcommands so that you have to call `cargo run -- deploy ...` or `cargo run -- update` etc?

Bargsteen: That would make it easier for developers to extend this boilerplate with more functions.

Bargsteen (approved):
Nice work @DOBEN! I have some suggestions and found a few smaller issues, but nothing blocking ;)

templates/default/deploy-scripts/README.md
```diff
@@ -0,0 +1,112 @@
...
+- `deploy_wasm_module`: To deploy a new smart contract module on chain.
+- `init_contract`: To initialize a smart contract instance on chain.
+- `update_contract`: To update a smart contract instance on chain.
+
+Event parsing helper functions:
+- `parse_deploy_module_event`: To parse the chain events after deploying a module.
+- `parse_contract_init_event`: To parse the chain events after initialization of a smart contract instance.
+- `parse_contract_update_event`: To parse the chain events after updating a smart contract instance.
+
+The `main.rs` file has a section (marked with `// Write your own deployment/initialization script below. An example is given here.`) that you should replace with your custom logic to deploy, initialize, and update your smart contract protocol.
```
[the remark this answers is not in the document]
DOBEN: What I meant here was that you can execute "deploy", "initialize", and contract "update" transactions. Meaning all three types of transactions are supported.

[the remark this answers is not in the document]
DOBEN: I updated the comment to:
```
The `main.rs` file has a section (marked with `// Write your own deployment/initialization script below. An example is given here.`) that you should replace with your custom logic. You can write your script using `deploy`, `initialize`, and contract `update` transactions.
```

templates/default/deploy-scripts/src/main.rs
Bargsteen: I meant the latter, but I see that my examples were ambiguous. 
Ok, you've convinced me that the current approach is sufficient ;)

templates/default/deploy-scripts/src/deployer.rs
DOBEN: I am using the following error text now:


```
The parsed account transaction effect should be of type `ContractInitialized` or `None` (in case the transaction reverted)
```
                     
and 
```
Can only parse an account transaction (no account creation transaction or chain update transaction)
```

templates/default/deploy-scripts/README.md
```diff
@@ -0,0 +1,112 @@
...
+- `parse_contract_init_event`: To parse the chain events after initialization of a smart contract instance.
+- `parse_contract_update_event`: To parse the chain events after updating a smart contract instance.
+
+The `main.rs` file has a section (marked with `// Write your own deployment/initialization script below. An example is given here.`) that you should replace with your custom logic. You can write your script using `deploy`, `initialize`, and contract `update` transactions.
+
+# Running the Example
+
+The `main.rs` file has a section (marked with `// Write your own deployment/initialization script below. An example is given here.`) that provides an example that you can run.
+
+ATTENTION: You have to have created a smart contract with the name `default` to run the given example. This can be done by answering `default` to the two questions when creating the project via `cargo-generate`.
```
limemloh, line 83: Could we use the name directly from the template instead of this message. I assume when the user reads this, they already called cargo-generate

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,459 @@
...
+/// A struct representing a smart contract instance on the chain.
+#[derive(Clone, Debug)]
+pub struct ContractInitialized {
+    /// Smart contract address on the chain.
+    pub contract: ContractAddress,
+}
+
+/// A struct containing connection and wallet information.
+#[derive(Debug)]
+pub struct Deployer {
```
limemloh, line 40: I am not sure that this struct makes a lot of sense in the first place, maybe just provide a bunch of functions which takes the client and keys as well

templates/default/deploy-scripts/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+```
+cargo generate --git https://github.com/Concordium/concordium-rust-smart-contracts.git
+```
+
+Any of the two commands will work and will give you several templates to choose from. 
+
+- Choose the `templates/default` and answer the questions to complete the setup process (answer `default` for the two questions if you want to run the below example).
+
+At the end, you will have a Rust project setup with this boilerplate code included.
```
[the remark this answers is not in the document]
DOBEN: Yes, that makes sense. I will move this `setup` to the developer documentation and remove it from this Read.me.

templates/default/deploy-scripts/src/deployer.rs
DOBEN: I agree that currently, this struct does not hold a lot of information but it could contain many more settings in the future when we add additional functionalities. I see its value in combining/bundling all these setup settings. E.g. possible additional settings that this struct can hold: 

- several account keys to deploy with different accounts
- different network settings 
- dry-run settings (e.g. the script first dry-runs the deployment before executing it live on mainnet)

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,459 @@
...
+                            "Expected transaction type to be DeployModule".into(),
+                        ));
+                    }
+
+                    match reject_reason {
+                        RejectReason::ModuleHashAlreadyExists {
+                            contents,
+                        } => Ok(ModuleDeployed {
+                            module_ref: contents,
+                        }),
```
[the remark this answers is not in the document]
DOBEN: Removed it.

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,163 @@
...
+
+        let (_, module) = deployer.deploy_wasm_module(wasm_module, None).await?;
+
+        modules_deployed.push(module);
+    }
+
+    // Write your own deployment/initialization script below. An example is given
+    // here.
+
+    let param: OwnedParameter = OwnedParameter::default(); // Example
```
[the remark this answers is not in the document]
DOBEN: I import the input parameter `MyInputType` to showcase how this can be done now. 

I am not sure if events and error parsing are done in a deployment script commonly. Usually, such scripts are not meant to fail and we have these `check_outcome_of_update_transaction` functions that can be used to check for failed transactions already (with error codes).

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,170 @@
...
+use deployer::{Deployer, ModuleDeployed};
+use hex::FromHexError;
+use std::{
+    io::Cursor,
+    path::{Path, PathBuf},
+};
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum DeployError {
```
abizjak, line 30: Since this is an application, is it important to have all of these refined errors?

Would it not be enough to just use `anyhow::Error` everywhere?

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,170 @@
...
+                --modules ./myPath/default.wasm.v1 --modules ./default2.wasm.v1)."
+    )]
+    modules:  Vec<PathBuf>,
+}
+
+/// Main function: It deploys to chain all wasm modules from the command line
+/// `--modules` flags. Write your own custom deployment/initialization script in
+/// this function. An deployment/initialization script example is given in this
+/// function for the `default` smart contract.
+#[tokio::main(flavor = "current_thread")]
```
abizjak, line 108: Why do you choose this flavor over the defaults? Unless there is a reason to do it I would just use `tokio::main` without attributes.

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        let (_, module) = deployer.deploy_wasm_module(wasm_module, None).await?;
+
+        modules_deployed.push(module);
+    }
+
+    // Write your own deployment/initialization script below. An example is given
+    // here.
+
+    let param: OwnedParameter = OwnedParameter::default(); // Example
```
abizjak, line 129: Use `empty` instead of `default` to be more explicit.

templates/default/deploy-scripts/src/main.rs
abizjak: That would make it simpler to write and simpler to look at as well.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// If successful, the transaction energy is returned by this function.
+    /// This function can be used to dry-run a transaction.
+    ///
+    /// An optional max_energy value for the transaction can
+    /// be given. If `None` is provided, 50000 energy is used as a default
+    /// max_energy value.
+    pub async fn estimate_energy(
+        &mut self,
+        payload: UpdateContractPayload,
+        max_energy: Option<Energy>,
```
abizjak, line 275:
```suggestion
```

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,453 @@
...
+            } => Err(DeployError::InvokeContractFailed(format!(
+                "Contract invoke failed: {reason:?}, used_energy={used_energy}, return \
+                 value={return_value:?}"
+            ))),
+            InvokeContractResult::Success {
+                return_value: _,
+                events: _,
+                used_energy,
+            } => {
+                let e = used_energy.energy;
```
abizjak, line 312: Not sure why you are unwrapping it here.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,453 @@
...
+                println!("Contract invoke success: estimated_energy={e}");
+                Ok(Energy {
+                    energy: e,
```
abizjak, line 315: And then wrapping it up.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,453 @@
...
+            .await
+            .map_err(DeployError::TransactionRejected)?;
+
+        println!("Sent tx: {tx_hash}");
+
+        let (_, block_item) = self.client.wait_until_finalized(&tx_hash).await?;
+
+        self.check_outcome_of_update_transaction(block_item)?;
+
+        println!("Transaction finalized, tx_hash={}", tx_hash,);
```
abizjak, line 258: I don't like that these fucntions both print and return stuff. We should probably just return values and print them in main. And this module is just responsible for doing things.

abizjak (changes requested):

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,453 @@
...
+pub struct ContractInitialized {
+    /// Smart contract address on the chain.
+    pub contract: ContractAddress,
+}
+
+/// A struct containing connection and wallet information.
+#[derive(Debug)]
+pub struct Deployer<'a> {
+    /// The client to establish a connection to a Concordium node (V2 API).
+    pub client: &'a mut v2::Client,
```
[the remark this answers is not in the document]
DOBEN: The idea was to get rid of the cloning through out the code.
https://github.com/Concordium/concordium-rust-smart-contracts/pull/324#discussion_r1356725318

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,459 @@
...
+pub struct ContractInitialized {
+    /// Smart contract address on the chain.
+    pub contract: ContractAddress,
+}
+
+/// A struct containing connection and wallet information.
+#[derive(Debug)]
+pub struct Deployer {
+    /// The client to establish a connection to a Concordium node (V2 API).
+    pub client: v2::Client,
```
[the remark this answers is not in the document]
abizjak: @DOBEN I don't think this is such a good idea.

At the cost of a few clones, which are really cheap for the client since it's an Arc, we add complexity if they do something odd.

If this was some tight inner loop it might be worthwhile, but it's not. So I recommend not doing this.

templates/default/deploy-scripts/src/deployer.rs
DOBEN: In my opinion, printing these statements should be the default behavior. We don't want every developer to have to add their own custom "logging/printing" to the deployment script in every script.

Logging/Printing is very helpful in case the script fails to execute successfully, then the script can be easily re-run from the point it failed (fixing the bug that caused it to fail beforehand of course).

I suggest to add a flag to "switch off" the verbose logging in case somebody wants to write their custom logging.

DOBEN: I suggest adding logging to the functions that execute a transaction on chain (meaning `deploy`/`initialize`/`update` transactions). All other functions (read functions e.g. estimate energy/ get nonce) should not log any output.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,453 @@
...
+                "Contract invoke failed: {reason:?}, used_energy={used_energy}, return \
+                 value={return_value:?}"
+            ))),
+            InvokeContractResult::Success {
+                return_value: _,
+                events: _,
+                used_energy,
+            } => {
+                let e = used_energy.energy;
+                println!("Contract invoke success: estimated_energy={e}");
```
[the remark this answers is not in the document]
DOBEN: Removed logging/printing in read functions.

templates/default/deploy-scripts/src/deployer.rs
abizjak: Ok, then keep it simple without switches to turn it off.

If it is meant to log then make sure to log consistently before every action that might fail.

DOBEN: I added already a commit that adds the flag to switch off logging yesterday.

abizjak: But why is it sensible to have this option? Doesn't it just complicate the code?

As I see it this is not a library, this is the deployment script for this particular contract.  And you argued that we should just log things since it's easier to debug and users don't have to recreate it every time. That makes sense to me.

But then adding this flag seems to me to go against this simplicity of use. If this was a library that we publish then I do agree we should have logging configurable, but then we should use the `log` crate for it since that is the standard way of doing logging in Rust, and allows the caller to configure it, enable it or disable it.

But since this is not a library, it is the template for the code they'll write it seems unnecessary complexity to add a flag like this.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,453 @@
...
+    }
+
+    /// A function that checks the outcome of the initialization transaction.
+    /// It throws if the `block_item` is not a initialization transaction.
+    /// It returns the error code if the transaction reverted.
+    fn check_outcome_of_initialization_transaction(
+        &self,
+        block_item: BlockItemSummary,
+    ) -> Result<ContractInitialized, DeployError> {
+        match block_item.details {
```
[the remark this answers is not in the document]
DOBEN: hmm, `is_rejected_account_transaction()` just returns a boolean if the transaction got rejected. The developer writing such scripts most likely wants to know the `reject_reason` to be able to debug their deployment script. 
To a smaller extent, if we use the `contract_init()` function, we also could just return some minimal information about that the given tx is not a `contract init transaction` in case the function returns `None`. 

The idea with these verbose errors is to give the developer as much as possible feedback on what could have been wrong in the transaction in case it failed.

templates/default/deploy-scripts/src/deployer.rs
DOBEN: Removed logging flag. Logging is done on the `deploy/initialize/update` transactions.

templates/default/deploy-scripts/src/deployer.rs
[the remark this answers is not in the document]
abizjak: https://docs.rs/concordium-rust-sdk/latest/concordium_rust_sdk/types/struct.BlockItemSummary.html#method.is_rejected_account_transaction

This function returns the reject reason which I think is what you want to print if I understood correctly.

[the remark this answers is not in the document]
DOBEN: ok, I see we can simplify the `check_outcome_of_init_transaction`. To do the same for the `check_outcome_of_update_transaction`/`check_outcome_of_deploy_transaction` function we would need something like `contract_update()`/`contract_deploy()`. So to keep the three parsing functions consistent, I would think it makes sense to keep the old logic. What do you think?

templates/default/deploy-scripts/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "1.0.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0"
+chrono = "0.4.26"
+tokio = {version = "1.18", features = ["rt", "macros", "rt-multi-thread"] }
+clap = { version = "4", features = ["derive", "env"]}
+concordium-rust-sdk="2.4"
```
abizjak, line 13: I just noticed this, the latest version of the Rust SDK is 3, which supports P6. We need to use that.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+            send::{deploy_module, init_contract, GivenEnergy},
+            InitContractPayload, UpdateContractPayload,
+        },
+        AccountTransactionEffects, BlockItemSummary, BlockItemSummaryDetails, ContractAddress,
+        Energy, TransactionType, WalletAccount,
+    },
+    v2::{self, BlockIdentifier},
+};
+use std::path::Path;
+use std::sync::Arc;
```
abizjak, line 22: Are you sure you formatted this? It looks like it was not formatted since there are multiple concordium_rust_sdk imports.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+    pub async fn deploy_wasm_module(
+        &mut self,
+        wasm_module: WasmModule,
+        expiry: Option<TransactionTime>,
+    ) -> Result<
+        (
+            Option<TransactionHash>,
+            Option<BlockItemSummary>,
+            ModuleReference,
+        ),
```
abizjak, line 82: It usually makes sense to add a struct when there is a triple.

Is there a reason not to do so here?

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+
+        let nonce = self.get_nonce(self.key.address).await?;
+
+        if !nonce.all_final {
+            anyhow::bail!("Nonce not final")
+        }
+
+        let expiry = expiry.unwrap_or(TransactionTime::from_seconds(
+            (chrono::Utc::now().timestamp() + 300) as u64,
+        ));
```
abizjak, line 107: Are you sure you ran clippy on this script? I would expect it to complain about this and suggest to use `unwrap_or_else`.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+            self.key.address,
+            nonce.nonce,
+            expiry,
+            wasm_module,
+        );
+        let bi = transactions::BlockItem::AccountTransaction(tx);
+
+        let tx_hash = self.client.clone().send_block_item(&bi).await?;
+
+        println!("Sent tx: {tx_hash}");
```
abizjak, line 120:
```suggestion
        println!("Sent transaction with hash: {tx_hash}");
```

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+        println!("Transaction finalized: tx_hash={}", tx_hash,);
+
+        Ok((tx_hash, block_item))
+    }
+
+    /// A function to estimate the energy needed to send a transaction on the
+    /// chain.
+    ///
+    /// If successful, the transaction energy is returned by this function.
+    /// This function can be used to dry-run a transaction.
```
abizjak, line 251: Note that it's important to clarify here that this will only estimate **execution energy**.

This is the energy that should be supplied to the `update` function, but it's important to clarify that the actual NRG cost on the chain will also have the administrative overhead based on the size and number of transactions.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+            InvokeContractResult::Success {
+                return_value: _,
+                events: _,
+                used_energy,
+            } => Ok(used_energy),
+        }
+    }
+
+    /// A function to get the next nonce of the wallet account.
+    pub async fn get_nonce(
```
abizjak, line 282: We should use the `sequence_number` terminology since that's the terminology in the node API nowadays.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+        block_item: &BlockItemSummary,
+    ) -> Result<ContractAddress, Error> {
+        match &block_item.details {
+            BlockItemSummaryDetails::AccountTransaction(a) => match &a.effects {
+                AccountTransactionEffects::None {
+                    transaction_type,
+                    reject_reason,
+                } => {
+                    if *transaction_type != Some(TransactionType::InitContract) {
+                        bail!("Expected transaction type to be InitContract");
```
abizjak, line 341:
```suggestion
                        bail!("Expected transaction type to be of type InitContract but it was instead {transaction_type:?}");
```

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+use itertools::Itertools;
+
+use concordium_rust_sdk::types::transactions::InitContractPayload;
+use deployer::Deployer;
+use std::{
+    io::Cursor,
+    path::{Path, PathBuf},
+};
+
+/// Reads the wasm module from a given file path and file name.
```
abizjak, line 26:
```suggestion
/// Reads the wasm module from a given file path.
```

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+    },
+    types::{
+        smart_contracts::{ModuleReference, WasmModule},
+        transactions,
+        transactions::send::GivenEnergy,
+    },
+    v2,
+};
+use itertools::Itertools;
+
```
abizjak, line 18:
```suggestion
```

And please format this script with our usual formatter.

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+async fn main() -> Result<(), Error> {
+    let app: App = App::parse();
+
+    let concordium_client = v2::Client::new(app.url).await?;
+
+    let mut deployer = Deployer::new(concordium_client, &app.key_file)?;
+
+    let mut modules_deployed: Vec<ModuleReference> = Vec::new();
+
+    for contract in app.module.iter().unique() {
```
abizjak, line 73: Is there a reason why we do `unique` here as opposed to doing what the user requested?

Seems to me like an odd place to filter this out, especially since it's filtering out by file name, rather than by module hash.

I think it would be better if the deploy_wasm_module function simply checked if what you're deploying already exists, and that's it. We don't need to complicate it here by filtering out duplicate file names.

templates/default/deploy-scripts/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+        let wasm_module = get_wasm_module(contract.as_path())?;
+
+        let (_, _, module) = deployer.deploy_wasm_module(wasm_module, None).await?;
```
abizjak, line 76:
```suggestion
        let (_, _, module) = deployer.deploy_wasm_module(wasm_module, None).await.context("Failed to deploy a module.")?;
```

abizjak (approved):
Looks good, but before merging can you please make sure that the generated code from the template has no clippy lint issues, and is formatted correctly using our usual formatter.

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+        }
+    }
+
+    /// A function to deploy a wasm module on the chain.
+    ///
+    /// If successful, the transaction hash, the block item, and
+    /// the module reference are returned.
+    /// If the module already exists on
+    /// chain, this function returns the module reference of the already
+    /// deployed module (no transaction hash or the block item returned).
```
limemloh, line 68: This sound like something you should capture in an enum

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+        let tx = deploy_module(
+            &*self.key,
+            self.key.address,
+            nonce.nonce,
+            expiry,
+            wasm_module,
+        );
+        let bi = transactions::BlockItem::AccountTransaction(tx);
+
+        let tx_hash = self.client.clone().send_block_item(&bi).await?;
```
limemloh, line 118: I could be wrong, but I don't think clone is needed here
```suggestion
        let tx_hash = self.client.send_block_item(&bi).await?;
```

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+            self.key.address,
+            nonce.nonce,
+            expiry,
+            payload,
+            energy,
+        );
+
+        let bi = transactions::BlockItem::AccountTransaction(tx);
+
+        let tx_hash = self.client.clone().send_block_item(&bi).await?;
```
limemloh, line 174:
```suggestion
        let tx_hash = self.client.send_block_item(&bi).await?;
```

templates/default/deploy-scripts/src/deployer.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// An optional energy for the transaction can be given. If `None` is
+    /// provided, 5000 energy is used as a default energy value. An optional
+    /// expiry time for the transaction can be given. If `None` is provided,
+    /// the local time + 300 seconds is used as a default expiry time.
+    pub async fn init_contract(
+        &mut self,
+        payload: InitContractPayload,
+        energy: Option<Energy>,
+        expiry: Option<TransactionTime>,
+    ) -> Result<(TransactionHash, BlockItemSummary, ContractAddress), Error> {
```
limemloh, line 148: Please consider introducing a struct for the return value.

limemloh (approved):

templates/default/deploy-scripts/src/deployer.rs
DOBEN: yes, the files are formatted.

The formatter's default setting does not group the imports automatically.
https://github.com/Concordium/concordium-rust-smart-contracts/pull/324#pullrequestreview-1684302368

We could change the behavior of the formatter by e.g. adding a `rustfmt.toml` file and adding the below line in it.
```
imports_granularity = "Crate"
```

For now, I just run the formatter with the `imports_granularity = "Crate"` once to make sure we group similar imports together.

templates/default/deploy-scripts/src/deployer.rs
DOBEN: There are no warnings by Clippy. Both options are fine and I chose the latter one because it is a bit less code.

```
expiry.unwrap_or_else( || TransactionTime::from_seconds(
            (chrono::Utc::now().timestamp() + 300) as u64,
        )
```
or 
```
expiry.unwrap_or( TransactionTime::from_seconds(
            (chrono::Utc::now().timestamp() + 300) as u64,
        )
```

templates/default/deploy-scripts/src/deployer.rs
DOBEN: The return value is called `AccountNonceResponse` and hence I would stick to the `nonce` terminology. In my opinion, an account nonce is clear what it is used for while a `sequence_number` is not so clear what it relates to.

templates/default/deploy-scripts/src/main.rs
DOBEN: removed. 

The idea was to prevent people from accidentally deploying the same module twice but we do some checks if a module already exists on chain as well, so it seems like not needed this check now.

templates/default/deploy-scripts/src/deployer.rs
abizjak: It's not that it's less code, the problem is that in the option `unwrap_or` you are making a call `Utc::now` to get the current time **always**, not just in the case it is not supplied. Both of them are correct in this case though.

So I'm surprised clippy does not complain, it usually does when there's a complex calculation in the unwrap_or clause.

PR #328, opened 2023-08-11
Remove credential storage contract

DOBEN:
## Purpose

Remove the credential storage contract. The web3ID infrastructure has changed from its initial proposal. Mainly credentials are not stored on-chain anymore and hence the storage contract is not needed anymore.

## Changes

- Remove the credential storage contract.


abizjak (approved):
Long live the storage contract ;-(

PR #336, opened 2023-09-15
Merge contract testing library

Bargsteen:
## Purpose

Merge the smart contract testing library into this repository.
This prepares us for migrating the old example tests to use the new testing library.

## Changes

- Merge in the library.
- Replace the contracts-common submodule with the rust-sdk (since it is needed by the testing lib and contains the contracts-common lib via the base submodule)
- Add a CI job for it that only runs when changes occur in the testing lib, the rust SDK or the rustfmt config file.
- Update the readme to the new reality.
- Checkout submodules recursively in the CI.
   - This is needed because the rust-sdk submodule has base as a submodule.  

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "concordium-contracts-common"]
-	path = concordium-contracts-common
-	url = ../concordium-contracts-common.git
+[submodule "concordium-rust-sdk"]
+	path = concordium-rust-sdk
+	url = git@github.com:Concordium/concordium-rust-sdk.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "concordium-rust-sdk"]
 	path = concordium-rust-sdk
-	url = git@github.com:Concordium/concordium-rust-sdk.git
+	url = ../concordium-rust-sdk.git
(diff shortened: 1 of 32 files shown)


.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "concordium-contracts-common"]
-	path = concordium-contracts-common
-	url = ../concordium-contracts-common.git
+[submodule "concordium-rust-sdk"]
+	path = concordium-rust-sdk
+	url = git@github.com:Concordium/concordium-rust-sdk.git
```
limemloh, line 3: This should probably be kept, as a relative path
```suggestion
	url = ../concordium-rust-sdk.git
```

limemloh (approved):
LGTM

.gitmodules
Bargsteen: Ok, that's fine. I am not sure of the benefits though. Which do you see?

abizjak (approved):

.gitmodules
abizjak: It works regardless of whether you check out the repository via https or ssh.

Bargsteen: Ah, neat.

PR #343, opened 2023-09-22
Fix bug in `StateMap::get_mut`

Bargsteen:
## Purpose

This PR fixes an issue in `StateMap::get_mut` which allowed users to get inconsistent views of the data. There is an example below.

When using `StateMap::get_mut` a value is loaded from the persistent storage into memory. The in-memory value can then be mutated and it will be written to the persistent storage once the value is dropped (via the `Drop` implementation on `StateRefMut`).

If you are able to look up multiple mutable values from the state at once, you can get inconsistent views of the data.
Here is an example:
image
`token_count` has the in-memory value `x + 2` and `token_count` has the value `x` as it is freshly loaded from persistence.

For this reason, we use lifetimes to ensure that you cannot have multiple mutable references to the state at the same time.
Or, at least that was what we meant to do!
We accidentally gave an immutable reference of `self` which exactly allows the example above to run instead of being caught by the borrow-checker.

## Changes

- Add three letters: `mut` (and a space)
- Add proper tests using `testbuild` to ensure that it won't occur again

## Checklist

      hard-to-understand areas.


concordium-std/tests/state/map-multiple-state-ref-mut.rs
```diff
@@ -0,0 +1,17 @@
...
+use concordium_std::*;
+
+pub fn main() {
+    let mut state_builder = StateBuilder::open(ExternStateApi::open());
+    let mut map: StateMap<u8, u8, _> = state_builder.new_map();
+    map.insert(0, 1);
+    map.insert(1, 2);
+    // Get two mutable references and unwrap the options.
+    let e1 = map.get_mut(&0u8).unwrap();
+    let e2 = map.get_mut(&1u8).unwrap();
```
abizjak, line 14: We don't really have to unwrap here.

abizjak (approved):
Good catch.

concordium-std/tests/state/map-multiple-state-ref-mut.rs
limemloh: Well, if both of these are `None`, we have a bug

limemloh: nevermind, I realize the point of the test is to fail

limemloh (approved):

concordium-std/tests/state/map-multiple-state-ref-mut.rs
Bargsteen:
```suggestion
    let e1 = map.get_mut(&0u8);
    let e2 = map.get_mut(&1u8);
```

PR #346, opened 2023-10-03
Update library version in docs

schwartz-concordium:
## Purpose

Update to latest version in module docs.

Both here and in the documentation on our web site they both reference version 1.0.

I haven’t gone through all the examples in documentation to see if any of them reference something from before 3.0. It did however work for me to use the latest (3.0) and follow the documentation.

### DIFF
--- a/contract-testing/src/lib.rs
+++ b/contract-testing/src/lib.rs
@@ -13,7 +13,7 @@
 //! edition = "2021"
 //!
 //! [dev-dependencies]
-//! concordium-smart-contract-testing = "1.0"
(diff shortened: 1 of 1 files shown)


Bargsteen (approved):
Thanks ;) I'll go through and update the example versions in my upcoming PR about migrating the tests.

abizjak:
> Thanks ;) I'll go through and update the example versions in my upcoming PR about migrating the tests.

In the interest of maintainability it might make sense to use the wildcard version `*` instead.

Issue #337, opened 2023-09-15
Rewrite tests of all example contracts to use the testing library


Bargsteen:
**Task description**

Rewrite tests of all example contracts to use the new(ish) testing library.
The examples with quick check tests won't be migrated yet, as the testing lib does not yet support quick check tests.

Depends on:
- https://github.com/Concordium/concordium-rust-smart-contracts/pull/336

Issue #338, opened 2023-09-15
Simplify the init and receive methods by reducing abstractions needed for the old testing infrastructure


Bargsteen:
**Task description**

The init and receive methods use several traits and generics that are only there to support the old testing infrastructure.
Once https://github.com/Concordium/concordium-rust-smart-contracts/issues/337 is completed, we can simplify the code by using the concrete types instead of the traits (e.g. `ExternHostApi` instead of `HasHost`), and we can also remove some of the generic parameters.

Bargsteen:
Closed by https://github.com/Concordium/concordium-rust-smart-contracts/pull/347

PR #347, opened 2023-10-06
Migrate tests for example contracts

Bargsteen:
## Purpose

Migrate the tests of most of our example contracts from the old testing infrastructure to the new integration testing library.
Closes https://github.com/Concordium/concordium-rust-smart-contracts/issues/337 and https://github.com/Concordium/concordium-rust-smart-contracts/issues/338.

## Changes

- [x] Migrate tests and remove use of abstractions on example contracts
  -  [x] account-signature-checks (none)
  - [x] auction
  - [x] cis2-multi
  - [x] cis2-multi-royalties
  - [x] cis2-nft
  - [x] cis2-wccd
  - [x] cis3-nft-sponsored-txs
  - [x] counter-notify
  - [x] credential-registry
     - [x] Migrate the tests that use entrypoints (will keep the quickcheck tests that use the state methods directly)
  - [x] eSealing
  - [x] fib
  - [x] icecream
  - [x] memo (none)
  - [x] nametoken
  - [ ] offchain-transfers (won't do, needs larger contract changes)
     - In particular, we need to add some more view entrypoints to be able to test it completely. In the interest of time, we'll skip that for now. 
  - [x] piggybank
  - [x] proxy
  - [x] recorder
  - [x] signature-verifier
  - [x] smart-contract-upgrade
  - [x] transfer-policy-check
  - [ ] two-step-transfer (won't do, uses quickcheck)
  - [x] voting
- [x] Derive some `PartialEq` / `Eq` for some cis2 types
- [x] Add type aliases the `Extern*` types, e.g. `type StateApi = ExternStateApi`
- [x] Include events from interrupts in `ContractInvokeSuccess::events`
- [x] Update rust-sdk dependency with a fix to policy serialization (bug found when migrating `transfer-policy-check`
- [x] Use `cargo concordium test --out `./concordium-out/module.wasm.v1` in the CI instead of `cargo test`
- [x] Update templates
   - [x] default
   - [x] cis2-nft
   - [x] credential-registry
- [x] Update documentation and deprecate old testing infrastructure: https://github.com/Concordium/concordium-rust-smart-contracts/pull/349

## Checklist

      hard-to-understand areas.


DOBEN (approved):
Very good to see how much this improves the contract's readability once we remove these generics (not needed anymore because the old unit testing will be deprecated soon). Nice work.

concordium-std/src/types.rs
```diff
@@ -1013,6 +1013,17 @@ pub type ReceiveResult<A> = Result<A, Reject>;
 /// ```
 pub type InitResult<S> = Result<S, Reject>;
 
+// TODO: Document these.
```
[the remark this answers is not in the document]
Bargsteen: See https://github.com/Concordium/concordium-rust-smart-contracts/pull/349

examples/cis2-multi-royalties/tests/tests.rs
```diff
@@ -0,0 +1,542 @@
...
+            from:     BOB_ADDR,
+            to:       CHARLIE_ADDR,
+        }),
+    ]);
+}
+
+// Helper function that sets up the contract with two types of tokens minted to
+// Alice. She has 400 of `TOKEN_0` and 1 of `TOKEN_1`.
+//
+// The royalty percentage is set to 10%.
```
Bargsteen, line 484:
```suggestion
/// Helper function that sets up the contract with two types of tokens minted to
/// Alice. She has 400 of `TOKEN_0` and 1 of `TOKEN_1`.
```

examples/cis2-wccd/tests/tests.rs
```diff
@@ -0,0 +1,670 @@
...
+        })
+        .expect_err("SetMetadataUrl");
+
+    // Check that the return value is correct.
+    let rv: ContractError =
+        from_bytes(&update.return_value().expect("Return value known to exist"))
+            .expect("Parsing ContractError");
+    assert_eq!(rv, ContractError::Unauthorized);
+
+    // Set the metadata URL as Alice, who is the admin.
```
Bargsteen, line 84:
```suggestion
    // Set the metadata URL from Alice's account, who is the admin.
```

examples/cis2-wccd/tests/tests.rs
```diff
@@ -0,0 +1,670 @@
...
+    assert_eq!(events, [WccdEvent::Cis2Event(Cis2Event::Transfer(TransferEvent {
+        token_id: TOKEN_ID_WCCD,
+        amount:   TokenAmountU64(1),
+        from:     ALICE_ADDR,
+        to:       BOB_ADDR,
+    })),]);
+}
+
+/// Test that you can add an operator.
+/// Initialize the contract with two tokens owned by Alice.
```
Bargsteen, line 144:
```suggestion
/// Initialize the contract with one token owned by Alice.
```

examples/cis3-nft-sponsored-txs/Cargo.toml
```diff
@@ -6,15 +6,17 @@ edition = "2021"
 license = "MPL-2.0"
 
 [features]
-default = ["std", "wee_alloc", "crypto-primitives"]
+default = ["std", "wee_alloc"]
```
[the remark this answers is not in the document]
Bargsteen: We still use the attribute `crypto-primitives` for some functions, yes, but we do not need this feature anymore.
With the `test_infrastructure` you use mocks for all the crypto operations by default, but you can use real operations via the `concordium-std/crypto-primitives` feature.
The new testing lib always uses the real crypto operations 😊

examples/cis3-nft-sponsored-txs/tests/tests.rs
```diff
@@ -0,0 +1,357 @@
...
+            operators:    vec![BOB_ADDR],
+        }),
+        (BOB_ADDR, ViewAddressState {
+            owned_tokens: vec![TOKEN_0],
+            operators:    Vec::new(),
+        }),
+    ]);
+}
+
+/// Helper function that sets up the contract with twp tokens minted to
```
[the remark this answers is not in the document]
Bargsteen: ✌️

examples/nametoken/tests/tests.rs
```diff
@@ -0,0 +1,712 @@
...
+            receive_name: OwnedReceiveName::new_unchecked("NameToken.view".to_string()),
+            address:      contract_address,
+            message:      OwnedParameter::empty(),
+        })
+        .expect("Invoke view");
+    from_bytes(&invoke.return_value).expect("Deserialize ViewState")
+}
+
+/// Deserialize the events from an update.
+fn deserialize_update_events(
```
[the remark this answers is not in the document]
Bargsteen: Yes, good point. I'll make a change request about it. 😊

limemloh (approved):

examples/cis3-nft-sponsored-txs/Cargo.toml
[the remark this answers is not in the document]
abizjak: That'll also help with the reproducible builds thing.

Having to support "crypto-primitives" is annoying since a lot of extra dependencies need to be installed there.

examples/counter-notify/tests/tests.rs
```diff
@@ -0,0 +1,115 @@
...
+            data: InstanceUpdatedEvent {
+                receive_name: rcv_name_3,
+                ..
+            },
+        },
+    ] if rcv_name_1 == "counter-notify.just-increment" && rcv_name_2 == "reentrancy-attacker.call-just-increment" && rcv_name_3 == "counter-notify.increment-and-notify"));
+
+    // Check that the contract observed the reentrancy attack.
+    let rv: bool = from_bytes(&update.return_value).unwrap();
+    assert_eq!(rv, false);
```
abizjak, line 114:
```suggestion
    assert!(!rv, "Re-entrancy attack not observed.");
```

examples/fib/tests/tests.rs
```diff
@@ -0,0 +1,80 @@
...
+            UpdateContractPayload {
+                amount:       Amount::zero(),
+                address:      contract_address,
+                receive_name: OwnedReceiveName::new_unchecked("fib.view".to_string()),
+                message:      OwnedParameter::empty(),
+            },
+        )
+        .expect("Calling receive");
+
+    let rv: u64 = from_bytes(&update.return_value).expect("Return value");
```
abizjak, line 78: This also keeps repeating a lot. I think we should add a small helper function to `ContractInvokeSuccess`.

```
pub fn parse_return_value<A: Deserial>(&self) -> ParseResult<A> {

}
```

That'll reduce the noise in tests.

examples/nametoken/tests/tests.rs
```diff
@@ -0,0 +1,712 @@
...
+const NAME_0: &str = "MyEvenCoolerName";
+const NAME_0_TOKEN_ID: ContractTokenId = TokenIdFixed([
+    24, 188, 23, 153, 239, 160, 1, 1, 235, 169, 225, 59, 20, 47, 103, 115, 3, 188, 0, 164, 87, 161,
+    21, 92, 67, 206, 126, 235, 0, 101, 54, 231,
+]);
+const NAME_1: &str = "MyCoolName";
+const NAME_1_TOKEN_ID: ContractTokenId = TokenIdFixed([
+    183, 115, 55, 205, 199, 72, 252, 206, 49, 209, 119, 201, 194, 140, 103, 91, 216, 188, 175, 84,
+    169, 226, 17, 22, 227, 62, 16, 117, 35, 207, 236, 137,
+]);
```
abizjak, line 23: Maybe we could implement `FromStr` for TokenIdFixed (we already have a hex-based Display).

That'll simplify tests.

examples/proxy/tests/tests.rs
```diff
@@ -0,0 +1,64 @@
...
+            message:      OwnedParameter::from_serial(&parameter).expect("Serialize parameter"),
+        })
+        .expect("Invoke proxy contract");
+
+    // Check that the return value can be deserialized and is correct.
+    // This ensures that the `RawReturnValue`s serial implementation is correct,
+    // in that it *doesn't* include the option tag and length values in the return
+    // value. If they were included, the string would have some extra bytes at
+    // the beginning.
+    let return_value: String = from_bytes(&update.return_value).expect("Deserialize return value");
```
abizjak, line 62: Also generally when we add those helper functions, we should make sure that they consume the entire input when parsing.

examples/signature-verifier/tests/tests.rs
```diff
@@ -0,0 +1,82 @@
...
+            amount:       Amount::zero(),
+            address:      init.contract_address,
+            receive_name: OwnedReceiveName::new_unchecked("signature-verifier.verify".to_string()),
+            message:      OwnedParameter::from_serial(&parameter_valid)
+                .expect("Parameter has valid size."),
+        })
+        .expect("Call signature verifier contract with a valid signature.");
+    // Check that it returns `true`.
+    let rv: bool = from_bytes(&update.return_value).expect("Deserializing bool");
+    assert_eq!(rv, true);
```
abizjak, line 81:
```suggestion
    assert!(true, "Signature checking failed.");
```

examples/signature-verifier/tests/tests.rs
```diff
@@ -0,0 +1,82 @@
...
+        public_key: PublicKeyEd25519([
+            53, 162, 168, 229, 46, 250, 217, 117, 219, 246, 88, 14, 119, 52, 228, 242, 73, 234,
+            165, 234, 138, 118, 62, 147, 74, 134, 113, 205, 126, 68, 100, 153,
+        ]),
+        signature:  SignatureEd25519([
+            170, 242, 191, 224, 247, 247, 70, 49, 133, 3, 112, 66, 33, 24, 243, 14, 135, 135, 197,
+            113, 122, 74, 21, 82, 122, 94, 29, 15, 252, 121, 27, 102, 59, 21, 9, 177, 33, 2, 46,
+            242, 96, 134, 179, 120, 89, 0, 29, 9, 100, 38, 116, 250, 59, 226, 1, 247, 217, 220, 39,
+            8, 245, 230, 236, 2,
+        ]),
```
abizjak, line 65: Could we add hex-based fromstr instnaces for these types? This is very ugly.

At least make an issue in base for it.

templates/cis2-nft/tests/tests.rs
```diff
@@ -0,0 +1,356 @@
...
+        .contract_invoke(ALICE, ALICE_ADDR, Energy::from(10000), UpdateContractPayload {
+            amount:       Amount::zero(),
+            receive_name: OwnedReceiveName::new_unchecked("{{crate_name}}.operatorOf".to_string()),
+            address:      contract_address,
+            message:      OwnedParameter::from_serial(&query_params).expect("OperatorOf params"),
+        })
+        .expect("Invoke view");
+
+    let rv: OperatorOfQueryResponse =
+        from_bytes(&invoke.return_value).expect("OperatorOf return value");
```
abizjak, line 209: And similarly to the ContractUpdateSuccess, we should add a `parse_return_value` helper to the invoke success as well.

abizjak (approved):
Looks good.

I'm wondering what the problem is with migrating credential-registry tests?

At least the ones in the credential_registry example are very simple. We can keep the existing tests that are for the state as they are, that was also the idea, and we migrate endpoint tests to use the testing library.

examples/auction/tests/tests.rs
```diff
@@ -0,0 +1,291 @@
...
+        .expect_err("Attempt to finalize auction after it has been finalized");
+    let rv: FinalizeError =
+        from_bytes(&update_10.return_value().expect("Return value known to exist"))
+            .expect("Return value is valid");
+    assert_eq!(rv, FinalizeError::AuctionAlreadyFinalized);
+}
+
+/// Setup auction and chain.
+///
+/// CAROL is the owner of the auction, which ends at `1000` milliseconds after
```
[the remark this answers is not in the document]
Bargsteen: No ;)

examples/nametoken/tests/tests.rs
Bargsteen: `FromStr::from_str` is not a `const` function, so that wouldn't help much in this case.
But it seems like I could implement a `const` method that would allow us to write `TokenIdFixed::convert_to_hex_and_parse(NAME_1)` by using https://docs.rs/const-hex/latest/const_hex/fn.const_encode.html and https://docs.rs/const-decoder/latest/const_decoder/.

But this might be too heavy a solution for now.
What do you think?

abizjak: Yeah, not for now.

examples/signature-verifier/tests/tests.rs
Bargsteen: Turns out we already have that 🎉

Bargsteen:
> Looks good.
> 
> I'm wondering what the problem is with migrating credential-registry tests?
> 
> At least the ones in the credential_registry example are very simple. We can keep the existing tests that are for the state as they are, that was also the idea, and we migrate endpoint tests to use the testing library.

I didn't notice that. Good point. I'll migrate those tests tomorrow.

Issue #340, opened 2023-09-15
Deprecate the old testing infrastructure


Bargsteen:
**Task description**

Make it clear that the old testing infrastructure is deprecated and write a migration guide in the developer documentation about how to migrate your code and tests.

Bargsteen:
Closed with https://github.com/Concordium/concordium-rust-smart-contracts/pull/347

Issue #339, opened 2023-09-15
Documentation following the simplification of init and receive methods


Bargsteen:
**Task description**

After https://github.com/Concordium/concordium-rust-smart-contracts/issues/338 is complete, the concrete types that are used directly must be updated to the same degree as the corresponding traits.

The code snippets in the doc comments must also be updated to use fewer abstractions.

Bargsteen:
Closed by https://github.com/Concordium/concordium-rust-smart-contracts/pull/347

PR #349, opened 2023-10-12
Deprecate `test_infrastructure` and supporting traits

Bargsteen:
## Purpose

Deprecate the `test_infrastructure` and supporting traits.
Closes https://github.com/Concordium/concordium-rust-smart-contracts/issues/339.
Closes https://github.com/Concordium/concordium-rust-smart-contracts/issues/340.

Depends on:
 - https://github.com/Concordium/concordium-base/pull/457
 - https://github.com/Concordium/concordium-rust-sdk/pull/130

## Changes

- Deprecate the `test_infrastructure` using the `deprecated` attribute.
- Deprecate a number of traits, including `HasHost` via the doc comments.
   - While using the `deprecated` attribute would be ideal, it would require a substantial amount of other changes to avoid deprecation warnings in user code. Since we are removing the `test_infrastructure` and traits completely in a version or two, the extra work was not worth the effort. 
- Document how to migrate your contracts to use the concrete types directly in the library docs.
- Added type aliases that remove the `Extern` prefixes for the concrete types.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std/src/lib.rs
+++ b/concordium-std/src/lib.rs
@@ -290,7 +290,7 @@
 //!    ) -> ReceiveResult<MyReturnValue> { todo!() }
 //!    ```
 //!
-//!    If you use logging, crypto-primitives, or similar, you must also also
+//!    If you use logging, crypto-primitives, or similar, you must also
 //! replace those uses of traits with concrete types. E.g. replacing `&mut impl
 //! HasLogger` with `&mut Logger`.
 //!
@@ -299,11 +299,10 @@
 //!    For an introduction to the library, see our [guide](https://developer.concordium.software/en/mainnet/smart-contracts/guides/integration-test-contract.html).
 //!
 //!    If you follow our [recommended structure](https://developer.concordium.software/en/mainnet/smart-contracts/best-practices/development.html#recommended-structure) in your contract,
-//!    then you can easily keep unit tests that only call methods directly on
-//!    your state struct.
-//!    Unit tests that call init or receive methods must, on the other hand, be
-//! migrated.
-//!
+//!    then you have a mix of unit and integrations tests:
+//!    - Unit tests that call methods directly on your state struct (without any
+//!      init/receive calls)
+//!    - Integration tests that call the init and receive methods
 //!
 //! [1]: https://doc.rust-lang.org/std/primitive.unit.html
 //! [test_infrastructure]: ./test_infrastructure/index.html
--- a/concordium-std/src/lib.rs
+++ b/concordium-std/src/lib.rs
@@ -373,7 +373,9 @@ pub mod constants;
 mod impls;
 pub mod prims;
 mod traits;
-mod types;
+// TODO: For reviewers: I made this pub to avoid a never used warning for
+// `ChainMetadata`. I am not sure why `pub use types::*;` isn't sufficient.
+pub mod types;
 pub use concordium_contracts_common::*;
 pub use impls::*;
 pub use traits::*;
@@ -384,4 +386,8 @@ pub use types::*;
 #[cfg_attr(feature = "wee_alloc", global_allocator)]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
+#[deprecated(
+    since = "9.0.0",
+    note = "Deprecated in favour of [concordium-smart-contract-testing](https://docs.rs/concordium-smart-contract-testing)."
+)]
 pub mod test_infrastructure;
--- a/concordium-std/src/lib.rs
+++ b/concordium-std/src/lib.rs
@@ -17,7 +17,7 @@
 //! Also note that `concordium-std` version 4 only works with `cargo-concordium`
 //! version 2.1+.
 //!
-//! Version 9 deprecates the module [test_infrastructure] in favour of the
+//! Version 8.1 deprecates the module [`test_infrastructure`] in favor of the
 //! library [concordium_smart_contract_testing], which should be used instead.
 //! For more details including how to migrate your contract, see the
 //! [Deprecating the
@@ -144,7 +144,7 @@
 //! contracts. The structure of these are, at present, a bit odd without the
 //! historic context, which is explained below.
 //!
-//! Prior to version 9, a number of traits and generics were used when writing
+//! Prior to version 8.1, a number of traits and generics were used when writing
 //! smart contracts, e.g. [`HasHost`], to support the usage of
 //! [`crate::test_infrastructure`] for testing, where two primary
 //! implementations of each trait existed. The first one is supported by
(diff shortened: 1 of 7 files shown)


concordium-std/src/lib.rs
```diff
@@ -207,9 +225,89 @@
...
+//!    If you use logging, crypto-primitives, or similar, you must also also
+//! replace those uses of traits with concrete types. E.g. replacing `&mut impl
+//! HasLogger` with `&mut Logger`.
+//!
+//! 2. Migrate your tests to use the new testing library.
+//!
+//!    For an introduction to the library, see our [guide](https://developer.concordium.software/en/mainnet/smart-contracts/guides/integration-test-contract.html).
+//!
+//!    If you follow our [recommended structure](https://developer.concordium.software/en/mainnet/smart-contracts/best-practices/development.html#recommended-structure) in your contract,
+//!    then you can easily keep unit tests that only call methods directly on
```
DOBEN, line 302: It is not completely clear what you mean here. Do you suggest that people can keep the unit tests when the methods directly call on the state struct but should add the `#[allow(deprecated)]` to the test case for now?

concordium-std/src/types.rs
```diff
@@ -1004,39 +991,106 @@ pub type ReceiveResult<A> = Result<A, Reject>;
...
+/// See [`ExternContext`] for more details.
 pub type InitContext = ExternContext<ExternInitContext>;
+
+/// Type alias for the context of receive methods.
+///
+/// See [`ExternContext`] for more details.
 pub type ReceiveContext = ExternContext<ExternReceiveContext>;
+
+/// The host, which supports interactions with the chain, such as querying
+/// balance of the contract, accessing its state, and invoking operations on
```
DOBEN, line 1014:
```suggestion
/// the balance of the contract, accessing its state, and invoking operations on
```

concordium-std/src/lib.rs
Bargsteen: Thanks for pointing this out, it now reads:

```
//!    If you follow our [recommended structure](https://developer.concordium.software/en/mainnet/smart-contracts/best-practices/development.html#recommended-structure) in your contract,
//!    then you have a mix of unit and integrations tests:
//!    - Unit tests that call methods directly on your state struct (without any
//!      init/receive calls)
//!    - Integration tests that call the init and receive methods
//!
//! If you do not want to migrate your contract and tests yet, then you can add
//! the `#[allow(deprecated)]` attribute to your test modules to avoid the
//! deprecation warnings.
```

concordium-std/src/lib.rs
```diff
@@ -17,6 +17,12 @@
 //! Also note that `concordium-std` version 4 only works with `cargo-concordium`
 //! version 2.1+.
 //!
+//! Version 9 deprecates the module [test_infrastructure] in favour of the
```
abizjak, line 20: I did not think we'd be bumping a major version for deprecation.

Are there breaking changes now?

concordium-std/src/lib.rs
```diff
@@ -130,30 +139,39 @@
...
-//! - [HasReceiveContext] for accessing all the context data available to the
-//!   receive functions (note that this includes all the common data)
-//! - [HasLogger] for logging data during smart contract execution
-//! - [HasPolicy] for accessing the policy of the sender, either of the init or
+//! # Essential types
+//! This crate has a number of essential types that are used when writing smart
+//! contracts. The structure of these are, at present, a bit odd without the
+//! historic context, which is explained below.
+//!
+//! Prior to version 9, a number of traits and generics were used when writing
```
abizjak, line 147: Same version here.

concordium-std/src/lib.rs
```diff
@@ -207,9 +225,92 @@
 //! | [QueryAccountBalanceError] | `-2147483623` |
 //! | [QueryContractBalanceError] | `-2147483622` |
 //!
-//! [1]: https://doc.rust-lang.org/std/primitive.unit.html
 //! Other error codes may be added in the future and custom error codes should
 //! not use the range `i32::MIN` to `i32::MIN + 100`.
+//!
+//! # Deprecating the `test_infrastructure`
+//! Version 9 deprecates the [test_infrastructure] in favour of the library
```
abizjak, line 232:
```suggestion
//! Version 9 deprecates the [`test_infrastructure`] in favor of the library
```

concordium-std/src/types.rs
```diff
@@ -1004,39 +997,108 @@ pub type ReceiveResult<A> = Result<A, Reject>;
...
+/// See [`ExternReturnValue`] for more details.
 pub type ReturnValue = ExternReturnValue;
 
+/// A type representing the return value of contract invocation.
+///
+/// A contract invocation **may** return a value. It is returned in the
+/// following cases
+/// - an entrypoint of a V1 contract was invoked and the invocation succeeded
+/// - an entrypoint of a V1 contract was invoked and the invocation failed due
+///   to a [CallContractError::LogicReject]
```
abizjak, line 1073:
```suggestion
///   to a [`CallContractError::LogicReject`]
```

concordium-std/src/types.rs
```diff
@@ -1004,39 +997,108 @@ pub type ReceiveResult<A> = Result<A, Reject>;
...
+///
+/// In all other cases there is no response.
+///
+/// This type is designed to be used via its [Read](crate::Read) and
+/// [HasCallResponse](crate::HasCallResponse) traits.
```
abizjak, line 1078:
```suggestion
/// [`HasCallResponse`](crate::HasCallResponse) traits.
```

concordium-std/src/lib.rs
```diff
@@ -286,4 +389,8 @@ pub use types::*;
 #[cfg_attr(feature = "wee_alloc", global_allocator)]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
+#[deprecated(
+    since = "9.0.0",
```
abizjak, line 393: I think this should be 8.1, shouldn't it?

concordium-std/src/lib.rs
```diff
@@ -275,7 +376,9 @@ pub mod constants;
 mod impls;
 pub mod prims;
 mod traits;
-mod types;
+// TODO: For reviewers: I made this pub to avoid a never used warning for
+// `ChainMetadata`. I am not sure why `pub use types::*;` isn't sufficient.
+pub mod types;
```
abizjak, line 381: There is a type called `ChainMetadata` in contracts_common, which is also re-exported from here.

Maybe we should have some other name for the type alias.

I don't think we want to export the `types` module.

concordium-std/src/lib.rs
Bargsteen: Right, I was unsure about which version this would be released as, so I just went with `9` or `9.0.0` everywhere so it would be easy to change.

Bargsteen: And no, there are no breaking changes.

concordium-std/src/lib.rs
Bargsteen: Ah, that's why. I've called it `ChainMeta` now.

PR #357, opened 2023-11-05
Add example how to generate Ed25519 signature in the test case

DOBEN:
## Purpose

It would be useful if we have test cases in our example smart contracts that showcase how to generate signatures in the test case directly side by side with our current approach of using external tools to create such signatures (e.g. https://cyphr.me/ed25519_tool/ed.html). This in-code signature generation would have simplified the umbrella smart contracts tests (checking ed25519 signatures) as well as simplified some of the cis3-sponsored-transaction smart contract tests (checking account signatures).

## Changes

- Add test case to `signature-verifier` to showcase how ed25519 signatures can be generated within a test case.

### DIFF
--- a/examples/signature-verifier/Cargo.toml
+++ b/examples/signature-verifier/Cargo.toml
@@ -12,7 +12,7 @@ concordium-std = {path = "../../concordium-std"}
 
 [dev-dependencies]
 concordium-smart-contract-testing = { path = "../../contract-testing" }
-ed25519-dalek = { version = "2.0.0", features = ["rand_core"]  }
+ed25519-dalek = { version = "2.0", features = ["rand_core"]  }
 rand = "0.8.5"
 
 [features]
--- a/examples/signature-verifier/Cargo.toml
+++ b/examples/signature-verifier/Cargo.toml
@@ -12,6 +12,8 @@ concordium-std = {path = "../../concordium-std"}
 
 [dev-dependencies]
 concordium-smart-contract-testing = { path = "../../contract-testing" }
+ed25519-dalek = { version = "2.0.0", features = ["rand_core"]  }
+rand = "0.8.5"
 
 [features]
 default = ["std", "wee_alloc"]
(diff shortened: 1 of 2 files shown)


examples/signature-verifier/Cargo.toml
```diff
@@ -12,6 +12,8 @@ concordium-std = {path = "../../concordium-std"}
 
 [dev-dependencies]
 concordium-smart-contract-testing = { path = "../../contract-testing" }
+ed25519-dalek = { version = "2.0.0", features = ["rand_core"]  }
```
abizjak, line 15: We usually only set major.minor version, not patch.

examples/signature-verifier/tests/tests.rs
```diff
@@ -73,3 +74,81 @@ fn test_signature_check() {
...
+            address:      init.contract_address,
+            receive_name: OwnedReceiveName::new_unchecked("signature-verifier.verify".to_string()),
+            message:      OwnedParameter::from_serial(&parameter_invalid)
+                .expect("Parameter has valid size."),
+        })
+        .expect("Call signature verifier contract with an invalid signature.");
+
+    // Check that it returns `false`.
+    let rv: bool = update_invalid.parse_return_value().expect("Deserializing bool");
+    assert_eq!(rv, false);
```
abizjak, line 131:
```suggestion
    assert!(!rv, "Signature verification should fail.");
```

examples/signature-verifier/tests/tests.rs
```diff
@@ -73,3 +74,81 @@ fn test_signature_check() {
...
+        .contract_update(SIGNER, ALICE, ALICE_ADDR, Energy::from(10_000), UpdateContractPayload {
+            amount:       Amount::zero(),
+            address:      init.contract_address,
+            receive_name: OwnedReceiveName::new_unchecked("signature-verifier.verify".to_string()),
+            message:      OwnedParameter::from_serial(&parameter_valid)
+                .expect("Parameter has valid size."),
+        })
+        .expect("Call signature verifier contract with a valid signature.");
+
+    // Check that it returns `true`.
```
abizjak, line 151:
```suggestion
    // Check that signature verifies.
```

abizjak (approved):

limemloh (approved):

Bargsteen (approved):

PR #359, opened 2023-11-07
Generating cis3 signatures in test case

DOBEN:
## Purpose

Some of the test cases from cis3 haven't been migrated yet. Migrating some of them. In addition, add test cases that showcase how to generate account signatures within a test case (meaning how to create keypairs and how to use the signing function).

## Changes

- Add test case `test_inside_signature_permit_transfer`
- Add test case `test_outside_signature_permit_transfer`
- Add test case `test_inside_signature_permit_update_operator`
- Add test case `test_outside_signature_permit_update_operator`

### DIFF
--- a/examples/cis3-nft-sponsored-txs/Cargo.toml
+++ b/examples/cis3-nft-sponsored-txs/Cargo.toml
@@ -16,7 +16,6 @@ concordium-cis2 = {path = "../../concordium-cis2", default-features = false}
 
 [dev-dependencies]
 concordium-smart-contract-testing = { path = "../../contract-testing" }
-ed25519-dalek = "1.0"
 rand = "0.7.0"
 
 [lib]
--- a/examples/cis3-nft-sponsored-txs/Cargo.toml
+++ b/examples/cis3-nft-sponsored-txs/Cargo.toml
@@ -16,6 +16,8 @@ concordium-cis2 = {path = "../../concordium-cis2", default-features = false}
 
 [dev-dependencies]
 concordium-smart-contract-testing = { path = "../../contract-testing" }
+ed25519-dalek = "1.0"
+rand = "0.7.0"
 
 [lib]
 crate-type=["cdylib", "rlib"]
--- a/examples/cis3-nft-sponsored-txs/src/lib.rs
+++ b/examples/cis3-nft-sponsored-txs/src/lib.rs
@@ -79,13 +79,15 @@ const SUPPORTS_PERMIT_ENTRYPOINTS: [EntrypointName; 2] =
 pub const NONCE_EVENT_TAG: u8 = u8::MAX - 5;
 
 /// Tagged events to be serialized for the event log.
-#[derive(Debug, Serial)]
+#[derive(Debug, Serial, Deserial, PartialEq, Eq)]
 #[concordium(repr(u8))]
 pub enum Event {
     /// The event tracks the nonce used by the signer of the `PermitMessage`
     /// whenever the `permit` function is invoked.
     #[concordium(tag = 250)]
     Nonce(NonceEvent),
+    #[concordium(forward = cis2_events)]
+    Cis2Event(Cis2Event<ContractTokenId, ContractTokenAmount>),
 }
 
 /// The NonceEvent is logged when the `permit` function is invoked. The event
@@ -744,7 +746,7 @@ fn contract_view_message_hash(
     // or sign a message (in that case the prepend is `account` address and 8 zero
     // bytes). Hence, the 8 zero bytes ensure that the user does not accidentally
     // sign a transaction. The account nonce is of type u64 (8 bytes).
-    let mut msg_prepend = vec![0; 32 + 8];
+    let mut msg_prepend = [0; 32 + 8];
     // Prepend the `account` address of the signer.
     msg_prepend[0..32].copy_from_slice(param.signer.as_ref());
     // Prepend 8 zero bytes.
@@ -962,10 +964,10 @@ fn contract_operator_of(
 
 /// Parameter type for the CIS-2 function `balanceOf` specialized to the subset
 /// of TokenIDs used by this contract.
-type ContractBalanceOfQueryParams = BalanceOfQueryParams<ContractTokenId>;
+pub type ContractBalanceOfQueryParams = BalanceOfQueryParams<ContractTokenId>;
 /// Response type for the CIS-2 function `balanceOf` specialized to the subset
 /// of TokenAmounts used by this contract.
-type ContractBalanceOfQueryResponse = BalanceOfQueryResponse<ContractTokenAmount>;
+pub type ContractBalanceOfQueryResponse = BalanceOfQueryResponse<ContractTokenAmount>;
 
 /// Get the balance of given token IDs and addresses.
 ///
--- a/examples/cis3-nft-sponsored-txs/tests/tests.rs
+++ b/examples/cis3-nft-sponsored-txs/tests/tests.rs
@@ -69,19 +69,18 @@ fn test_inside_signature_permit_update_operator() {
     };
     let payload = UpdateOperatorParams(vec![update_operator]);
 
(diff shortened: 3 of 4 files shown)


examples/cis3-nft-sponsored-txs/Cargo.toml
```diff
@@ -16,6 +16,8 @@ concordium-cis2 = {path = "../../concordium-cis2", default-features = false}
 
 [dev-dependencies]
 concordium-smart-contract-testing = { path = "../../contract-testing" }
+ed25519-dalek = "1.0"
+rand = "0.7.0"
```
abizjak, line 20: I guess rand is needed, but I'd prefer if we did not require users to manually add the dalek crate. They should be able to just use signature stuff from base. If this is not available we should add it.

examples/cis3-nft-sponsored-txs/src/lib.rs
```diff
@@ -79,13 +79,15 @@ const SUPPORTS_PERMIT_ENTRYPOINTS: [EntrypointName; 2] =
...
 /// Tagged events to be serialized for the event log.
-#[derive(Debug, Serial)]
+#[derive(Debug, Serial, Deserial, PartialEq, Eq)]
 #[concordium(repr(u8))]
 pub enum Event {
     /// The event tracks the nonce used by the signer of the `PermitMessage`
     /// whenever the `permit` function is invoked.
     #[concordium(tag = 250)]
     Nonce(NonceEvent),
+    #[concordium(forward = cis2_events)]
```
abizjak, line 89: So this changes the schema which seems like it should be a separate change since it changes the contract API that affects the dapps.

examples/cis3-nft-sponsored-txs/tests/tests.rs
```diff
@@ -304,19 +780,77 @@ fn initialize_contract_with_alice_tokens() -> (Chain, ContractAddress, ContractI
...
+                threshold: SignatureThreshold::ONE,
+            };
+
+            let mut key_map: BTreeMap<CredentialIndex, CredentialPublicKeys> = BTreeMap::new();
+            key_map.insert(
+                CredentialIndex {
+                    index: 0u8,
+                },
+                credential_public_keys,
+            );
```
abizjak, line 833: Like we have the `singleton` method on AccountKeys, we should have either a singleton method on AccountAccessStructure that constructs it from a public key, or add a From instance.

examples/cis3-nft-sponsored-txs/tests/tests.rs
```diff
@@ -19,11 +26,419 @@ const ACC_INITIAL_BALANCE: Amount = Amount::from_ccd(10000);
...
+    ]);
+
+    // Check balances in state.
+    let balance_of_alice_and_bob = get_balances(&chain, contract_address);
+
+    assert_eq!(balance_of_alice_and_bob.0, [TokenAmountU8(0), TokenAmountU8(1)]);
+}
+
+/// Test permit transfer function. The signature is generated outside this test
+/// case (e.g. with https://cyphr.me/ed25519_tool/ed.html). TOKEN_1 is transferred from Alice to Bob.
```
abizjak, line 349: I don't understand why the test is written this way. It adds complications. Why don't we write it in the way that we recommend, generating signatures inside the tests themselves.

examples/cis3-nft-sponsored-txs/tests/tests.rs
```diff
@@ -19,11 +26,419 @@ const ACC_INITIAL_BALANCE: Amount = Amount::from_ccd(10000);
...
+        signature: AccountSignatures {
+            sigs: signature_map,
+        },
+        signer:    ALICE,
+        message:   PermitMessage {
+            timestamp:        Timestamp::from_timestamp_millis(10000000000),
+            contract_address: ContractAddress {
+                index:    0,
+                subindex: 0,
+            },
```
abizjak, line 281:
```suggestion
            contract_address: ContractAddress::new(0,0),
```

examples/cis3-nft-sponsored-txs/tests/tests.rs
```diff
@@ -19,11 +26,419 @@ const ACC_INITIAL_BALANCE: Amount = Amount::from_ccd(10000);
...
+    };
+    let payload = TransferParams::from(vec![transfer]);
+
+    let mut inner_signature_map = BTreeMap::new();
+    inner_signature_map.insert(0u8, concordium_std::Signature::Ed25519(DUMMY_SIGNATURE));
+
+    let mut signature_map = BTreeMap::new();
+    signature_map.insert(0u8, CredentialSignatures {
+        sigs: inner_signature_map,
+    });
```
abizjak, line 269: This seems a bit odd. There is a "dummy" signature that is hardcoded, but the test header claims that signatures are generated as part of the test.

It should be explained why this is done, and what role the dummy signature plays.

abizjak (changes requested):

examples/cis3-nft-sponsored-txs/tests/tests.rs
```diff
@@ -19,11 +26,419 @@ const ACC_INITIAL_BALANCE: Amount = Amount::from_ccd(10000);
...
+
+    let mut permit_update_operator_param = PermitParam {
+        signature: AccountSignatures {
+            sigs: signature_map,
+        },
+        signer:    ALICE,
+        message:   PermitMessage {
+            timestamp:        Timestamp::from_timestamp_millis(10000000000),
```
Bargsteen, line 86:
```suggestion
            timestamp:        Timestamp::from_timestamp_millis(10_000_000_000),
```

Bargsteen: Consider adding a comment about why this timestamp is chosen.

Bargsteen (changes requested):
It's good that you added the tests, but as Ales also pointed out, it is not clear why the tests with generated keys also use the dummy constants. Please explain and/or fix that 😊

examples/cis3-nft-sponsored-txs/tests/tests.rs
DOBEN: I agree that the recommended way would be to pick either to generate all signatures in the test case (which should be the default way) or generate all signatures outside of the test cases.

This is an example contract providing both test case options, because sometimes you still need hardcoded signatures (e.g. if you have a setup such as the umbrella smart contracts where production account addresses are hardcoded in the smart contract, in that case, you don't want or don't can expose your private key in the smart contract testing file).

examples/cis3-nft-sponsored-txs/src/lib.rs
DOBEN: It appends the schema so I thought it would be fine to include (no breaking changes).

Testing transactions that emit the `Nonce` event as well as cis2 events are now easier done in the test cases.

examples/cis3-nft-sponsored-txs/tests/tests.rs
DOBEN: Added `singleton` method.

DOBEN:
```
It's good that you added the tests, but as Ales also pointed out, it is not clear why the tests with generated keys also use the dummy constants. Please explain and/or fix that 
```
I added the following explanation:
```
    // The `viewMessageHash` function uses the same input parameter `PermitParam` as
    // the `permit` function. The `PermitParam` type includes a `signature` and
    // a `signer`. Becuase these two values (`signature` and `signer`) are not
    // read in the `viewMessageHash` function, any value can be used and we choose
    // to use `DUMMY_SIGNATURE` and `ALICE` in the test case below.
    ```

Bargsteen (approved):

examples/cis3-nft-sponsored-txs/src/lib.rs
abizjak: Ok.

abizjak (approved):

PR #362, opened 2023-11-10
Bump rust version

DOBEN:
## Purpose

Bump rust version because integration tests are not running:
https://github.com/Concordium/concordium-rust-smart-contracts/actions/runs/6823199665/job/18556876930

## Changes

Bump rust version


concordium-std/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "concordium-std"
 version = "8.1.0"
 authors = ["Concordium <developers@concordium.com>"]
 edition = "2021"
-rust-version = "1.66"
+rust-version = "1.67"
```
abizjak, line 6: As far as I understand there is no problem with concordium-std as such. Only one of the (transitive) testing library dependencies no longer supports 1.66. So it's a bit odd to now set a minimum rust version here, preventing the use of the library with 1.66 for no reason.

The only reason for this seems to be that we use the same RUST_VERSION in the CI checks both for building concordium-std and the testing library.

I guess 1.67 is old enough, but it would be better to split the two versions in general since there are non-trivial diffferences, both in code generation quality and resulting code size between the different complier versions.

abizjak (approved):

Bargsteen (approved):
LGTM

concordium-std/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Set minimum Rust version to 1.67.
```
abizjak, line 5: I think this is no longer true for concordium-std. This should go to the testing library's changelog.

DOBEN: Removed the change log from `concordium-std`.

The testing library's changelog has already the following statement:
```
- Bump minimum supported Rust version to `1.67`.
```

Issue #379, opened 2024-01-09
Investigate whether to repair or replace Wee Alloc


Bargsteen:
**Task description**

We support a small allocator, [wee_alloc](https://github.com/rustwasm/wee_alloc/), via a feature `wee_alloc` for our smart contracts.
`wee_alloc` is not maintained and it has some outstanding bugs related to memory leaks (https://github.com/rustwasm/wee_alloc/issues/106).

We, therefore, want to investigate whether we should repair and replace wee_alloc.
The main goal is that we want an allocator that has a small code size and works well for most smart contracts. It can be a very simple allocator as smart contracts are always short-lived and usually quite simple (especially in terms of allocation patterns).

PR #392, opened 2024-01-30
Add a "bump allocator", a small and simple allocator for Wasm

Bargsteen:
## Purpose

This PR implements a new allocator for our smart contracts. It is simple and has a small code size. It uses more memory than a linked-list allocator such as wee_alloc, but for many contracts, this simplicity is sufficient and better.

Something I discovered while building this, is that the Rust compiler sets the initial number of pages in wasm modules to 17, which seems like quite a lot for smart contracts. Each page is 64KiB, so the initial memory is always ~1.1 MiB.

Closes https://github.com/Concordium/concordium-rust-smart-contracts/issues/379

## TODO
- The templates do not compile since they use the published version of concordium-std, so we need to publish it first, then merge this PR. Business as usual, but still a bit annoying and error-prone.

## Changes

- Implement a bump allocator
- Add a `bump_alloc` feature to `concordium_std`
- Add tests via the contract `bump-alloc-tests`
- Remove the `wee_alloc` feature **(breaking)**
- Change all contracts that use `wee_alloc` to use `bump_alloc`
- Update the `concordium-std` library documentation to reflect the new reality
- Bump MSRV to `1.72` in the testing lib to fix an issue with the indirect dependency `ahash` which requires MSRV `1.72`.

## Checklist

      hard-to-understand areas.


concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,184 @@
...
+/// assert_eq!((-a).div_ceil(-b), 3);
+/// ```
+const fn div_ceil(lhs: usize, rhs: usize) -> usize {
+    let d = lhs / rhs;
+    let r = lhs % rhs;
+    if r > 0 {
+        d + 1
+    } else {
+        d
+    }
```
limemloh, line 183: **Suggestion:** I did not test this, but I think we could avoid branching here, maybe the compiler already does this already internally.
```suggestion
    let rounding = r > 0 as usize;
    d + rounding
```

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,184 @@
...
+        let heap_end = &mut *self.heap_end.get();
+        let next = &mut *self.next.get();
+
+        // On the first allocation, the heap end is 0, as we cannot know how
+        // many pages the contract is configured to start with. So we get the
+        // actual size and update the value.
+        if *heap_end == 0 {
+            let actual_size = self.size().size_in_bytes();
+            *heap_end = actual_size;
+        }
```
limemloh, line 105: **Question:** Why is this done as part of `alloc` and not in `BumbAllocator::new`?

limemloh (approved):
Look great, very simple :+1:

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,184 @@
+//! This module contains a "bump allocator", which is a simple type of allocator
+//! designed for a small code footprint and to suit most smart contracts.
+//!
+//! A bump allocator has a `next` pointer which points to the next address in
+//! the heap to hand out. The `next` pointer is incremented by the size of the
```
DOBEN, line 5:
```suggestion
//! the heap to hand out. The `next` pointer is incremented by the size of
```

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,184 @@
...
+//! each allocation. This is sufficient for the allocator to work, but it also
+//! leaks memory. However, for short-lived programs such as smart contracts,
+//! memory leaks as perfectly fine and even preferable since the need for
```
DOBEN, line 8:
```suggestion
//! memory leaks are perfectly fine and even preferable since the need for
```

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,184 @@
...
+/// defining it as the global allocator.
+unsafe impl Sync for BumpAllocator {}
+
+pub struct BumpAllocator {
+    /// The next pointer to be handed out.
+    next:        UnsafeCell<usize>,
+    /// The end of the heap. Cannot be known at compile time and thus the
+    /// initial value is 0. It is updated on the first allocation.
+    heap_end:    UnsafeCell<usize>,
+    /// The number of active allocations. Used for resetting `next` to 1 if
```
DOBEN, line 52: Suggestion: In case that value changes

```suggestion
    /// The number of active allocations. Used for resetting `next` to `MIN_PTR_ADDR` if
```

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,184 @@
...
+        Self {
+            next:        UnsafeCell::new(MIN_PTR_ADDR),
+            heap_end:    UnsafeCell::new(0),
+            allocations: UnsafeCell::new(0),
+            // Must be set to the same initial address as `next`.
+            last_alloc:  UnsafeCell::new(MIN_PTR_ADDR),
+        }
+    }
+
+    /// Grow the memory by `delta` pages, each of size [PAGE_SIZE].
```
DOBEN, line 77: Suggestion: You could mentioned that the function returns the previous `PageCount`.

DOBEN (approved):
Tried it out and well done.

concordium-std/src/bump_alloc.rs
Bargsteen: I'll document that, thanks for pointing it out.
The reason is that the `new` must be a `const` method so we can declare a static with it.

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,184 @@
...
+use core::{
+    alloc::{GlobalAlloc, Layout},
+    cell::UnsafeCell,
+    ptr,
+};
+
+/// The byte size of Wasm pages.
+const PAGE_SIZE: usize = 65536;
+/// FIXME: Handing out memory address 0 doesn't work, but starting with 1 does.
+const MIN_PTR_ADDR: usize = 1;
```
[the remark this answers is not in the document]
Bargsteen: Ah, that makes sense. I didn't realize that the null pointer is the same as `0usize`. I thought it was some separate sentinel value or something like it.

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,184 @@
...
+/// let a = 8;
+/// let b = 3;
+/// assert_eq!(a.div_ceil(b), 3);
+/// assert_eq!(a.div_ceil(-b), -2);
+/// assert_eq!((-a).div_ceil(b), -2);
+/// assert_eq!((-a).div_ceil(-b), 3);
+/// ```
+const fn div_ceil(lhs: usize, rhs: usize) -> usize {
+    let d = lhs / rhs;
+    let r = lhs % rhs;
```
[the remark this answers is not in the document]
abizjak: Final comment, since this is only used when rhs = PAGE_SIZE, I think in that special case this would be
```
lhs >> 16 + (lhs & 0xffff != 0 as usize)
```

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,194 @@
...
+    ///
+    /// This must be a `const` method for it to be used as a global allocator,
+    /// where it is called staticly. Since we cannot know the actual size of
+    /// the heap, `heap_end` is initialized to `0` and updated on the first
+    /// allocation.
+    ///
+    /// # Safety
+    /// - Can only be used in single-threaded environments. The [`Sync`]
+    ///   implementation is only invalid and is only there to support setting
+    ///   this as the global allocator.
```
abizjak, line 76: Are we sure we are not breaking some invariants with implementing this?

I'm guessing not

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,184 @@
...
+/// Calculates the quotient of `lhs / rhs`, while rounding up to the nearest
+/// integer.
+///
+/// ```
+/// let a = 8;
+/// let b = 3;
+/// assert_eq!(a.div_ceil(b), 3);
+/// assert_eq!(a.div_ceil(-b), -2);
+/// assert_eq!((-a).div_ceil(b), -2);
+/// assert_eq!((-a).div_ceil(-b), 3);
```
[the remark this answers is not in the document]
Bargsteen: Fixed before you wrote this ;)

concordium-std/src/bump_alloc.rs
abizjak: No, let's do this.

concordium-std/src/bump_alloc.rs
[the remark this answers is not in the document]
Bargsteen: I tried out your proposed solution Ales, but it seemed off by one in a few cases. It also isn't the easiest thing to read, and the compiler may end up doing something similar on its own.
But as you pointed out, the `rhs` is constant, so I created a more bespoke function for calculating it.

Feel free to add the bitwise arithmetic if you find it essential.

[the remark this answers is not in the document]
abizjak: Hmm, did you use what I intended rather than what I wrote (the parenthesis)
```
fn f1(xs: usize) -> usize {
    (xs >> 16) + ((xs & 0xffff != 0) as usize)
}
```
This program
```
fn f1(xs: usize) -> usize {
    (xs >> 16) + ((xs & 0xffff != 0) as usize)
}

const N: usize = 1 << 16;

fn f2(xs: usize) -> usize {
    let a = xs / N;
    if xs % N > 0 {
        a + 1
    } else {
        a
    }
}

fn main() {
    for x in 0..usize::MAX {
        assert_eq!(f1(x), f2(x), "Failed for {x}");
    }
}
```
Does not panic.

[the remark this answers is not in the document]
Bargsteen: Ah, turns out I had `PAGE_SIZE` set to `65535` in the Rust playground instead of `65536`, which caused the discrepancy. I've used your suggested approach now ;)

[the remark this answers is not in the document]
abizjak: :)

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,241 @@
...
+    ///   this as the global allocator.
+    pub const unsafe fn new() -> Self {
+        Self {
+            // Initialized to the dummy value `0`.
+            next:        UnsafeCell::new(0),
+            // Initialized to the dummy value `0`.
+            heap_start:  UnsafeCell::new(0),
+            // Initialized to the dummy value `0`, which is checked during first initialization.
+            heap_end:    UnsafeCell::new(0),
+            allocations: UnsafeCell::new(0),
```
abizjak, line 105: Could you document what this field is for, and why it is needed?

concordium-std/src/bump_alloc.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// Get the size of memory in terms of pages.
+    ///
+    /// The memory has three sections in the following order:
+    ///
+    ///  - The data section.
+    ///  - The stack.
+    ///  - The heap.
+    ///
+    /// To get the start location of the heap, use `__heap_base`.
+    fn size(&self) -> PageCount {
```
abizjak, line 132: Would it be possible to rename this to `memory_size` or `linear_memory_size` since I got confused for a bit thinking this is the size of the heap.

concordium-std/src/lib.rs
```diff
@@ -121,24 +121,22 @@
...
-//! See Rust [allocator](https://doc.rust-lang.org/std/alloc/index.html#the-global_allocator-attribute) documentation for more context and details.
+//! to `bump_alloc`, which ships with `concordium-std`. This can be used both
+//! with and without the `std` feature.
+//!
+//! The main reason for using `bump_alloc` instead of the default allocator,
+//! even in `std` builds, is that `bump_alloc` has a smaller code footprint,
+//! i.e, the resulting smart contracts are going to be smaller by about 6-10kB,
+//! which means they are cheaper to deploy and run. `bump_alloc` is designed to
+//! be simple and fast, but it does not use the memory very efficiently. For
+//! short-lived programs, such as smart contracts, this is usually a perfectly
```
abizjak, line 134: Perhaps qualify here that for most smart contracts such as tokens this is the right tradeoff.

For very complex ones perhaps it's not the right choice.

abizjak (approved):
Beautiful.

examples/bump-alloc-tests/tests/tests.rs
```diff
@@ -0,0 +1,139 @@
...
+                receive_name: OwnedReceiveName::new_unchecked(
+                    "bump_alloc_tests.max_parameter_len".to_string(),
+                ),
+                message:      OwnedParameter::from_serial(&[42u8; 65535])
+                    .expect("Parameter size is below limit."),
+            },
+        )
+        .print_emitted_events()
+        .expect("Update should succeed");
+}
```
abizjak, line 66: Could you please add a test where you 
- have some fixed static string
- allocate some non-trivial amount, say 500kB of data
- make a function call with non-trivial parameters, and that function does something with the parameters
- recurse
- do something more in the function with the static string and parameters or some locals in the function defined before the recursive call.

I want us to have a test to make sure that we are not overwriting the stack or the static data in the allocator.

concordium-std/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Make the `concordium_dbg!` and related macros also usable with the full syntax
   that `println!` supports.
+- Remove the feature `wee_alloc` and replace it with `bump_alloc`, which enables a small and simple global allocator that can only be used in Wasm.
```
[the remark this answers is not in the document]
Bargsteen: Ah, classic merge error. Good catch ;)

concordium-std/src/bump_alloc.rs
Bargsteen: It is documented on the type. But I also added a line here.

concordium-std/src/bump_alloc.rs
Bargsteen: Yep. Renamed to `memory_size` to match the actual name in wasm. I also renamed `grow_memory` to `memory_grow` for the same reason.

examples/bump-alloc-tests/tests/tests.rs
Bargsteen: Yes, I've added that now. Let me know if you deem it sufficient ;)

abizjak:
@Bargsteen I suppose this is just waiting for us to release concordium-std/testing library before it can be merged?

Bargsteen:
> @Bargsteen I suppose this is just waiting for us to release concordium-std/testing library before it can be merged?

Yep :-)

PR #393, opened 2024-01-30
Fix macros like debug, fail so that they accept arguments as expected.

abizjak:
## Purpose

Fix macros defined in concordium-std so that they take arbitrary streams of tokens so that they behave as println!, assert!, etc.

## Checklist

      hard-to-understand areas.


Bargsteen (approved):
We should also mention it in the changelog. But looks good ;)

PR #394, opened 2024-01-31
Fix CI pipeline

DOBEN:
## Purpose

There are some tests that are currently failing but the CI pipeline does not catch them:
https://github.com/Concordium/concordium-rust-smart-contracts/actions/runs/7665712194/job/20892217229

The reason for the pipeline not failing is that there is a bug in `cargo-concordium` in versions below 3.2.0 related to zero status code returned even if an error occurred. The bug was fixed:
https://github.com/Concordium/concordium-smart-contract-tools/blob/main/cargo-concordium/CHANGELOG.md

The CI pipeline currently runs not the up-to-date version of `cargo-concordium`, which causes the problem.

## Changes

- Update `cargo-concordium` in CI pipeline to version 3.2.0.
- Update `rand` dependency to fix:
```
      let keypairs = AccountKeys::singleton(rng);
     |                            ---------------------- ^^^ the trait `rand_core::CryptoRng` is not implemented for `ThreadRng`
```
- Add `sponsored-tx-enabled-auction` contract to pipeline.


annenkov (changes requested):
Seems like `sponsored-tx-enabled-auction` and `cis2-multi` have an outdated `rand` dependency.

DOBEN:
The version is `0.8` fixes the problem, let me check if I can use `0.8.5` (most up-to-date version) as well.

annenkov:
> The version is 0.8 fixes the problem, let me check if I can use 0.8.5 (most up-to-date version) as well.

I thought that we leave the minor version unspecified, unless we are interested in some specific version for some reason. I can see locally that cargo installs the most recent version 0.8.5.

annenkov (approved):
Seems to be fixed everywhere now. Looks good.

abizjak (approved):

Issue #385, opened 2024-01-23
Macros for embedding constants in a convenient manner


abizjak:
**Task description**

It is often useful to embed constants into contracts. We do this often for keys or addresses in tests. Currently the only way to do this is to embed a representation such as
```
const ACC = [0u8; 32]
```
meaning a low-level representation. Similar for keys and signatures which is even more inconvenient since those are typically generated as hex strings by various tools.

A related issue is embedding compile time constants defined by environment variables. This can only be done if the constant is a `str` without resorting to writing custom macros. We should provide macros in concordium-std that allow for embedding of such constants.

PR #398, opened 2024-02-20
Procedural macros for creating constant values from strings

rasmus-kirk:
## Purpose

Addresses #385 

## Changes

Added macros for creating constant values from strings for the following types:
  - `AccountAddress`
  - `ContractAddress`
  - `ModuleReference`
  - `PublicKeyEcdsaSecp256k1`
  - `PublicKeyEd25519`
  - `SignatureEcdsaSecp256k1`
  - `SignatureEd25519`

## Checklist

      hard-to-understand areas.


concordium-std-derive/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+rust-version = "1.65"
+license = "MPL-2.0"
+description = "Procedural macros to generate concordium type constants from strings, that makes writing smart contracts on the Concordium blockchain easier."
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+proc-macro = true
+
+[dependencies]
+# Replace with new version, obviously...
```
rasmus-kirk, line 15: Todo

concordium-std-derive/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+};
+use proc_macro::TokenStream;
+use std::str::FromStr;
+use syn::{parse_macro_input, LitStr};
+
+/// Procedural macro for instantiating account addresses.
+/// Input must be a valid base58-encoding.
+#[proc_macro]
+pub fn account_address(item: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(item as LitStr);
```
rasmus-kirk, line 15: This also correctly handles the case where we pass in anything that is _not_ a string literal, fx:

```rust
const ACC: AccountAddress = account_address!(0);
```

Will correctly be caught, and produce the following error:

```
error: Invalid account address: Invalid Base58Check encoding.
 --> tests/test-programs/fail-acc-invalid-base58.rs:6:46
  |
6 | const ACC: AccountAddress = account_address!("30Bx2h5Y2veb4hZgAJWPrr8RyQESKm5TjzF3ti1QQ4VSYLwK1G");
  |                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

concordium-std-derive/Cargo.toml
abizjak: You don't have to use the `git` dependency. The contracts-common is part of concordium-base, which is a submodule of this repository. Please use taht dependency.

concordium-std-derive/src/lib.rs
abizjak: That does not look like the right error. We want to tell the user they need a string literal.

concordium-std-derive/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+        }
+    };
+
+    match format!("AccountAddress({:?})", addr_bytes).parse() {
+        Ok(o) => o,
+        Err(e) => {
+            let err = syn::Error::new(input.span(), format!("LexError: {}", e));
+            err.to_compile_error().into()
+        }
+    }
```
abizjak, line 32: So we can't use `quote!` to construct this? Something like
```
quote!(AccountAddress(#addr_bytes)
```

I don't really like parsing this thing again.

concordium-std-derive/tests/test-programs/fail-acc-multiple-arguments.stderr
```diff
@@ -0,0 +1,5 @@
+error: unexpected token
+ --> tests/test-programs/fail-acc-multiple-arguments.rs:6:98
+  |
+6 | const ACC: AccountAddress = account_address!("3kBx2h5Y2veb4hZgAJWPrr8RyQESKm5TjzF3ti1QQ4VSYLwK1G", "3kBx2h5Y2veb4hZgAJWPrr8RyQESKm5TjzF3t...
+  |                                                                                                  ^
```
abizjak, line 5: That's not the best error we can have. Could we have this case print also 'expected a string literal'?

concordium-std-derive/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/// Procedural macro for instantiating contract addresses.
+/// Input must be of the form "<index,subindex>", where index and subindex
+/// are integers.
```
abizjak, line 137:
```suggestion
/// are unsigned integers.
```

abizjak:
Could you also please go over the example contracts and use these macros in tests there.

rasmus-kirk:
Yes, of course

concordium-std/src/impls.rs
```diff
@@ -765,6 +765,7 @@ where
     StateApi: HasStateApi,
 {
     /// Ensures a value is in the entry by inserting the default value if empty.
+    #[allow(clippy::unwrap_or_default)]
```
rasmus-kirk, line 768: Clippy, unhelpfully, suggested to use `self.or_default()` instead...

abizjak (approved):

PR #400, opened 2024-02-21
Query contract module

### LINKED ISSUE #399: Expose Module Reference: concordium-std & testing library
Add support for the smart contract functionality for querying the module reference and contract name of a smart contract instance to

- [ ] the `concordium-std` library
- [ ] the smart contract testing library

td202:
## Purpose

Closes #399 
Depends on https://github.com/Concordium/concordium-rust-sdk/pull/183

Adds support for smart contracts to query the module reference and contract name of smart contract instances.

## Changes

- Add `contract_module_reference` and `contract_name` host functions to `concordium-std`.
- Add support for the new functionality in the testing library.
- Add tests for the implementation in the testing library matching the testing in the scheduler.
- Add a factory smart contract example that demonstrates and tests the functionality.

## Checklist

- [X] My code follows the style of this project.
- [X] I have performed a self-review of the changes.
- [X] I have documented my code, in particular the intent of the
      hard-to-understand areas.
- [X] (If necessary) I have updated the CHANGELOG.

### DIFF
--- a/contract-testing/CHANGELOG.md
+++ b/contract-testing/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased changes
 
+- Add support for querying the module reference and contract name of an instance.
+  This functionality is only available on-chain from protocol version 7.
+
 ## 4.1.0
 
 - Fix a bug in debug output. The events emitted before some contract queries
--- a/contract-testing/tests/contract_queries.rs
+++ b/contract-testing/tests/contract_queries.rs
@@ -265,7 +265,8 @@ fn test_module_ref_and_name() {
 }
 
 #[test]
-/// Test the interaction between upgrading a contract and querying the contract module reference.
+/// Test the interaction between upgrading a contract and querying the contract
+/// module reference.
 fn test_upgrade_module_ref() {
     let mut chain = Chain::new();
     let initial_balance = Amount::from_ccd(100);
--- a/contract-testing/tests/contract_queries.rs
+++ b/contract-testing/tests/contract_queries.rs
@@ -8,8 +8,8 @@ mod helpers;
 /// This test deploys three different smart contracts, from two different
 /// modules.  One of these contracts <0,0> has a function that queries and logs
 /// the module reference of a  specified contract address. Another <2,0> queries
-/// and logs the contract name of a specified  contract. Both functions are
-/// invoked for each instance, and for non-existant contract  addresses,
+/// and logs the contract name of a specified contract. Both functions are
+/// invoked for each instance, and for non-existant contract addresses,
 /// ensuring the values are as expected.
 ///
 /// The entrypoints are designed to succeed in the case of a match, fail with
--- a/contract-testing/tests/contract_queries.rs
+++ b/contract-testing/tests/contract_queries.rs
@@ -117,35 +117,35 @@ fn test_module_ref_and_name() {
         res.0, res_deploy_contract_inspection.module_reference,
         "Contract 0 should match contract inspection contract"
     );
-    assert_eq!(res.1, 776.into(), "Expected energy to get module reference '{}'", res.0);
+    assert_eq!(res.1, 775.into(), "Expected energy to get module reference '{}'", res.0);
     let res = get_module_ref(&mut chain, res_init_1.contract_address)
         .expect("Contract 1 should be valid");
     assert_eq!(
         res.0, res_deploy_account_balance.module_reference,
         "Contract 1 should match contract inspection contract"
     );
-    assert_eq!(res.1, 776.into(), "Expected energy to get module reference '{}'", res.0);
+    assert_eq!(res.1, 775.into(), "Expected energy to get module reference '{}'", res.0);
     let res = get_module_ref(&mut chain, res_init_2.contract_address)
         .expect("Contract 2 should be valid");
     assert_eq!(
         res.0, res_deploy_contract_inspection.module_reference,
         "Contract 2 should match contract inspection contract"
     );
-    assert_eq!(res.1, 776.into(), "Expected energy to get module reference '{}'", res.0);
+    assert_eq!(res.1, 775.into(), "Expected energy to get module reference '{}'", res.0);
     let err = get_module_ref(&mut chain, ContractAddress {
         index:    3,
         subindex: 0,
     })
     .expect_err("Contract <3,0> should be invalid");
     assert_eq!(err.reject_code(), Some(-1), "Rejection code for get_module_ref on <3,0>");
-    assert_eq!(err.energy_used, 744.into(), "Expected energy for failed get module ref.");
+    assert_eq!(err.energy_used, 743.into(), "Expected energy for failed get module ref.");
     let err = get_module_ref(&mut chain, ContractAddress {
         index:    1,
         subindex: 1,
     })
     .expect_err("Contract <0,1> should be invalid");
     assert_eq!(err.reject_code(), Some(-1), "Rejection code for get_module_ref on <0,1>");
-    assert_eq!(err.energy_used, 744.into(), "Expected energy for failed get module ref.");
+    assert_eq!(err.energy_used, 743.into(), "Expected energy for failed get module ref.");
 
     // Use contract 2 to check the contract name of the contracts.
     let get_name = |chain: &mut Chain,
@@ -183,7 +183,7 @@ fn test_module_ref_and_name() {
     );
     assert_eq!(
         res.1,
-        755.into(),
+        754.into(),
         "Expected energy for get_contract_name on {}",
         res.0.as_contract_name()
     );
@@ -196,7 +196,7 @@ fn test_module_ref_and_name() {
     );
     assert_eq!(
         res.1,
-        755.into(),
+        754.into(),
         "Expected energy for get_contract_name on {}",
         res.0.as_contract_name()
     );
@@ -209,7 +209,7 @@ fn test_module_ref_and_name() {
     );
     assert_eq!(
         res.1,
-        756.into(),
+        755.into(),
         "Expected energy for get_contract_name on {}",
         res.0.as_contract_name()
     );
@@ -219,14 +219,14 @@ fn test_module_ref_and_name() {
     })
     .expect_err("Contract <3,0> should be invalid");
     assert_eq!(err.reject_code(), Some(-1), "Rejection code for get_module_ref on <3,0>");
-    assert_eq!(err.energy_used, 742.into(), "Expected energy for failed get_contract_name");
+    assert_eq!(err.energy_used, 741.into(), "Expected energy for failed get_contract_name");
     let err = get_name(&mut chain, ContractAddress {
         index:    1,
         subindex: 1,
     })
     .expect_err("Contract <0,1> should be invalid");
     assert_eq!(err.reject_code(), Some(-1), "Rejection code for get_module_ref on <0,1>");
-    assert_eq!(err.energy_used, 742.into(), "Expected energy for failed get_contract_name");
+    assert_eq!(err.energy_used, 741.into(), "Expected energy for failed get_contract_name");
 }
 
 #[test]
--- a/contract-testing/tests/contract_queries.rs
+++ b/contract-testing/tests/contract_queries.rs
@@ -1,40 +1,25 @@
 //! This module contains tests for smart contract module reference and name
 //! queries. These tests match equivalent tests for the scheduler in the node.
-use std::io::Write;
 
 use concordium_smart_contract_testing::*;
 mod helpers;
 
-#[macro_export]
-/// Macro for a pattern that matches a [ContractInvokeError] where the contract
-/// rejects with the supplied error code.
-macro_rules! ContractInvokeErrorCode {
-    ($x:expr) => {
-        ContractInvokeError {
-            kind: ContractInvokeErrorKind::ExecutionError {
-                failure_kind: InvokeFailure::ContractReject {
-                    code: $x,
-                    ..
-                },
-            },
-            ..
-        }
-    };
-}
-
 #[test]
 /// This test deploys three different smart contracts, from two different
-/// modules. One of these contracts <0,0> has a function for checking if the
-/// module reference matches an expected value. Another <2,0> has a function for
-/// checking if the contract name matches an expected value. Both functions are
-/// invoked for each instance, and for non-existant contract addresses, ensuring
-/// the values are as expected.
+/// modules.  One of these contracts <0,0> has a function that queries and logs
+/// the module reference of a  specified contract address. Another <2,0> queries
+/// and logs the contract name of a specified  contract. Both functions are
+/// invoked for each instance, and for non-existant contract  addresses,
+/// ensuring the values are as expected.
 ///
 /// The entrypoints are designed to succeed in the case of a match, fail with
 /// code -1 if there is a mismatch, and fail with code -2 if the contract
 /// address does not exist. If the protocol version does not support the
 /// contract inspection functionality, then the call should fail with
 /// a runtime exception.
+///
+/// In addition to checking the results, this test also checks that the energy
+/// used matches the expected value.
 fn test_module_ref_and_name() {
     let mut chain = Chain::new();
     let initial_balance = Amount::from_ccd(100);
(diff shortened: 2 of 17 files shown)


contract-testing/tests/contract_queries.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    // Deploy the modules
+    let res_deploy_0 = chain
+        .module_deploy_v1(Signer::with_one_key(), helpers::ACC_0, mod_0)
+        .expect("Deploying valid module should work");
+    let res_deploy_1 = chain
+        .module_deploy_v1(Signer::with_one_key(), helpers::ACC_0, mod_1)
+        .expect("Deploying valid module should work");
+
+    // Initialise three contracts.
```
DOBEN, line 289: Initializing just one contract.

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+//! then the factory would need to know the module reference that products
+//! should use. This could be hard-coded in the contract, or set as a parameter
+//! when the factory is initialized.
+//!
+//! Since the construction and initialization of the product occur in two
+//! separate transactions, it is possible that a third party might try to hijack
+//! the process by inserting their own transaction to initialize the product
+//! (for instance, invoking a different factory instance than intended by the
+//! originator). To prevent this possibility, the product checks in its
+//! `initialize` method that the invoker of the `produce` transaction is the
```
DOBEN, line 47: We try to avoid using `ctx.invoker` (`tx_origin` equivalent on Ethereum) for authorization in smart contracts.

https://consensys.github.io/smart-contract-best-practices/development-recommendations/solidity-specific/tx-origin/

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+//!
+//! In this example the factory acts as a registry of all the products it has
+//! "produced", assigning each a fresh index and recording the contract address
+//! for each. The products are initialized with their index, and also keep a
+//! record of the address of the factory contract that produced them. In a
+//! practical application, the factory might supply additional paramaters when
+//! initializing the products.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
```
DOBEN, line 58:
```suggestion

```

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+        // Update the state
+        let state = host.state_mut();
+        state.next_product = next_product + 1;
+        state.products.insert(next_product, product_address);
+        Ok(())
+    }
+
+    /// Get the number of product instances that have been successfully
+    /// initialized by this factory.
+    #[receive(contract = "factory", name = "view_product_count")]
```
DOBEN, line 182: Can you add `return_value`
```suggestion
    #[receive(contract = "factory", name = "view_product_count", return_value = "u64")]
```

Not feeling strongly about it: With regards to the errors, we can keep it as it is to keep the contract simple or we can add.
```suggestion
    #[receive(contract = "factory", name = "view_product_count", error = "FactoryError")]
```

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+    pub fn view_product_count(
+        _ctx: &ReceiveContext,
+        host: &Host<FactoryState>,
+    ) -> Result<u64, FactoryError> {
+        Ok(host.state().next_product)
+    }
+
+    /// Get the contract address of the product instance for a particular index.
+    /// Indexes are assigned sequentially starting from 0.
+    #[receive(contract = "factory", name = "view_product", parameter = "u64")]
```
DOBEN, line 192:
```suggestion
    #[receive(contract = "factory", name = "view_product", parameter = "u64", return_value="ContractAddress")]
```

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+        ctx: &ReceiveContext,
+        host: &mut Host<ProductState>,
+    ) -> Result<(), ProductError> {
+        let state = host.state_mut();
+        // Check that the contract has not already been initialized.
+        if let ProductState::Initialized(_) = *state {
+            Err(ProductError::AlreadyInitialized)?;
+        }
+        // Check that the invoker is the owner of the contract.
+        if ctx.invoker() != ctx.owner() {
```
DOBEN, line 277: We try to avoid using `ctx.invoker` (`tx_origin` equivalent on Ethereum) for authorization in smart contracts.

https://consensys.github.io/smart-contract-best-practices/development-recommendations/solidity-specific/tx-origin/

`ctx.sender` should be used for authorization.

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+            Err(ProductError::NotAuthorized)?;
+        }
+        // The index is supplied as a parameter by the factory.
+        let index: u64 = ctx.parameter_cursor().get()?;
+        // This endpoint should only be called by another smart contract, namely the
+        // factory, which we record in the state.
+        let factory = match ctx.sender() {
+            Address::Contract(ca) => ca,
+            _ => Err(ProductError::NotInitializedByFactory)?,
```
DOBEN, line 286: If we call the error `NotInitializedByFactory` then I suggest checking if the error matches this one factory address.

Alternatively, I would call this error `NoContract`

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+        // Initialize the state.
+        let product = Product {
+            index,
+            factory,
+        };
+        *state = ProductState::Initialized(product);
+        Ok(())
+    }
+
+    #[receive(contract = "product", name = "view")]
```
DOBEN, line 297: Please add `return_value`.

examples/factory/tests/tests.rs
```diff
@@ -0,0 +1,231 @@
...
+/// Test that a product instance cannot be produced using an account that did
+/// not create the product instance.
+#[test]
+fn test_not_authorized() {
+    let (mut chain, module_ref) = initialize_chain();
+    let factory = create_factory(&mut chain, module_ref);
+    // ALICE creates the product instance, but BOB invokes produce on the factory.
+    // This should fail.
+    let product = create_product(&mut chain, module_ref, ALICE);
+    call_produce(&mut chain, factory, product, BOB).expect_err(
```
DOBEN, line 66: It would be good to check if the correct error is produced and not only some error/revert in the smart contract happened.

Similar in other test cases.

examples/factory/src/lib.rs
DOBEN: I suggest we don't have this check. If we want the `owner` of the product to initialize the `product` with some custom values then this should be done in the `constructor` logic of the `product` contract.

Is there a specific use case where we have to restrict the `product` to be added to the factory via the `produce` entrypoint by the deployer address of the `product`?

DOBEN (commented):
Well written.

DOBEN:
The contract name that we return from the query is appended by the `init_` phrase (e.g. "init_contract"). Would it be clear if we just return the `contract` part of the name for simplicity?

examples/factory/src/lib.rs
td202: The reason for this check is that we don't want a malicious third-party to add the product to a factory (which might, for instance, be a completely different factory than the initiator intended). Ideally, we would like the product to be constructed in the same transaction as the factory produces it. But since that is impossible, ensuring the transactions are from the same origin gets close enough, which is what this achieves.

The alternatives seem to be:

1. We assume that no third party will be sufficiently motivated to DoS attack the process. This could be fine, depending on the application.
2. The constructor logic of the product takes all of the custom initialization. This would include indicating which factory is authorised to produce the product. A third party would then be constrained to only produce the product exactly as the originator intended.
3. The product authorises using `ctx.sender`, but is then responsible for calling `produce` on the factory itself. i.e. instead of `user -> factory.produce -> product.initialize` the flow is `user -> product.call_produce -> factory.produce -> product.initialize`. I think this is a bit too complex.

I think the approach of checking the invoker is relatively clean and simple. It allows all of the initial parameters to be set in the call to `produce`, rather than splitting initialization across the constructor and `initialize`. I agree that in general it could be a mistake to use `ctx.invoker` for authorization, but I think this is a special case.

contract-testing/tests/contract_queries.rs
```diff
@@ -0,0 +1,363 @@
...
+            kind: ContractInvokeErrorKind::ExecutionError {
+                failure_kind: InvokeFailure::ContractReject {
+                    code: $x,
+                    ..
+                },
+            },
+            ..
+        }
+    };
+}
```
abizjak, line 23: It would be better to add a general helper method to `ContractInvokeError` since this is

```rust
/// If invocation failed with an execution error due to contract logic return the return status code and return value.
/// The return status will always be negative.
pub fn logic_reject(&self) -> Option<(i32, &[u8])>) {
   ..
}
```

These kind of helpers are useful.

And then just
```rust
matches!(res.logic_reject(), Some((-1, _)))
```

contract-testing/tests/contract_queries.rs
```diff
@@ -0,0 +1,363 @@
...
+                amount:    Amount::zero(),
+            },
+        )
+        .expect("Initializing valid contract should work");
+
+    let check_module_ref =
+        |chain: &mut Chain, address: ContractAddress, mod_ref: ModuleReference| {
+            let input_param = (address, mod_ref);
+            let op = OwnedParameter::from_serial(&input_param);
+            println!("{:?}", &op);
```
abizjak, line 308: Could we please not print.

examples/factory/src/lib.rs
DOBEN: I suggest leaving a comment above the `ctx.invoker() != ctx.owner()` pointing out that in some edge cases `ctx.invoker` shouldn't be used in authorization. An even include your alternative suggestion in a comment in the contract :).

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    /// Errors
+    #[derive(Serialize, Debug, PartialEq, Eq, Reject, SchemaType)]
+    pub enum FactoryError {
+        /// Failed parsing the parameter.
+        #[from(ParseError)]
+        ParseParams,
+        /// [`produce`] was called with a non-existent product contract.
+        /// Or [`view_product`] was called with an indext for a non-existent
```
abizjak, line 93:
```suggestion
        /// Or [`view_product`] was called with an index for a non-existent
```

contract-testing/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased changes
 
+- Add support for querying the module reference and contract name of an instance.
+  This functionality is only available on-chain from protocol version 7.
```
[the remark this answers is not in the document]
td202: I understood that the testing library shouldn't have a p7 feature guard.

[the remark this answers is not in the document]
abizjak: Never mind, I thought I was reading under the std project.

examples/factory/tests/tests.rs
```diff
@@ -0,0 +1,231 @@
...
+        .expect_err("Should fail when product has already been initialized.");
+}
+
+/// Test that `produce` fails when called with an instance of a contract other
+/// than the product.
+#[test]
+fn test_wrong_product() {
+    let (mut chain, module_ref) = initialize_chain();
+    let factory = create_factory(&mut chain, module_ref);
+    let factory2 = create_factory(&mut chain, module_ref);
```
annenkov, line 101: [nitpick] The test specifically tests this condition `product_name != PRODUCT_INIT_NAME`, which is a special case of "a contract other than the product." Since this is something relevant to the contract logic, I think it makes sense to reflect it in the documentation. It would be nice to have a test with another module reference, but it's a bit annoying, because we would need another contract for that.

annenkov (commented):
Looks good overall.
I left some minor remarks. I guess one thing we should decide on is whether using the `invoker` authorisation is satisfactory in this case.

examples/factory/src/lib.rs
td202: Of course this could happen, but I think that is not really worth guarding against. It requires the user to sign off on a scam transaction right between initialising the product and calling `produce`. Practically, the window between those two operations should be kept as small as possible. We also cannot completely prevent a user for falling for scams. But I think generally the result of a product being initialized incorrectly should generally be that the product is not really usable for its intended purpose, rather than it is useful for some malicious purpose. (The ways I could see it being used for some malicious purpose (as a result of this kind of hijacking) is if the ownership of the contract is somehow used for authorization, or if some privileges are granted to the contract before it is initialized.)

I think it's also worth noting that the flow `user -> contract -> factory.produce -> product.initialize` might also be useful in practice, so is not necessarily something we want to prohibit on the basis that it could be used maliciously.

examples/factory/src/lib.rs
td202: I think having the empty line makes sense because the attribute macro is applying to the file (`#!`) rather than the module below.

examples/factory/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+license = "MPL-2.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+default = ["std", "wee_alloc"]
+std = ["concordium-std/std"]
+wee_alloc = ["concordium-std/wee_alloc"]
+
+[dependencies]
+concordium-std = {path = "../../concordium-std", default-features = false, features = ["p7"]}
```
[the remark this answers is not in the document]
td202: OK.

examples/factory/src/lib.rs
td202: I have added an extended discussion in the top-level documentation of this file, which will hopefully address these issues.

examples/factory/tests/tests.rs
td202: I agree. Updated the doc, but I'm not going to the effort of testing it with a whole different module.

td202:
> The contract name that we return from the query is appended by the `init_` phrase (e.g. "init_contract"). Would it be better if we just return the `contract` part of the name for simplicity?

I went this way (partly) because `[Owned]ContractName` includes the `init_` prefix, so it seemed sensible to use that. I didn't want to introduce a new type for the prefix-less contract name, as I feel that would be likely to create confusion, and I didn't really want to mess around too much with the representation of these existing types. While the prefix is clearly redundant, I don't think removing it achieves simplicity, and absent a stronger argument, I propose to keep it.

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+        ensure_eq!(*state, ProductState::Uninitialized, ProductError::AlreadyInitialized);
+        // Check that the invoker is the owner of the contract.
+        ensure_eq!(ctx.invoker(), ctx.owner(), ProductError::NotAuthorized);
+        // The index is supplied as a parameter by the factory.
+        let index: u64 = ctx.parameter_cursor().get()?;
+        // This endpoint should only be called by another smart contract, namely the
+        // factory, which we record in the state.
+        let factory = match ctx.sender() {
+            Address::Contract(ca) => ca,
+            _ => Err(ProductError::NoFactoryContract)?,
```
DOBEN, line 400: Suggestion: call it `OnlyFactoryContract/OnlyContract` as we do this in our other examples.

The `NoFactoryContract` is confusing because it could be interpreted as that this entry point cannot be invoked by a factory contact `NoFactoryContract`.

https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/examples/auction/src/lib.rs#L133

examples/factory/tests/tests.rs
```diff
@@ -0,0 +1,256 @@
...
+    let product = create_product(&mut chain, module_ref, ALICE);
+    let err = call_produce(&mut chain, factory, product, BOB).expect_err(
+        "Should fail when product is called from a different account than the creator of the \
+         product.",
+    );
+    assert_eq!(
+        err.reject_code(),
+        Some(Reject::from(FactoryError::InitializeFailed).error_code.get()),
+        "Reject code should be InitializeFailed"
+    );
```
DOBEN, line 77: Just for info: This is how we check the error in our other examples. Not feeling strongly about either way. 

    // Check that the correct error is returned.
    let rv: ContractError = update.parse_return_value().expect("ContractError return value");
    assert_eq!(rv, ContractError::Unauthorized);
    
https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/examples/cis2-multi/tests/tests.rs#L235C1-L237C49

DOBEN (approved):

DOBEN:
It would be great if the factory logs an event when a new contract is created. This is particularly useful in (e.g. the `uniswapV2` protocol) since indexers pick up the new DEXMarket and provide analytics from that point on for the new trading pair.

annenkov (approved):

examples/factory/src/lib.rs
DOBEN: If the error should state what went wrong then the correct error description is:
`SenderIsAccountAddress`

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+//! address.
+//!
+//! The main disadvantage of this approach is that the isolation between the
+//! states of each product must be enforced by the contract itself. If the state
+//! became corrupted (due to a bug in the contract) then it could affect any or
+//! all of the encapsulated products. With the factory pattern, the runtime
+//! system of the chain enforces isolation. As always, the balance of risks
+//! should be considered when choosing the approach for any application.
+
+#![cfg_attr(not(feature = "std"), no_std)]
```
abizjak, line 154: Is it checked in CI that no_std actually compiles?

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// The state of the factory contract.
+    /// This keeps a record of all of the product contracts.
+    pub struct FactoryState<S: HasStateApi = StateApi> {
+        /// The index that will be assigned to the next product contract.
+        /// Equivalently, the number of product contracts that have been
+        /// produced.
+        ///
+        /// Note, for all practical purposes, using a `u64` to count the
+        /// products cannot overflow. In particular, the blockchain does
+        /// not (currently) allow for more than 2^64 smart contracts.
```
abizjak, line 177: I mean it will never allow for that. It's such an obscene number.

abizjak (approved):

examples/factory/src/lib.rs
```diff
@@ -0,0 +1,423 @@
+//! An example demonstrating how a factory pattern can be implemented with smart
+//! contracts on Concordium.
+//!
+//! > **Important Note:** The factory pattern is not generally idiomatic on
+//! > Concordium. In many cases, a more appropriate solution is to incorporate
+//! > the factory and products into a single smart contract instance, rather
+//! > than having a separate instance for each product.
```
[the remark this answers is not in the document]
td202: I've added a little more discussion. To me, maintaining an index of contracts is not an intrinsic part of the factory pattern - it is simply one smart contract that instantiates others. The fact that it can maintain a registry is incidental, and the main reason for doing it in this example is that there's no justification at all for using a factory if the factory doesn't have some additional functionality (on Concordium).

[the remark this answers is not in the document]
DOBEN: I agree that usually, the factory does not assign a specific index to the contract registered. This is something we added in this example but it is not part of the factory pattern/standard.

What is important so is that the factory functions as a registry in many use cases for people to look up/resolve addresses off-chain. For example, if you want to trade tokens in the `ETH-CCD` token pair on a DEX in UNISWAP V2 design. You would look up if the `ETH-CCD` token pair is already registered in the factory to find the contract that you should interact with if you want to trade these tokens in a pool.
https://github.com/Uniswap/v2-core/blob/master/contracts/UniswapV2Factory.sol#L23

examples/factory/src/lib.rs
td202: I believe it does now.

Issue #403, opened 2024-03-01
Implement ordered map for smart contracts


limemloh:
Introduce a cost-efficient ordered map as part of the smart contract library.

- [X] The ordered map data structure with the usual operations.
- [X] Tests, ensuring the map works as expected.
- [X] Documentation for each of the operations.

PR #405, opened 2024-03-01
Implement ordered maps for smart contracts

limemloh:
## Purpose

Closing #403

## Changes

- Add B-Tree based ordered maps where each node is stored separately in the low-level key-value store.

## Checklist

      hard-to-understand areas.


td202 (commented):
Looks good. I'd like to see clearer documentation of the invariants. Ideally, I'd like to see some randomized testing, especially testing structural invariants (e.g. the ordering of keys and size of nodes), and a relational invariant with respect to a model (e.g. `std::collections::BTreeSet`).

concordium-std/src/impls.rs
```diff
@@ -3114,6 +3179,998 @@ impl Deserial for MetadataUrl {
...
+            };
+
+            if node.is_leaf() {
+                return if key <= &node.keys[0] {
+                    lower_so_far
+                } else {
+                    // lower_key_index cannot be 0 in this case, since the binary search will only
+                    // return 0 in the true branch above.
+                    Some(StateRef::new(node.keys.swap_remove(lower_key_index - 1)))
+                };
```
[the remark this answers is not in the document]
limemloh: I'll fix this

concordium-std/src/impls.rs
```diff
@@ -3114,6 +3179,990 @@ impl Deserial for MetadataUrl {
...
+        // The node is a leaf, meaning we simply remove it.
+        // This will not violate the minimum keys invariant, since a node
+        // ensures a child can spare a key before iteration.
+        node.keys.remove(0)
+    }
+
+    /// Internal function for key rotation, preparing a node for deleting a key.
+    /// Assumes:
+    /// - The provided `node` is not a leaf and has a child at `index`.
+    /// -
```
DOBEN, line 3779: Is there something missing or can the line be removed?

concordium-std/src/impls.rs
```diff
@@ -3114,6 +3179,990 @@ impl Deserial for MetadataUrl {
...
+        let mut higher_so_far = None;
+        loop {
+            let higher_key_index = match node.keys.binary_search(key) {
+                Ok(index) => index + 1,
+                Err(index) => index,
+            };
+
+            if node.is_leaf() {
+                return if higher_key_index < node.keys.len() {
+                    Some(StateRef::new(node.keys.swap_remove(higher_key_index)))
```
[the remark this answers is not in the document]
limemloh: Yes, the swapping part is not important in itself, but `swap_remove` is faster than `remove`, which is why I use it here.

concordium-std/src/types.rs
```diff
@@ -1258,3 +1264,268 @@ pub struct MetadataUrl {
...
+        /// This list should never be empty and contain between `M - 1` and `2M
+        /// - 1` elements.
+        pub(crate) keys:     Vec<K>,
+        /// List of nodes which are children of this node in the tree.
+        ///
+        /// This list is empty when this node is representing a leaf.
+        /// When not empty it will contain exactly `keys.len() + 1` elements.
+        ///
+        /// The elements are ordered such that the node `children[i]` contains
+        /// keys that are strictly smaller than `keys[i]`.
```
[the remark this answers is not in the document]
limemloh: I have tried to rephrase to capture this

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+        // bunch of `MaybeUninit`s, which do not require initialization.
+        let mut prefixed: [mem::MaybeUninit<u8>; BTREE_NODE_KEY_SIZE] =
+            unsafe { mem::MaybeUninit::uninit().assume_init() };
+        for i in 0..STATE_ITEM_PREFIX_SIZE {
+            prefixed[i].write(prefix[i]);
+        }
+        let id_bytes = self.id.to_le_bytes();
+        for i in 0..id_bytes.len() {
+            prefixed[STATE_ITEM_PREFIX_SIZE + i].write(id_bytes[i]);
+        }
```
abizjak, line 1158:
```suggestion
        for (place, value) in prefixed[STATE_ITEM_PREFIX_SIZE..].iter_mut().zip(id_bytes) {
            place.write(value);
        }
```

abizjak: I'm also a bit confused as to why we do this here. Is it worse to allocate a [0u8; _] and just copy-from-slice?

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+impl<const M: usize, K, V, S> Deletable for StateBTreeMap<K, V, S, M>
+where
+    S: HasStateApi,
+    K: Serialize,
+    V: Serial + DeserialWithState<S> + Deletable,
+{
+    fn delete(mut self) { self.clear(); }
+}
+
+/// This test module rely on the runtime providing host functions and can only
```
abizjak, line 1249:
```suggestion
/// This test module relies on the runtime providing host functions and can only
```

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+/// be run using `cargo concordium test`.
+#[cfg(feature = "wasm-test")]
+mod wasm_test_btree {
+    use crate::{claim, claim_eq, concordium_test, StateApi, StateBuilder};
+
+    #[concordium_test]
+    fn test_btree_insert_6() {
```
abizjak, line 1256: Could you please have a sentence or two for each of these tests to explain what it's testing?

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+    /// node ID and create an entry in the smart contract key-value store.
+    fn create_node<'b>(
+        &mut self,
+        keys: Vec<K>,
+        children: Vec<NodeId>,
+    ) -> (NodeId, StateRefMut<'b, Node<M, K>, S>)
+    where
+        K: Serialize,
+        S: HasStateApi, {
+        let node_id = self.next_node_id.copy_then_increment();
```
abizjak, line 859: These node ids are u32, but I don't see them being reset unless you clear the set fully.

That worries me a bit even though it's probably still quite generous.

If we have 5 inserts per second this would run out in less than 30 years :)

I guess using u64 for node ids is more expensive.

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.id.serial(out) }
+}
+
+impl Deserial for NodeId {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        Ok(Self {
+            id: source.get()?,
+        })
+    }
+}
```
abizjak, line 1238: Why not just derive this?

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        self.keys.serial(out)?;
+        self.children.serial(out)
+    }
+}
+
+impl<const M: usize, K: Deserial> Deserial for Node<M, K> {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        let keys = source.get()?;
+        let children = source.get()?;
```
abizjak, line 1220: Won't this use u64 for length? Can't we use less?

abizjak: Also why not just derive it.

YOu can use derive macros as long as you import `self as concordium_std`.

concordium-std/src/types.rs
```diff
@@ -384,7 +384,12 @@ impl<'a, V> crate::ops::Deref for StateRef<'a, V> {
 /// that the value is properly stored in the contract state maintained by the
 /// node.
 pub struct StateRefMut<'a, V: Serial, S: HasStateApi> {
-    pub(crate) entry:            UnsafeCell<S::EntryType>,
+    /// This is set as an `UnsafeCell`, to be able to get a mutable reference to
+    /// the entry without `StateRefMut` being mutable.
+    /// The `Option` allows for having an internal method destroying the
+    /// `StateRefMut` into its raw parts without `Drop` causing a write to the
+    /// contract state.
```
abizjak, line 391: I don't really understand this, and I missed where this was actually needed. Can you point?

Also, why is std::mem::forget not a good alternative?

limemloh: It is needed in the `StateRefMut::into_raw_parts` method that I have introduced.
The point is to be able to take out the value and entry of the `StateRefMut` while preventing it from writing mutations to the key-value store.

`std::mem::forget` takes ownership of the value, so it won't be dropped, but we lose access to it as well in that case.

concordium-std/src/state_btree.rs
limemloh: It uses u32, since `Serial` in `concordium_contracts_common` defaults to `u32` for length.
However, `Serial` in `concordium-base` defaults to `u64`.

concordium-std/src/state_btree.rs
limemloh: I could use something like uleb128 if this is a concern, not sure how much it will impact the cost though

concordium-std/src/types.rs
abizjak: I see, you have problems moving out of it because we have a manual Drop implementation, but as far as I see this is actually just used once in delete_node.

Would it be very inconvenient to just directly call prims::state_delete_entry?

It's annoying to make the common cases worse for StateRefMut to work around a type system limitation.

limemloh: I'll try to avoid introducing the option, maybe replace `into_raw_parts` with something for deleting the entry directly and dropping the `StateRefMut` instead.

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+pub struct StateBTreeMap<K, V, S, const M: usize = 8> {
+    pub(crate) map:         StateMap<K, V, S>,
+    pub(crate) ordered_set: StateBTreeSet<K, S, M>,
+}
+
+impl<const M: usize, K, V, S> StateBTreeMap<K, V, S, M> {
+    /// Insert a key-value pair into the map.
+    /// Returns the previous value if the key was already in the map.
+    pub fn insert(&mut self, key: K, value: V) -> Option<V>
+    where
```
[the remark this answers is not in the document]
limemloh: Yes, that would make sense.
`StateMap::insert` seems to be missing that as well

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+    }
+
+    /// Get a mutable reference to the value corresponding to the key.
+    pub fn get_mut(&mut self, key: &K) -> Option<StateRefMut<V, S>>
+    where
+        K: Serialize,
+        S: HasStateApi,
+        V: Serial + DeserialWithState<S>, {
+        if self.ordered_set.is_empty() {
+            None
```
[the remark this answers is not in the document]
limemloh: We don't need it, but it is a small optimization.
Since the `ordered_set` is tracking the length, we can complete avoid the host call for looking up the key if the collection is empty.

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+            }
+        }
+    }
+
+    /// Returns the largest key in the set, which is strictly smaller than the
+    /// provided key. `None` meaning no such key is present in the set.
+    pub fn lower(&self, key: &K) -> Option<StateRef<K>>
+    where
+        S: HasStateApi,
+        K: Serialize + Ord, {
```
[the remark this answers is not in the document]
limemloh: For this function we use `Ord` doing the binary search of the keys, it is probably fine to do a linear search instead.
But I am not sure how we would implement `insert` with only `PartialOrd`.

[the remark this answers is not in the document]
abizjak: On a normal architecture it would probably make sense to do a linear search for small collections, but for our case that's not going to be helpful I think since costs are assigned based on number of operations.

concordium-std/src/state_btree.rs
limemloh: Not sure that I follow.
Do you mean avoid using `MaybeUninit` completely?

abizjak: Yes.

abizjak: I'm wondering if there will actually be a difference.

limemloh: A comparison with and without should settle that. I'll note this down when I get to benchmarking again

concordium-std/src/types.rs
limemloh: Calling `prims::state_delete_entry` would work, but it means that this data structure won't work with the old infrastructure. If we decide that is okay, then we might as well remove the type parameter `S` and hard-code `StateApi` for `StateBTree*`.

Another, more involved solution is to introduce a separate `StateRefMut`-like type just for this.

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+    pub fn higher(&self, key: &K) -> Option<StateRef<K>>
+    where
+        S: HasStateApi,
+        K: Serialize + Ord, {
+        self.ordered_set.higher(key)
+    }
+
+    /// Returns the largest key in the map, which is strictly smaller than the
+    /// provided key. `None` meaning no such key is present in the map.
+    pub fn lower(&self, key: &K) -> Option<StateRef<K>>
```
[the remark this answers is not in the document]
limemloh: Added

concordium-std/src/types.rs
limemloh: We discussed this offline, and the conclusion was that since we have deprecated the old unit testing infrastructure, we don't have to support it for this data structure.

concordium-std/src/types.rs
```diff
@@ -1258,3 +1264,268 @@ pub struct MetadataUrl {
...
+/// - `V`: Values stored in the map. Most operations on the map require this to
+///   implement [`Serial`](crate::Serial) and
+///   [`DeserialWithState<S>`](crate::DeserialWithState).
+/// - `S`: The low-level state implementation used, this allows for mocking the
+///   state API in unit tests, see
+///   [`TestStateApi`](crate::test_infrastructure::TestStateApi).
+/// - `M`: A `const usize` determining the _minimum degree_ of the B-tree.
+///   _Must_ be a value of `2` or above for the tree to be work. This can be
+///   used to tweak the height of the tree vs size of each node in the tree. The
+///   default is set based on benchmarks.
```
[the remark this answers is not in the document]
limemloh: Yes, I agree.
I'll make sure to describe the benchmark, I just need to make sure we are not changing anything first.

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,1481 @@
...
+/// - `V`: Values stored in the map. Most operations on the map require this to
+///   implement [`Serial`](crate::Serial) and
+///   [`DeserialWithState<S>`](crate::DeserialWithState).
+/// - `S`: The low-level state implementation used, this allows for mocking the
+///   state API in unit tests, see
+///   [`TestStateApi`](crate::test_infrastructure::TestStateApi).
+/// - `M`: A `const usize` determining the _minimum degree_ of the B-tree.
+///   _Must_ be a value of `2` or above for the tree to work. This can be used
+///   to tweak the height of the tree vs size of each node in the tree. The
+///   default is set based on benchmarks.
```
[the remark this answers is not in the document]
limemloh: Yes, I agree.
I'll make sure to describe the benchmark, I just need to make sure we are not changing anything first.

concordium-std/src/state_btree.rs
limemloh: This was discussed offline, and the conclusion from that was to first have benchmarks comparing `u32` vs `u64` for the NodeId and decide based on the results.

limemloh: The benchmarks showed a small cost increase. Small enough, so the decision was to use `u64` to remove the concern of running out of node IDs.

concordium-std/src/types.rs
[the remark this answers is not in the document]
limemloh: I have update the docs

concordium-std/src/state_btree.rs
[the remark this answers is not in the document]
limemloh: I have update the docs

concordium-std/src/state_btree.rs
limemloh: I compared these and it does improve it. Very little, but still an improvement

limemloh:
I have asked for a review again.
Currently, the only remaining task for this PR is to have the CI run the internal wasm unit-tests, but this is blocked until a new version of cargo-concordium is released.

abizjak:
> I have asked for a review again. Currently, the only remaining task for this PR is to have the CI run the internal wasm unit-tests, but this is blocked until a new version of cargo-concordium is released.

You can just build a binary now before it is released and download it. We don't want to be building cargo-concordium every time anyhow.

concordium-std/src/impls.rs
```diff
@@ -2373,48 +2402,37 @@ where
...
-    fn prepend_generic_map_key<K: Serial>(key: K) -> Vec<u8> {
-        let mut key_with_map_prefix = to_bytes(&GENERIC_MAP_PREFIX);
-        key_with_map_prefix.append(&mut to_bytes(&key));
-        key_with_map_prefix
+    /// Create a new empty [`StateBTreeSet`](crate::StateBTreeSet), setting the
+    /// minimum degree `M` of the B-Tree explicitly.
+    pub fn new_btree_set_degree<const M: usize, K>(&mut self) -> state_btree::StateBTreeSet<K, M> {
+        let (state_api, prefix) = self.new_state_container();
+        state_btree::StateBTreeSet::new(state_api, prefix)
+    }
```
abizjak, line 2425: I would prefer if we checked here that the degree is not invalid.

I think something like
```
    pub fn new_btree_set_degree<const M: usize, K>(&mut self) -> state_btree::StateBTreeSet<K, M> {
        if M >= 2 {
            let (state_api, prefix) = self.new_state_container();
            state_btree::StateBTreeSet::new(state_api, prefix)
        } else {
             // panic or maybe we can do report_error
        }
    }
```

in any smart contract code the M should be fully instantiated at some point so the if should be optimized away unless the person is using a wrong M.

concordium-std/src/impls.rs
```diff
@@ -2373,48 +2402,37 @@ where
...
+
+    /// Create a new empty [`StateBTreeMap`](crate::StateBTreeMap), setting the
+    /// minimum degree `M` of the B-Tree explicitly.
+    pub fn new_btree_map_degree<const M: usize, K, V>(
```
abizjak, line 2429: Same here.

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,2111 @@
...
+    fn test_btree_insert_present_key() {
+        let mut state_builder = StateBuilder::open(StateApi::open());
+        let mut tree = state_builder.new_btree_set_degree::<2, _>();
+        for n in [0, 3, 4, 1, 2].into_iter() {
+            tree.insert(n);
+        }
+        claim!(!tree.insert(1));
+    }
+
+    #[allow(deprecated)]
```
abizjak, line 1761: Why do we need this?

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,2111 @@
...
+
+                Self {
+                    expected_keys: crate::collections::BTreeSet::from_iter(
+                        inserted_keys.into_iter(),
+                    ),
+                    mutations,
+                }
+            }
+
+            fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
```
abizjak, line 2066: Could you please document the intent here, what is the shrinking strategy.

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,2111 @@
...
+
+    /// Get a reference to the value corresponding to the key.
+    pub fn get(&self, key: &K) -> Option<StateRef<V>>
+    where
+        K: Serialize,
+        V: Serial + DeserialWithState<StateApi>, {
+        // Minor optimization by first checking whether the collection is empty, since
+        // this information is kept at the root of the ordered set it is already loaded
+        // into memory. In the case of the empty collection we can then save a key
+        // lookup.
```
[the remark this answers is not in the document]
limemloh: I have rephrased this

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,2111 @@
...
+    where
+        K: Serialize + Ord, {
+        self.key_order.eq_or_higher(key)
+    }
+
+    /// Returns the largest key in the map, which is strictly smaller than the
+    /// provided key. `None` meaning no such key is present in the map.
+    pub fn lower(&self, key: &K) -> Option<StateRef<K>>
+    where
+        K: Serialize + Ord, {
```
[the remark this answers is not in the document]
limemloh: I did that at first, but I find this more readable in the documentation and code, since the constraints are next to each function rather than grouped somewhere above the function description.
Also, not every function depends on these traits.

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,2111 @@
...
+            );
+
+            stack.extend(node.children);
+        }
+        string
+    }
+
+    /// Check a number of invariants, producing an error if any of them are
+    /// violated. Should only be used while debugging and testing the btree
+    /// itself.
```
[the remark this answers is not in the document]
limemloh: I have rephrased this

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,2111 @@
...
+        let key = node_id.as_key(&self.prefix);
+        let entry = self.state_api.lookup_entry(&key).unwrap_abort();
+        StateRefMut::new(entry, self.state_api.clone())
+    }
+
+    /// Construct a string for displaying the btree and debug information.
+    /// Should only be used while debugging and testing the btree itself.
+    #[cfg(feature = "internal-wasm-test")]
+    pub(crate) fn debug(&self) -> String
+    where
```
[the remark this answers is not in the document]
limemloh: Since this is printing the internal structure of the btree, I wanted to avoid implementing `Debug`.
Don't think the user should ever need this specific format.

concordium-std/src/impls.rs
limemloh: Yes, that seems like a good idea

concordium-std/src/state_btree.rs
limemloh: We need this, because `#[concordium_quickcheck]` is located in a deprecated module

concordium-std/src/state_btree.rs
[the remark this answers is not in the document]
abizjak: Thank you

concordium-std/src/state_btree.rs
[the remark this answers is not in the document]
abizjak: It would only be implemented under this feature, at least that was my suggestion.

.github/workflows/linter.yml
```diff
@@ -12,9 +12,9 @@ name: Clippy & fmt
 
 env:
   RUST_FMT: nightly-2023-04-01
-  RUST_VERSION: "1.66"
-  RUST_VERSION_TESTING_LIBRARY: "1.72"
-  CARGO_CONCORDIUM_VERSION: "3.2.0"
+  RUST_VERSION: "1.73"
+  RUST_VERSION_TESTING_LIBRARY: "1.73"
+  CARGO_CONCORDIUM_VERSION: "3.3.0-internal"
```
abizjak, line 17:
```suggestion
  CARGO_CONCORDIUM_VERSION: "3.3.0"
```

.github/workflows/linter.yml
```diff
@@ -103,6 +103,35 @@ jobs:
...
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+
+      - name: Install Wasm target
+        run: rustup target install wasm32-unknown-unknown
+
+      - name: Run internal wasm unit test
+        run: |
+          CARGO_CCD=cargo-concordium_${{ env.CARGO_CONCORDIUM_VERSION }}
+          wget https://distribution.concordium.software/tools/linux/$CARGO_CCD
```
abizjak, line 129:
```suggestion
          curl https://github.com/Concordium/concordium-smart-contract-tools/releases/download/releases/cargo-concordium/${CARGO_CONCORDIUM_VERSION}/cargo-concordium-linux-amd64 -o cargo-concordium
          chmod +x cargo-concordium
          sudo mv cargo-concordium /usr/bin/cargo-concordium
```

concordium-std/src/impls.rs
```diff
@@ -2373,48 +2402,52 @@ where
...
-    /// Serializes the key and prepends [GENERIC_MAP_PREFIX].
-    /// This is similar to how [StateMap] works, where a unique prefix is
-    /// prepended onto keys. Since there is only one generic map, the prefix
-    /// is a constant.
-    fn prepend_generic_map_key<K: Serial>(key: K) -> Vec<u8> {
-        let mut key_with_map_prefix = to_bytes(&GENERIC_MAP_PREFIX);
-        key_with_map_prefix.append(&mut to_bytes(&key));
-        key_with_map_prefix
+    /// Create a new empty [`StateBTreeMap`](crate::StateBTreeMap), setting the
+    /// minimum degree `M` of the B-Tree explicitly.
```
abizjak, line 2436: Please add that M must be at least 2 here like in the `set_degree` above.

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,2132 @@
...
+    pub fn first_key(&self) -> Option<StateRef<K>>
+    where
+        K: Serialize + Ord, {
+        self.key_order.first()
+    }
+
+    /// Returns a reference to the last key in the map, if any. This key is
+    /// always the maximum of all keys in the map.
+    pub fn last_key(&self) -> Option<StateRef<K>>
+    where
```
abizjak, line 245: Since all of these just forward calls it'd probably be useful if they are inline(always). That probably helps to inline in case people compile with `-z` or `-s` options which is typical.

abizjak (approved):

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,2142 @@
...
+        let mut tree = state_builder.new_btree_set_degree::<2, _>();
+        let keys: Vec<u32> = (0..15).into_iter().collect();
+        for &k in &keys {
+            tree.insert(k);
+        }
+        let iter_keys: Vec<u32> = tree.iter().map(|k| k.clone()).collect();
+        claim_eq!(keys, iter_keys);
+    }
+
+    /// Testcase for dublicate keys in the set bug. Due to an edgecase where the
```
DOBEN, line 1763:
```suggestion
    /// Testcase for duplicate keys in the set bag. Due to an edge case where the
```

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,2142 @@
...
+        claim!(tree.contains(&1));
+        claim!(tree.remove(&1));
+        claim!(!tree.contains(&1));
+    }
+
+    /// Removing from a the root node which is at minimum keys, taking a child
+    /// from its lower child.
+    ///
+    /// - Builds a tree of the form: [[0, 1], 2, [3]]
+    /// - Remove 1
```
DOBEN, line 1736:
```suggestion
    /// - Remove 2
```

DOBEN (approved):

concordium-std/src/state_btree.rs
```diff
@@ -0,0 +1,2142 @@
...
+    use super::*;
+    use crate::{claim, claim_eq, concordium_test, StateApi, StateBuilder};
+
+    /// The invariants to check in a btree.
+    /// Should only be used while debugging and testing the btree itself.
+    #[derive(Debug)]
+    pub(crate) enum InvariantViolation {
+        /// The collection have length above 0, but no root.
+        NonZeroLenWithNoRoot,
+        /// The collection contain a root node, but this has no keys.
```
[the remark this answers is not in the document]
limemloh: Makes sense

td202 (approved):
Looks good!

PR #409, opened 2024-03-05
Add guide for choosing the right collection

### LINKED ISSUE #404: Add guide for choosing the right collection
For the [concordium-std documentation](https://docs.rs/concordium-std/latest/concordium_std/) we should introduce a guide to help the developer understand _when_ to use our different data-structures, possibly taking [inspiration from here](https://doc.rust-lang.org/std/collections/#when-should-you-use-which-collection).

limemloh:
## Purpose

Closes #404

### DIFF
--- a/concordium-std/src/lib.rs
+++ b/concordium-std/src/lib.rs
@@ -290,7 +290,9 @@
 //!
 //! - You want to track which keys you have seen.
 //! - Arbitrary values are associated with each of the keys.
-//! - The keys have some _ordering_ which is relevant.
+//! - The keys have some _ordering_ which is relevant e.g. if you need the key
+//!   which is located right above/below another key using
+//!   [`higher`](StateBTreeMap::higher)/[`lower`](StateBTreeMap::lower).
 //!
 //! ### Use `StateSet` when:
 //!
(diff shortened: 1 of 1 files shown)


concordium-std/src/lib.rs
```diff
@@ -242,7 +250,61 @@
...
+//! ### Use `StateMap` when:
+//!
+//! - You want to track which keys you have seen.
+//! - Arbitrary values are associated with each of the keys.
+//!
+//! ### Use `StateBTreeMap` when:
+//!
+//! - You want to track which keys you have seen.
+//! - Arbitrary values are associated with each of the keys.
+//! - The keys have some _ordering_ which is relevant.
```
abizjak, line 293: Maybe give an example here, "for example when you wish to query the smallest key".

concordium-std/src/lib.rs
```diff
@@ -242,7 +250,61 @@
...
+//! ### Use `StateSet` when:
+//!
+//! - You want to track which keys you have seen.
+//! - There is no meaningful value to associate with your keys.
+//!
+//! ### Use `StateBTreeSet` when:
+//!
+//! - You want to track which keys you have seen.
+//! - There is no meaningful value to associate with your keys.
+//! - The keys have some _ordering_ which is relevant.
```
abizjak, line 304: Same here.

abizjak (approved):

PR #411, opened 2024-03-12
Fix counter notify

rasmus-kirk:
## Purpose

It seems that the counter-notify example simply made the wrong check in
the test, note the following comment:

`This will be false if the reentrancy has occurred`

I added an enum to make the code clearer and fixed the assertion.

## Checklist

      hard-to-understand areas.


.github/workflows/linter.yml
```diff
@@ -26,34 +26,35 @@ jobs:
           - concordium-std-derive/Cargo.toml
           - concordium-std/Cargo.toml
           - concordium-cis2/Cargo.toml
-          - examples/voting/Cargo.toml
```
rasmus-kirk, line 29: Sorted alphabetically

abizjak (approved):

Issue #331, opened 2023-08-19
Templates are always taken from `main`


abizjak:
**Task description**

```
cargo concordium init
```

always takes the templates from `main`. This is a bit problematic since we often have unreleased changes on main and it is awkward to make those without breaking  something.

Instead of defaulting to main, we should specify a tag in `cargo generate` invocation.

We should support a special `template/latest` tag that we'd update each time we release new concordium-std, and this should be the default tag, and then we should also support normal release tags, e.g., `concordium-std/7.1.0`.

This would also make it simpler for us to develop.

abizjak:
@DOBEN @limemloh Any thoughts?

limemloh:
Seems like a reasonable approach, maybe even have template releases separate from `concordium-std` such that we can update these without updating `concordium-std`.

rasmus-kirk:
Solved with #419 and [this](https://github.com/Concordium/concordium-smart-contract-tools/pull/163)

PR #419, opened 2024-04-16
Added CI job for releasing templates

rasmus-kirk:
## Purpose

Since we added support in `cargo-concordium` to select template tags (see [this PR](https://github.com/Concordium/concordium-smart-contract-tools/pull/163)), it would be nice with a streamlined way to release new versions of these templates. This also fixes CI issues when updating templates, as they no longer run on lint, but only on release.

See also #331

## Changes

Added template release job and moved some template related checks from the linter to the new release job instead.

## Checklist

      hard-to-understand areas.


abizjak:
@limemloh I did not manage to fully complete this so I'm explaining what the plan was.

As you know it's currently rather annoying to update all the components since often templates rely on as of yet unpublished libraries.

In order to fix this two steps are needed
- cargo cocnordium is to be updated so that it does not always take things from main, but from a tag, which defaults to releases/templates/latest
- the CI in this repository should be updated so that templates are only checked when new template tags are created.

Now since we have cargo-concordium out in the wild we must first release a new version of cargo-concordium and wait some time, e.g., two weeks before we break templates on `main`.

Merge this PR.

The purpose of the change is that we will have normal releases for templates as well as for libraries. These template releases are created by a job defined in this PR. The purpose is to make sure that templates build and check out before making the tag.

This job is defined to create a release of the form releases/templates/v1.2.3 which should be protected and stable. Then we will have the special `releases/templates/latest` tag which will be updated to the "latest" version in separate step, which might as well be manual. This is similar to how docker images are versioned, with a special "latest" tag for lazy people.

In addition to the changes in this PR, the daily job https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/.github/workflows/check-generate-latest.yaml should be updated to check templates from `latest` with the latest cargo-generate.

abizjak (commented):
I think this generally looks good. But @limemloh should take over as described in the comment I added.

.github/workflows/release-templates.yaml
```diff
@@ -0,0 +1,339 @@
...
+      - clippy-credential-registry-template
+      - cargo-generate-cis2-nft-template-comparison
+      - cargo-generate-credential-registry-template-comparison
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Create a GitHub release
```
limemloh, line 333: I think we are missing a step here, which actually create the tag.
What I get from the documentation of [ncipollo/release](https://github.com/ncipollo/release-action) (at the bottom) it requires a tag already being there.

.github/workflows/release-templates.yaml
```diff
@@ -0,0 +1,339 @@
...
+        uses: ncipollo/release-action@v1
+        with:
+          tag: releases/templates/v${{ github.event.inputs.tag }}
+          draft: true
+          name: Template Release v${{ github.event.inputs.tag }}
+          body: ${{ steps.tag_version.outputs.changelog }}
```
limemloh, line 339: I don't think this works, `steps.tag_version` is referring to a step with `id: tag_version`, which is not here.

limemloh (changes requested):

.github/workflows/release-templates.yaml
rasmus-kirk: Yes, this is a mistake, the body was meant to be empty. Otherwise, we should parse the changelog and get changes from there, but that would be a little much for this.

.github/workflows/release-templates.yaml
rasmus-kirk: Tested this and the tag will be created once you "release" (undraft) the release.

limemloh (approved):

PR #423, opened 2024-04-23
[CIS-3 NFT] Fix nonce event serialization

lassemoldrup:
## Purpose

As per the [CIS-3 specification](https://proposals.concordium.software/CIS/cis-3.html), nonce events should be serialized with the `nonce` first, followed by the `account`. This is currently not the case.


DOBEN (approved):
can you apply these changes to:
- `cis2-multi` contract as well (implements the CIS3 standard).

Consider changing the order of the fields in the `NonceEvent` in the `cis5-smart-contract-wallet` (we haven't formally defined the standard yet, but keeping consistency in the `NonceEvent` seems meaningful across the contracts).

Bargsteen (approved):
If you also make the changes that Doris suggested, then it's good to go.
I assume that you tested it with CCD Scan as well, but did you also test the dApp?

Issue #422, opened 2024-04-22
Feedback/Improvements on the smart-contract-wallet implementation


DOBEN:
**Task description**

kudos to @soerenbf for his great input.

**Sub-tasks "Naming improvements to shorten/simplify the terminology"**
- [x] `nativeCurrency` rename to `CCD`.
- [x] `internalTransfer` rename to `transfer`.
- [x] view functions for hashes rename to `getCis2TransferMessageHash/getCCDTransferMessageHash/getCis2WithdrawMessageHash/getCCDWithdrawMessageHash`.
- [x] `balanceOfCis2Tokens` rename to `Cis2BalanceOf/CCDBalanceOf`.

**Sub-tasks "Adding serde feature to simplify parsing the type as JSON value between front end and back end"**
- [x] *Smart contract wallet*: add serde to all parameter types of the `internalTransfer/withdraw/deposit` functions.
- [x] *CIS2-library*: add serde to the `TokenAmount` and `TokenID` types, and all parameter types of `transfer/updateOperator` functions.
- [x] *wherever this type lives*: add serde to type `PublicKeyEd25519`.

limemloh:
Using `transfer` instead of `internalTransfer` makes the standard incompatible with CIS-2, so I would suggest using another name.

soerenbf:
> Using `transfer` instead of `internalTransfer` makes the CIS-5 standard incompatible with CIS-2, so I would suggest using some other name.

Ah, right.. didn't think of that! Good point 👍 

If the purpose is that these can be mixed, then maybe the transfer function should be prefixed with something that identifies this as a sc-wallet transfer? Can't figure out the best name just off the top of my head

DOBEN:
Suggestion: `cis5Transfer` for the name.

lassemoldrup:
It seems that the tests rely on `cis2-multi` being compiled into `../cis2-multi/concordium-out/module.wasm.v1`. Maybe this should be added as a step in a `build.rs` file when building tests? I don't know how well this plays with `cargo-concordium` though.

DOBEN:
> It seems that the tests rely on `cis2-multi` being compiled into `../cis2-multi/concordium-out/module.wasm.v1`. Maybe this should be added as a step in a `build.rs` file when building tests? I don't know how well this plays with `cargo-concordium` though.

Let's discuss and create a separate issue for it. There are 3 contracts that depend on other contracts currently:
- sponsored-tx-enabled-auction
[...]

PR #427, opened 2024-04-26
CIS-5 contract naming changes

lassemoldrup:
## Purpose

Addresses #422 by renaming according to the issue.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/cis5-smart-contract-wallet/src/lib.rs
+++ b/examples/cis5-smart-contract-wallet/src/lib.rs
@@ -13,7 +13,7 @@
 //! over the assets.
 //!
 //! Transfers of CCD and CIS-2 token balances (meaning `withdraw`
-//! and `internalTransfer` functions) do not require
+//! and `cis5Transfer` functions) do not require
 //! on-chain transaction submissions. Instead, the holder of the corresponding
 //! private key can generate a valid signature and identify a third party to
 //! submit the transaction on-chain, potentially incentivizing the third-party
@@ -26,8 +26,8 @@
 //! The three main actions in the smart contract wallet that can be taken are:
 //! - *deposit*: assigns the balance to a public key within the smart contract
 //!   wallet.
-//! - *internal transfer*: assigns the balance to a new public key within the
-//!   smart contract wallet.
+//! - *CIS-5 transfer*: assigns the balance to a new public key within the smart
+//!   contract wallet.
 //! - *withdraw*: withdraws the balance out of the smart contract wallet to a
 //!   native account or smart contract.
 //!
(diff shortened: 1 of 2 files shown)


examples/cis5-smart-contract-wallet/src/lib.rs
```diff
@@ -2,43 +2,44 @@
...
-//! flow on Concordium. Users can hold/transfer native currency/CIS-2 tokens
+//! flow on Concordium. Users can hold/transfer CCD/CIS-2 tokens
 //! without a native account as a starting point (no KYC required). Once users
 //! are ready to go through the KYC process to create a native account on
 //! Concordium, they can withdraw assets out of the smart contract wallet.
 //! The public key accounts in this smart contract wallet can't submit the
 //! transactions on chain themselves, but rely on someone with a native account
 //! (third-party) to do so.
+use concordium_cis2 as cis2;
 use concordium_cis2::*;
```
DOBEN, line 43:
```
use concordium_cis2::{self as cis2, *};
```

examples/cis5-smart-contract-wallet/src/lib.rs
```diff
@@ -75,27 +76,27 @@ pub enum Event {
...
+    WithdrawCcd(WithdrawCcdEvent),
     /// The event tracks every time a token amount held by a public key is
     /// withdrawn to an address.
     #[concordium(tag = 246)]
     WithdrawCis2Tokens(WithdrawCis2TokensEvent),
     /// The event tracks every time a CCD amount held by a public key is
     /// transferred to another public key within the contract.
     #[concordium(tag = 245)]
-    InternalNativeCurrencyTransfer(InternalNativeCurrencyTransferEvent),
+    CcdTransfer(CcdTransferEvent),
```
DOBEN, line 95: To be consistent with the `Deposit and Withdraw` events, I suggest to use `TransferCcd(TransferCcdEvent)`

examples/cis5-smart-contract-wallet/src/lib.rs
```diff
@@ -75,27 +76,27 @@ pub enum Event {
...
     /// The event tracks every time a token amount held by a public key is
     /// transferred to another public key within the contract.
     #[concordium(tag = 244)]
-    InternalCis2TokensTransfer(InternalCis2TokensTransferEvent),
+    Cis2TokensTransfer(Cis2TokensTransferEvent),
```
DOBEN, line 99: Same here:
```suggestion
    TransferCis2Tokens(TransferCis2TokensEvent),
```

DOBEN (approved):

examples/cis5-smart-contract-wallet/src/lib.rs
soerenbf: Side note: should this not be handled by the formatter?

soerenbf (approved):
Agree with @DOBEN, otherwise LGTM 😄

examples/cis5-smart-contract-wallet/src/lib.rs
DOBEN: Not tested, but it could be that the formatter config file has to be in every example folder to be picked up correctly.
https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/rustfmt.toml

or we are missing this specific rule in the config file.

lassemoldrup: I just checked, and it is definitely picking up the file. After comparing it with some of our other projects, the differences in `rustfmt.toml` are tiny and don't seem like they would influence imports. It seems like it is just not something that is done automatically.

PR #431, opened 2024-05-14
Add support for loading the contract module under test

### LINKED ISSUE #383: cargo concordium test should expose the location of the module
**Task description**

Currently when running `cargo concordium test` you need to either supply an `-o` parameter that matches what you put in the integration tests or don't supply any `-o` and have a path inside `target/concordium/...` in the integration tests when loading modules.

It would be more convenient that you don't have to hardcode these paths in the integration tests.

One way to achieve this would be that `cargo-concordium` sets an environment variable which is the location of the output artifact. The running test can access it.

lassemoldrup:
## Purpose

Closes #383. Depends on https://github.com/Concordium/concordium-smart-contract-tools/pull/167.

## Changes

Reads the environment variable set by `cargo-concordium` though the convenience function `module_load_output`. This also means that the user needs to update `cargo-concordium`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/contract-testing/src/constants.rs
+++ b/contract-testing/src/constants.rs
@@ -75,3 +75,9 @@ pub(crate) const INITIALIZE_CONTRACT_INSTANCE_CREATE_COST: Energy = Energy {
 pub(crate) const UPDATE_CONTRACT_INSTANCE_BASE_COST: Energy = Energy {
     energy: 300,
 };
+
+/// The name of the environment variable that holds the path to the contract
+/// module file. To load the module, use the
+/// [`module_load_contract`](crate::module_load_contract) function.
+pub const CONTRACT_MODULE_OUTPUT_PATH_ENV_VAR: &'static str =
+    "CARGO_CONCORDIUM_TEST_MODULE_OUTPUT_PATH";
--- a/contract-testing/src/constants.rs
+++ b/contract-testing/src/constants.rs
@@ -78,5 +78,5 @@ pub(crate) const UPDATE_CONTRACT_INSTANCE_BASE_COST: Energy = Energy {
 
 /// The name of the environment variable that holds the path to the contract
 /// module file. To load the module, use the
-/// [`module_load_contract`](crate::module_load_contract) function.
+/// [`module_load_output`](crate::module_load_output) function.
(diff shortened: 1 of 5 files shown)


contract-testing/src/impls.rs
```diff
@@ -2058,6 +2060,12 @@ pub fn module_load_v1(module_path: impl AsRef<Path>) -> Result<WasmModule, Modul
     Ok(module)
 }
 
+pub fn module_load_contract() -> Result<WasmModule, ContractModuleLoadError> {
```
limemloh, line 2063: **Suggestion:** Please add documentation for what the function does and when to use this.

limemloh, line 2063: **Suggestion:** I try not to care about naming things, but here I think the name should be more explicit of what it does compared to the `module_load_v1`.
Maybe `module_load_output`?

lassemoldrup: I like your suggestion better.

contract-testing/src/impls.rs
```diff
@@ -2058,6 +2060,13 @@ pub fn module_load_v1(module_path: impl AsRef<Path>) -> Result<WasmModule, Modul
     Ok(module)
 }
 
+/// Load the wasm module being tested output by `cargo concordium test`.
```
limemloh, line 2063:
```suggestion
/// Load the current smart contract module output using the environment variable `CARGO_CONCORDIUM_TEST_MODULE_OUTPUT_PATH` which is set when running using `cargo concordium test`.
```

contract-testing/src/constants.rs
```diff
@@ -75,3 +75,8 @@ pub(crate) const INITIALIZE_CONTRACT_INSTANCE_CREATE_COST: Energy = Energy {
 pub(crate) const UPDATE_CONTRACT_INSTANCE_BASE_COST: Energy = Energy {
     energy: 300,
 };
+
+/// The name of the environment variable that holds the path to the contract
+/// module file. To load the module, use the
+/// [`module_load_contract`](crate::module_load_contract) function.
```
limemloh, line 81:
```suggestion
/// [`module_load_output`](crate::module_load_output) function.
```

limemloh (approved):

PR #434, opened 2024-05-23
Change to use type with no length prepended in the CCD receive hook mechanism

DOBEN:
## Purpose

When a smart contract receives CCD via the receive hook mechanism from the smart contract wallet, we want to be able to represent any smart contract input parameter of the receiving smart contract. As such, it is problematic to use `AdditionalData`/ `Vec<u8>` types for the transfer since they all prepend the bytes with their length. The receiving smart contract input parameter might have no (or a different amount of) bytes prepended representing the length of the input parameter as such we send raw bytes now to be able to represent the input parameter of any receiving smart contract. 

## Changes

Send raw bytes in the CCD receive hook mechanism.

### DIFF
--- a/examples/cis5-smart-contract-wallet/src/lib.rs
+++ b/examples/cis5-smart-contract-wallet/src/lib.rs
@@ -534,6 +534,23 @@ pub struct TokenAmount {
     pub cis2_token_contract_address: ContractAddress,
 }
 
+/// Additional information to include with the CCD hook transfer. This is a type
+/// without the length of the bytes prepended when serializing. It can be used
+/// to represent any input parameter (no matter the length serialization of that
+/// input parameter) of the receiving contract.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(SerdeSerialize, SerdeDeserialize))]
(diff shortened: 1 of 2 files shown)


examples/cis5-smart-contract-wallet/src/lib.rs
```diff
@@ -792,10 +808,12 @@ fn withdraw_ccd(
                     Address::Account(account_address)
                 }
                 Receiver::Contract(contract_address, function) => {
+                    let receiving_contract_input_parameter = NoLengthPrependedBytes(data.as_ref());
+
                     // If the receiver is a contract: invoke the receive hook function.
                     host.invoke_contract(
                         &contract_address,
-                        &data,
+                        &receiving_contract_input_parameter,
```
limemloh, line 816: You can use [`host.invoke_contract_raw`](https://docs.rs/concordium-std/latest/concordium_std/trait.HasHost.html#tymethod.invoke_contract_raw) and pass `Parameter::new_unchecked(data.as_ref())` instead.
It would save allocating the bytes again and remove the need for `NoLengthPrependedBytes`

limemloh (approved):

PR #445, opened 2024-09-08
Allow cargo generate version 0.22

DOBEN:
# Purpose

Fixes https://github.com/Concordium/concordium-rust-smart-contracts/issues/442

# Changes

- Allow cargo-generate 0.22. The changes from 0.21.*/0.22.0 are not important https://github.com/cargo-generate/cargo-generate/releases/tag/v0.22.0.
https://github.com/cargo-generate/cargo-generate/releases.
- Fix some remaining tests/naming that were not updated after the rewrite of the `default` template https://github.com/Concordium/concordium-rust-smart-contracts/pull/441.

### DIFF
--- a/templates/cis2-nft/cargo-generate.toml
+++ b/templates/cis2-nft/cargo-generate.toml
@@ -1,5 +1,5 @@
 [template]
-cargo_generate_version = ">= 0.17.0, < 0.22.0"
+cargo_generate_version = ">= 0.17.0, < 0.23.0"
 
 [placeholders]
 description = { type="string", prompt="Description for the project?" }
--- a/templates/credential-registry/cargo-generate.toml
+++ b/templates/credential-registry/cargo-generate.toml
@@ -1,5 +1,5 @@
 [template]
-cargo_generate_version = ">= 0.17.0, < 0.22.0"
+cargo_generate_version = ">= 0.17.0, < 0.23.0"
 
 [hooks]
 pre = ["pre-script.rhai"] # a script for setting default values for the variables when `template_type = default`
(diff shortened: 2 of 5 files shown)


templates/cis2-nft/cargo-generate.toml
```diff
@@ -1,5 +1,5 @@
 [template]
-cargo_generate_version = ">= 0.17.0, < 0.22.0"
+cargo_generate_version = ">= 0.17.0, < 0.23.0"
```
Victor-N-Suadicani, line 2: Is there a reason why we don't just have the version set to "0.22.0" instead of this wide version?

Victor-N-Suadicani (approved):
Just one comment on the version. Also the concordium-std/src/lib.rs files seems to trigger some lints.

templates/cis2-nft/cargo-generate.toml
DOBEN: If people installed `cargo-generate` a while ago, their local version of `cargo-generate` is not up-to-date. However, they can still use our templates (meaning we don't want to force people to run a specific `cargo-generate` version locally just because our templates require them).

We introduced the templates when `cargo-generate` was at version `0.17.0` and tested from that version until the current version. Lower versions might also work, we just never tested them.

Victor-N-Suadicani: Makes sense

DOBEN:
We discussed that the lint warnings in `concordium-std/src/lib.rs` (which have been there for a long time) are not related to this PR and can be addressed in another PR.

PR #448, opened 2024-10-02
Update dependency versions in the cargo generate templates

Victor-N-Suadicani:
Not really sure if anything else is needed for #444.


DOBEN (commented):
Please update the versions in the github actions as well:

https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/.github/workflows/release-templates.yaml#L259

DOBEN (approved):

Issue #96, opened 2022-04-05
Add more introspection capabilities to TestHost


abizjak:
**Task description**

Add more tracing/introspection capabilities to TestHost. Currently you can observe which transfers occurred and in what order.

At least the following should be added

**Sub-tasks**
- [ ] Which contracts, in what order, and with what parameters were invoked, and what were the responses.

PR #450, opened 2024-10-03
Added functions in the primitive interface

rasmus-kirk:
## Purpose

See the related issue (#96)

Also see the accompanying PR ()  

Tested using this script, that compiles `cargo concordium` and runs the unit tests in `concordium-std`:

```bash
CONCORDIUM_STD_PATH="/path/to/concordium-std"
CARGO_CONCORDIUM_PATH="/path/to/cargo-concordium/Cargo.toml"

pushd "$CONCORDIUM_STD_PATH"
cargo run --manifest-path "$CARGO_CONCORDIUM_PATH" -- concordium test --only-unit-tests -- --features internal-wasm-test
popd
```

## Changes

- Added the following functions in the primitive interface and added unit tests:
  - `set_slot_time`
  - `set_receive_self_address`
  - `set_receive_self_balance`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-std/src/prims.rs
+++ b/concordium-std/src/prims.rs
@@ -288,9 +288,18 @@ extern "C" {
         column: u32,
     );
 
-    /// TODO
+    /// Set the slot time in milliseconds.
+    /// The slot time represents the beginning of the smart contract's block.
     #[cfg(all(feature = "wasm-test", target_arch = "wasm32"))]
     pub(crate) fn set_slot_time(slot_time: u64);
+
+    /// Sets the address of this smart contract
+    #[cfg(all(feature = "wasm-test", target_arch = "wasm32"))]
+    pub(crate) fn set_receive_self_balance(balance: u64);
+
+    /// Sets the current balance of this smart contract
+    #[cfg(all(feature = "wasm-test", target_arch = "wasm32"))]
+    pub(crate) fn set_receive_self_address(start: *const u8);
 
     #[cfg(feature = "debug")]
     /// Emit text together with the source location.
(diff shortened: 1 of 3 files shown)


concordium-std/src/lib.rs
```diff
@@ -463,6 +463,7 @@ pub mod constants;
 mod impls;
 pub mod prims;
 mod state_btree;
+pub mod test_env;
```
limemloh, line 466:
```suggestion
#[cfg(features = 'wasm-test')]
pub mod test_env;
```

concordium-std/src/prims.rs
```diff
@@ -288,6 +292,54 @@ extern "C" {
...
+    /// Gets event number `i` in the smart contract state. Returns `-1` if `i`
+    /// is an invalid index. Otherwise returns bytes written.
+    #[cfg(all(feature = "wasm-test", target_arch = "wasm32"))]
+    pub(crate) fn get_event(i: u32, start: *mut u8) -> i32;
+
+    /// Gets the size of event number `i` in the smart contract state. Returns
+    /// `-1` if `i` is an invalid index.
+    #[cfg(all(feature = "wasm-test", target_arch = "wasm32"))]
+    pub(crate) fn get_event_size(i: u32) -> i32;
+
```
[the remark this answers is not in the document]
rasmus-kirk: Don't think so

concordium-std/src/prims.rs
```diff
@@ -304,9 +341,20 @@ extern "C" {
         column: u32,
     );
 
-    #[cfg(all(feature = "wasm-test", feature = "concordium-quickcheck", target_arch = "wasm32"))]
+    /// Reporting back an error, only exists in debug mode
+    // TODO: Shouldn't this be feature gated with debug? Trivially adding it gives other errors.
```
limemloh, line 345: This is feature gated `#[cfg(feature = "wasm-test")]` at the top of the `extern 'C'`

limemloh (approved):

concordium-std/src/prims.rs
rasmus-kirk: Then why is the `debug_print` feature gated?

limemloh: right good thing this turned up, we don't want to change behavior for these, so `report_error` should now have `#[cfg(target_arch = "wasm32")]` to match the behavior from before.

`debug_print` should be feature gated separately from `wasm-test`, so it should be moved out into the above `extern 'C'` with `#[cfg(feature = "debug")]` like it was before. This function might be used within a debug build of a smart contract.

PR #457, opened 2025-03-14
Remove test infrastructure

rasmus-kirk:
## Purpose

See the related issue ([BASE-558](https://github.com/Concordium/concordium-base/issues/558))

## Checklist

      hard-to-understand areas.


examples/two-step-transfer/Cargo.toml
```diff
@@ -19,7 +19,7 @@ bump_alloc = ["concordium-std/bump_alloc"]
 version = "*"
 path = "../../concordium-std"
 default-features = false
-features = ["concordium-quickcheck"]
+features = ["concordium-quickcheck", "wasm-test"]
```
limemloh, line 22: the "wasm-test" feature is a bit special, since it changes how some functions gets compiled and should not be enabled in general, but only for unit test builds.
`cargo-concordium` should already enable this when building for unit testing.

allanbrondum:
I think I need a better description of what is in this PR. It seems some stuff was removed. What is the reference for that? The linked issue does not seem to match with the changes. In addition, the PR includes an update of the Rust version. Is that related or unrelated?

contract-testing/src/impls.rs
```diff
@@ -1694,20 +1694,15 @@ impl Chain {
...
-                .response
-            {
-                sdk::v2::ChainParameters::V0(p) => (p.euro_per_energy, p.micro_ccd_per_euro),
-                sdk::v2::ChainParameters::V1(p) => (p.euro_per_energy, p.micro_ccd_per_euro),
-                sdk::v2::ChainParameters::V2(p) => (p.euro_per_energy, p.micro_ccd_per_euro),
-                sdk::v2::ChainParameters::V3(p) => (p.euro_per_energy, p.micro_ccd_per_euro),
+                .response;
+            let exchange_rates = ExchangeRates {
+                euro_per_energy: chain_parameters.euro_per_energy.unwrap(),
+                micro_ccd_per_euro: chain_parameters.micro_ccd_per_euro.unwrap(),
```
limemloh: I think we need to extend ExternalNodeError with some variant representing node produced an unsupported response and fail for these instead of using `unwrap`

limemloh (approved):

concordium-std/src/test_infrastructure/trie.rs
```diff
@@ -1,369 +0,0 @@
-use super::{TestStateEntry, TestStateEntryData};
```
limemloh: Please update the changelog about this being removed

PR #479, opened 2026-08-12
Remove WASM quickcheck infrastructure

allanbrondum:
## Purpose

Remove the quickcheck infrastructure. Keep the `TestHost` `get_random` access from WASM tests.

## Changes

* Removed the quickcheck infrastructure
* Commented out the property based tests we have internally
* Exposed the randomness access via `TestEnv::get_random`
* Additionally entirely removed the `test_infrastructure` module. Moved out the stuff that is not deprecated (`report_error`)

## Checklist

      hard-to-understand areas.


concordium-std/src/types.rs
```diff
@@ -906,6 +904,31 @@ macro_rules! claim_ne {
...
+        )
+    };
+}
+
+/// Reports back an error to the host when compiled to wasm
+/// Used internally, not meant to be called directly by contract writers
+#[doc(hidden)]
+#[cfg(not(all(feature = "wasm-test", target_arch = "wasm32")))]
+pub fn report_error(_message: &str, _filename: &str, _line: u32, _column: u32) {}
+
```
allanbrondum, line 931: Moved here from `test_infrastructure`

concordium-std/src/traits.rs
```diff
@@ -22,15 +22,7 @@ use concordium_contracts_common::*;
...
-/// **This trait is deprecated along with
-/// [`crate::test_infrastructure`].**
-///
-/// Use [`ExternParameter`](crate::types::ExternParameter) instead unless you
-/// intend to use the deprecated test infrastructure.
-///
-/// See the [crate](../concordium_std/#deprecating-the-test_infrastructure)
-/// documentation for more details.
+/// Consider using the concrete type [`ExternParameter`](crate::types::ExternParameter) instead.
 pub trait HasParameter: Read + Seek + HasSize {}
```
limemloh, line 26: Are these traits still useful? maybe these should be marked deprecated as well.

limemloh (approved):

concordium-std/src/traits.rs
allanbrondum: I think we can remove them at some point, since they all (or almost all?) only have one implementation.

PR #480, opened 2026-08-28
Cor 2478 update templates

allanbrondum:
## Purpose

Update smart contract templates

## Changes

* updated templates to use new concordium-std
* updated templates to not use the low-level state type parameter `S`
* removed property based tests
* make deploy scripts in default template optional

## Checklist

      hard-to-understand areas.

### DIFF
--- a/templates/cis2-nft/src/lib.rs
+++ b/templates/cis2-nft/src/lib.rs
@@ -58,12 +58,11 @@ pub struct MintParams {
 
 /// The state for each address.
 #[derive(Serial, DeserialWithState, Deletable)]
-#[concordium(state_parameter = "S")]
-pub struct AddressState<S = StateApi> {
+pub struct AddressState {
     /// The tokens owned by this address.
-    pub owned_tokens: StateSet<ContractTokenId, S>,
+    pub owned_tokens: StateSet<ContractTokenId>,
     /// The address which are currently enabled as operators for this address.
-    pub operators:    StateSet<Address, S>,
+    pub operators:    StateSet<Address>,
 }
 
 impl AddressState {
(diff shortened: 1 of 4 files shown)


templates/default/cargo-generate.toml
```diff
@@ -1,2 +1,8 @@
 [template]
 cargo_generate_version = ">= 0.17.0"
+
+[placeholders]
+include_deploy_scripts = { type = "bool", prompt = "Include deploy scripts (a Rust program for deploying and initializing the contract on chain)?", default = false }
```
allanbrondum:
```suggestion
include_deploy_scripts = { type = "bool", prompt = "Include Rust based smart contract deployments scripts (alternative to using concordium-client to deploy the contract)?", default = false }
```

templates/cis2-nft/src/lib.rs
```diff
@@ -58,12 +58,11 @@ pub struct MintParams {
 
 /// The state for each address.
 #[derive(Serial, DeserialWithState, Deletable)]
-#[concordium(state_parameter = "S")]
```
limemloh, line 61: Don't recall how these templates are versioned and released, but support for deriving `DeserialWithState` without `#[concordium(state_parameter = "S")]` was added recently right? and concordium-std version in the template will need to point to a version which support this.

limemloh (approved):

templates/cis2-nft/src/lib.rs
allanbrondum: Yes, when we release the new templates, they must be updated to point to the new concordium-std release (which has also not been created yet), added a reminder here: https://linear.app/concordium/issue/COR-2479/release-for-wasm32v1#comment-6af6461c

